PR #22, opened 2023-12-05
Feat/sprout v1 diagrams: Adding diagrams for showing how data will be uploaded

### LINKED ISSUE #1: Align event storming items into the documentation or as an Issue if we need to discuss it
+ [x] User flow of Sprout (diagram) FOCUS #27 
+ [ ] Hold user flow up against event storming to ensure that we include everything we intended
+ [x] Create post documenting events in Sprout (general description of Sprout; things that are not in user flow) #27 
+ [X] Update sequence diagrams after user flow has been reviewed (Schema Creation and Data Upload) #22 
  - We have agreed to keep sequence out of date for now

signekb:
@K-Beicher and I have created initial diagrams of how users will upload data (a csv as a starting point) to start the discussion of how the flow will be. This started with us looking through the event storming png and then feeling the need to visualise the processes that are needed for data upload.

We imagine that the users will first need to prepare the system for data upload (Schema Creation), and then the data will be uploaded (Data Upload). We have divided these processes into two diagrams. 

We welcome any feedback with (even more) open arms (than usual) :o) 

Note: In these diagrams we ignore the question of which user roles will be able to do what.


lwjohnst86:
What's the difference between "creation-flow" and "creation-sequence"? They have the same title.

signekb:
@lwjohnst86 I understand your confusion 😛  If I remember correctly (my pre-flu self seems like another person), we started out with the flow diagram, since it was a bit simpler and easier to get an overview with that. Then, we moved over to the sequence ones instead. So, I think the flow diagram might be obsolete now. Unless you like it more than the sequence ones? 😉

K-Beicher:
As mentioned above by @signekb, for me the flow diagrams are the ones you create with/for the users, they can be easily worked on in workshops, and gives the 'normale' user a chance to specify what it is that they 'do'.  
The sequence diagrams then takes that information and translate it into something that a developer can work from.  Ideally (in my world), this is done between the person responsible for the user specifications and the dev team/developer.

signekb:
@lwjohnst86 The sequence diagrams are still not up to date, but I think we agreed to leave them be for now. However, the user flows + post are done and ready to be merged into main. Not sure where this leaves us. Should we just merge this branch (including the sequence diagrams) to main for now? Or leave them in a separate branch to be completed later?

lwjohnst86:
I think it's fine if they aren't finished, mark the pr as ready and I can look it over briefly and merge it in 🤩

lwjohnst86 (approved):

PR #24, opened 2023-12-07
Docs: Update users description

### LINKED ISSUE #4: Admin: Meetings and minutes

### LINKED ISSUE #4: Move description of user roles into Sprout repo from design docs

lwjohnst86:
Closes seedcase-project/seedcase-project#4


docs/design/users.qmd
```diff
@@ -0,0 +1,27 @@
+---
+title: Target users
+---
+
+{{< include ../_wip.qmd >}}
+
+Sprout is built for these target users.
+
+## Users inputting collected data into the database
```
signekb, line 9: Add specific name of this role?
As I suggested in a comment under seedcase-project/seedcase-project#197, it could be something like "Data Contributor" or "Data Uploader"? 
```suggestion
## Data Contributor: Users inputting collected data into the database
```

docs/design/users.qmd
```diff
@@ -0,0 +1,27 @@
...
+
+These users we assume are authorized research or data collection centers, researchers, or technicians. They would be the people responsible for inputting data into the database of the specific Data Resource managed by Sprout.
```
signekb, line 11:
```suggestion
We assume that these users are from authorized research or data collection centers, researchers, or technicians. They are the people responsible for inputting data into the database of the specific Data Resource managed by Sprout. 
```

docs/design/users.qmd
```diff
@@ -0,0 +1,27 @@
...
+{{< include ../_wip.qmd >}}
+
+Sprout is built for these target users.
+
+## Users inputting collected data into the database
+
+These users we assume are authorized research or data collection centers, researchers, or technicians. They would be the people responsible for inputting data into the database of the specific Data Resource managed by Sprout. 
+
+1. Authorized centers (general practitioner clinics or hospitals) upload standardized and routine data through a public web portal.
+2. Authorized researchers upload generated data (e.g. created new variables of existing ones) from projects using data from the Data Resource, either through an internal portal or a public web portal.
```
signekb, line 14:
```suggestion
2. Authorized researchers upload generated data (e.g. creates new variables of existing ones) from projects using data from the Data Resource, either through an internal portal or a public web portal.
```

docs/design/users.qmd
```diff
@@ -0,0 +1,27 @@
...
+3. Authorized researchers upload collected data through a local-only portal.
```
signekb, line 15: What is meant by a "local-only portal"? :)

docs/design/users.qmd
```diff
@@ -0,0 +1,27 @@
...
+4. An authorized research group uploads collected data through an internal portal.
```
signekb, line 16: Isn't this included in 2.? Not sure how to distinct between "research group uploads" and "researchers"? Wouldn't it always be one person, i.e., a specific user, uploading at a time? Or is it because a research group might have a joint user? :)

docs/design/users.qmd
```diff
@@ -0,0 +1,27 @@
...
+
+Some data will need to be manually processed and cleaned by the users before being inputted (or input in raw form and processed with a plugin or extension).
```
signekb, line 18: Include example?

docs/design/users.qmd
```diff
@@ -0,0 +1,27 @@
...
+
+## Users who are administrators and data controllers/owners
```
signekb, line 20: Include name of user role in header? Something like "User Access Administators"? But that doesn't include the data controllers/owners ... hmm... Can't think of a joint name for these two. I feel like this might be two distinct roles that should be split up?

docs/design/users.qmd
signekb, line 1: The numbered lists in the two user groups are numbering slightly different things, it seems. The first is going through who these users are (e.g., authorised researcher) + what they will do and where they will do it (e.g., web portal) and the second is solely going through what these users will do (e.g., "Approve and authorize who can upload to the database"). Can they be "homogenised"? :)

signekb:
Question: Would it make more sense it the "data entering person" handles potential issues with the data instead of the admin/data controllers? 

https://github.com/seedcase-project/seedcase-sprout/blob/bbf135f3a6cb6f57117344759e077af50044e53f/docs/design/users.qmd#L27

lwjohnst86:
@signekb made changes based on your suggestions. About your question on who handles the data... I think direct access to the database should be limited, but they always have control of the data *before* it gets entered.

signekb:
@lwjohnst86 Ah, I might have misunderstood then. I thought that "potential issues with the data" would be a validation error or something that would make the upload fail. And in that case, it makes sense - as you say - that the data entering person/contributor fixes the error in the data (which would be before data has entered the database).

docs/design/users.qmd
```diff
@@ -0,0 +1,33 @@
...
+- Upload collected data through an internal portal.
+- Upload collected data through a local-only portal (on a work computer and not a server).
+- Upload generated data of new variables (for instance, from new calculations of existing variables) from a project using data from the Data Resource, either through an internal portal or a public web portal.
+
+Some data will need to be manually processed and cleaned by the users before being inputted (or input in raw form and processed with a plugin or extension). For instance, some data may have been entered into an Excel spreadsheet, which is well-known to modify the values without the users knowledge (e.g. date values), and may incorrectly save data in the wrong data type. This type of data entry errors need human attention to fix before they can correctly be put into Sprout.
+
+## User Access Administrators
+
+These are users who are the administrators that are responsible for managing and controling access to who can contribute data and whether those data contributors have access to structure the database itself. These admin users may or may not also be the data controllers/owners or they may be acting on behalf of them.
```
signekb, line 22:
```suggestion
User Access Administrators are responsible for managing and controlling user accesses, i.e., who can contribute data and whether those data contributors have access to structure the database itself. These admin users may or may not also be the data controllers/owners or they may be acting on behalf of them. 
```
Mention of "data controllers/owners" here and on L30. Should we explicitly define what we mean by that? Here or somewhere else? :)

docs/design/users.qmd
```diff
@@ -0,0 +1,33 @@
...
+- Upload collected data through an internal portal.
+- Upload collected data through a local-only portal (on a work computer and not a server).
+- Upload generated data of new variables (for instance, from new calculations of existing variables) from a project using data from the Data Resource, either through an internal portal or a public web portal.
+
+Some data will need to be manually processed and cleaned by the users before being inputted (or input in raw form and processed with a plugin or extension). For instance, some data may have been entered into an Excel spreadsheet, which is well-known to modify the values without the users knowledge (e.g. date values), and may incorrectly save data in the wrong data type. This type of data entry errors need human attention to fix before they can correctly be put into Sprout.
+
+## User Access Administrators
+
+These are users who are the administrators that are responsible for managing and controling access to who can contribute data and whether those data contributors have access to structure the database itself. These admin users may or may not also be the data controllers/owners or they may be acting on behalf of them. 
+
```
signekb, line 23:
```suggestion

These users will:
```

docs/design/users.qmd
```diff
@@ -0,0 +1,33 @@
...
+- Approve and authorize who can upload to the database.
+- Approve and authorize who can make changes to the structure of the database.
+- Manage collection centers' access and authorization.
+
+## Data Management Administrators
+
+These users are the administrators responsible for the database itself and the quality of the data that gets entered into Sprout. They may or may not be the same user as the User Access Administrators and the data controllers/owners.
```
signekb, line 30:
```suggestion
Data Management Administrators are responsible for the database itself and the quality of the data that gets entered into Sprout. They may or may not be the same user as the User Access Administrators and the data controllers/owners.
```

docs/design/users.qmd
```diff
@@ -0,0 +1,33 @@
...
+
```
signekb, line 31:
```suggestion

These users will:
```

docs/design/users.qmd
```diff
@@ -0,0 +1,33 @@
...
+- Initial structuring of core database tables (e.g. the database key to link data together) and allocating space in the raw data file storage.
```
signekb, line 32:
```suggestion
- Initially structure the core database tables (e.g. the database key to link data together) and allocate space in the raw data file storage.
```

docs/design/users.qmd
```diff
@@ -0,0 +1,33 @@
...
+- Manage or handle potential issues with the data itself that are detected from validation checks, through direct access to the database and data storage.
```
signekb, line 33: Could we be a bit more specific on what kind of validation checks this user role will handle? :)

signekb (commented):
Nice! Still have a few comments/questions :)

docs/design/users.qmd
```diff
@@ -0,0 +1,33 @@
...
+
+Sprout is built for these target users.
+
+## Data contributors
+
+These are users who will be uploading or inputting collected data into the database. We assume that these users are authorized research or data collection centers, researchers, or technicians. They are the people responsible for inputting data into the database of the specific Data Resource managed by Sprout. 
+
+- Upload standardized and routine data through a public web portal.
+- Upload collected data through an internal portal.
+- Upload collected data through a local-only portal (on a work computer and not a server).
```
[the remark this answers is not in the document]
lwjohnst86: They are the same portal, but *where* they are located is different. Just like a Git repo is the same no matter where it is, whether it is a "remote" repo vs a "local" repo depends on where it is stored. So a "local" portal would be one only accessible the owner of the computer (like a local Git repo), while an "internal" portal would be something like the AUH or AU intranet (not accessible without an organizational login), while the public one is accessible to anyone with an internet connection (still need to login to upload data though). This isn't fully documented yet as I am doing that right now, but parts of this conversation/discussion are at https://github.com/seedcase-project/seedcase-project/issues/59

docs/design/users.qmd
```diff
@@ -0,0 +1,33 @@
...
+- Upload generated data of new variables (for instance, from new calculations of existing variables) from a project using data from the Data Resource, either through an internal portal or a public web portal.
+
+Some data will need to be manually processed and cleaned by the users before being inputted (or input in raw form and processed with a plugin or extension). For instance, some data may have been entered into an Excel spreadsheet, which is well-known to modify the values without the users knowledge (e.g. date values), and may incorrectly save data in the wrong data type. This type of data entry errors need human attention to fix before they can correctly be put into Sprout.
```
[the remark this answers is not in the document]
lwjohnst86: Yes, you are right :relaxed: the design docs don't need to be implemented all at once!

docs/design/users.qmd
lwjohnst86: Hmm, I don't think we need to, because these are legal terms that may depend on the country of the user. EU has these terms described in GDPR, and Denmark I believe has slightly different but fully compatible terms with GDPR. This is one of those cases where.... do we define literally every word we use? (e.g. database, variable, and data). As long as we have a clearly defined "audience/reader" of these docs and software, we can make some assumptions about some common terms.

docs/design/users.qmd
lwjohnst86: Ufff, I'm not sure we can say at this point without getting more information on the data itself. Plus, validation depends heavily on the field of research and the type of data collected.

K-Beicher (approved):
I can't see anything is missing on this.

docs/design/users.qmd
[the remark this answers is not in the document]
signekb: Ah, I see. Thanks for clarifying!

signekb (approved):
👍

PR #43, opened 2023-12-21
build: :tada: Set up start of a Poetry-managed Django app

lwjohnst86:
- Setting up poetry: `poetry init`
- Add Django: `poetry add django=4.2.8`
- Set up Django: `poetry run django-admin startproject seedcase_sprout .`
- Add Django app: `poetry run django-admin startapp app` while in the `seedcase_sprout` directory
- To run the Django project: First `poetry run python manage.py migrate` to make the migration files (we'll learn more about them later) and the `poetry run python manage.py runserver` which will start running the server with the Django contents. 

Poetry creates:

- `pyproject.toml`: This file contains the metadata that Python uses to recognize the project is a Package type setup. It is also used by Poetry to track package dependencies.
- `poetry.lock`: Contains the Python packages and versions that Poetry uses to build a virtual environment (like Docker, but Python-style).

Django creates:

- The `seedcase_sprout/` folder that has the standard files used by Django. We'll learn about what each file does as we go along.
- `manage.py`: This script is what we use to orchestrate/manage/modify the Django project. Running `poetry run python manage.py` will bring up a list of commands available to use.

Related to #38


philter87 (approved):

K-Beicher:
Is this something that we should all play around with, or will one person setting it up and merging it into sprout send it to the rest?

lwjohnst86:
@K-Beicher I'm not sure I understand. Are you asking if everyone needs to run these commands? No, they don't, because we already did this yesterday, that's what this PR is for. What do you mean about sending it around? Once this is merged in, we can build the Django project in anyone's computer.

K-Beicher:
You are very good at interpreting my pre-coffee ramblings, yes, that was exactly what I was wondering. That because you've done this once (yesterday), once it is merged and part of sprout, then we can play with the Django app on our local machines :-)

PR #45, opened 2024-01-02
chore: :wrench: Create custom conventional branch names to use.

lwjohnst86:
I want to start using Conventional Branches, but the options for names don't work well for our purposes. So I updated them.


signekb (approved):
Nice! 🎉

PR #51, opened 2024-01-04
'app' folder moved, packages added: whitenoise, dj-database-url and gunicorn

philter87:
Sorry if this PR looks a bit scary. It does several small things:

1. The 'app' folder is moved to the root of the repository. The convention is that apps are located at the same directory level as manage.py and the app seems break if this is not done
2. The package whitenoise is added to easily server static files from the Django application (It seems weirdly complex without it). For instance, I have added the seedcase logo at the url '/images/seedcase.png'
3. The template 'hello.html' is added to verify that we can show static content (ex. seedcase.png). The template is rendered at the url: '/hello'. See the picture below
4. The package dj-database-url is added, so we can configure the database with a single line. For a sqlite it is 'sqlite:///PATH' and for postgres it is 'postgres://USER:PASSWORD@HOST:PORT/NAME'. This is convenient, when we need to run the application in production or demo environment
5. gunicorn is added. This will be used when we need to run the application in production or in a demo environment. This was needed to verify that whitenoise is working as expected
6. The project can now use the two environment variables DJANGO_SECURITY_KEY and DJANGO_DEBUG to configure the django security key and to disable debug mode (which should not be used in production)

image


seedcase_sprout/settings.py
```diff
@@ -20,17 +23,18 @@
...
+DEBUG = os.environ.get('DJANGO_DEBUG', '') != 'False'
 
-ALLOWED_HOSTS = []
+ALLOWED_HOSTS = ['127.0.0.1', 'localhost']
 
 
 # Application definition
 
 INSTALLED_APPS = [
+    'app',
```
lwjohnst86, line 37: Does it make sense to name this something else? Something more descriptive? What's the standard here?

lwjohnst86 (approved):
Really nice! Thanks @philter87. Seems mostly straightforward! I only have one comment, that isn't necessary for a change (not at this point anyway).

seedcase_sprout/settings.py
philter87: Valid point! If we only expect one app, then another (maybe more) common name is 'pages'. 

Or if we expect to divide the project in more apps, we can name it based on the feature or domain. Maybe in this context it could be sprout, upload, data_upload, data_import or something related.

K-Beicher (approved):
It all sounds sensible, and as far as I can see all the stuff mentioned is where it is should be based on the description.

PR #52, opened 2024-01-05
Simple dockerfile and docker-compose is created

### LINKED ISSUE #42: Build of docker image and publish to Github Container Registry

philter87:
Dockerfile and related is added in this PR. The Dockerfile is needed when deploying to render.com/fly.io.

Changes: 
- **Dockerfile**. Instructions for building the docker image. Optimized a bit for build speed and size. We use gunicorn to run the Django application. The image size could be even smaller with an "alpine" image, but I prefer the image to contain more features (ssh, bash, etc)
- **docker-compose.yml**. A simple docker-compose.yml file is added to allow the build and execution of a container in one command "docker-compose up". 
- **.dockerignore** file is added to exclude files which are not relevant for the docker image. It follows the syntax of .gitignore

Also closes #52 and closes #42


.github/workflows/docker-build-and-publish.yml
```diff
@@ -0,0 +1,32 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Add tag for new version
+        id: semanticVersion
+        uses: anothrNick/github-tag-action@1.67.0
+        env:
+          DEFAULT_BUMP: patch
+          DRY_RUN: github.ref != 'refs/heads/main'
+          GITHUB_TOKEN: ${{ secrets.GIT_TOKEN}}
```
lwjohnst86, line 18:
```suggestion

      - name: Extract metadata (tags, labels) for Docker
        id: meta
        uses: docker/metadata-action@v5
        with:
          images: ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}
```

I'm not entirely sure what the step that you added does @philter87. Does it make the tag automatically? If so, I feel that should be a separate action maybe?

As described in https://docs.github.com/en/actions/publishing-packages/publishing-docker-images, using this to get information about the tags and labels we (manually?) add. To be honest, I don't completely know what meta data output this gives. How are tags and labels defined in Docker?

.github/workflows/docker-build-and-publish.yml
```diff
@@ -0,0 +1,32 @@
...
+      - name: 'Login to GitHub Container Registry'
+        uses: docker/login-action@v1
+        with:
+          registry: ghcr.io
+          username: ${{github.actor}}
+          password: ${{secrets.GITHUB_TOKEN}}
+      - name: 'Build and publish image'
+        run: |
+          docker build . --tag local-seedcase-sprout
+          docker push local-seedcase-sprout ghcr.io/seedcase-project/seedcase-sprout:${{ steps.semanticVersion.outputs.tag }}-${GITHUB_SHA::7}
```
lwjohnst86, line 28:
```suggestion
          
      - name: "Build and publish Docker image"
        uses: docker/build-push-action@v5
        with:
          context: .
          push: true
          tags: ${{ steps.meta.outputs.tags }}
          labels: ${{ steps.meta.outputs.labels }}
```

I'd rather use a standard action for doing this step, since it exists here with the `docker/build-push-action`.

.github/workflows/docker-build-and-publish.yml
```diff
@@ -0,0 +1,32 @@
...
+      - name: 'Release'
+        if: github.ref == 'refs/heads/main'
+        run: |
+          docker push local-seedcase-sprout ghcr.io/seedcase-project/seedcase-sprout:${{ steps.semanticVersion.outputs.tag }}
```
lwjohnst86, line 32:
```suggestion

```

Since I added the "push: branches: main" config at the top, this step isn't required anymore.

README.md
```diff
@@ -13,7 +13,22 @@ pipx install poetry
...
+
+```
+# Run application
+docker-compose up -d
+
+# Check "http://localhost:10000"
+
+# Terminate application
+docker-compose down
+```
```
lwjohnst86, line 34:
````suggestion
```
# Run application
docker compose up -d

# Check "http://localhost:10000"

# Terminate application
docker compose down
```

Or with the `justfile`:

```
just start-docker
just stop-docker
```
````

`docker-compose` is deprecated: https://docs.docker.com/compose/migrate/

**Review by lwjohnst86 — CHANGES_REQUESTED** [review]
Great @philter87!! I made some comments and suggestions, mostly about the GitHub action file. Let me know what you think.

File: .github/workflows/docker-build-and-publish.yml [hunk 5]
```diff
@@ -0,0 +1,40 @@
+name: Deploy Images to Github Container registry
+
+# Actions are triggered on: push to main branch, pull requests targeting the main branch and manual triggers
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    branches:
+      - main
```
lwjohnst86, line 10:
```suggestion
  # push:
  #   branches:
  #     - main
  # pull_request:
  #   branches:
  #     - main
```

PR #63, opened 2024-01-12
Fix/22-change-schema-to-table

### LINKED ISSUE #57: Discussion: Schema vs Table
@K-Beicher and I have discussed the use of the word "Schema" and that it might not be very transparent what that entails for most users (i.e., non-db people). So, we're suggesting that we use "table" instead and remove all mentions of schemas in the description of the user flow as well as in the wire frames. 
E.g., "Create New Table" , "Add Data to Existing Table" instead of "Create New Schema" and "Add Data to Existing Schema"? 

What do you think? 🌻

signekb:
As discussed at the update meeting on Tuesday 9 January, we will use "Table" instead of "Schema", since it is mostly a legacy term and the way we use it now fits better with "table" than "schema". 

I have replaced all mentions of schema(s) with table(s) in all files related to user flow and in the sequence diagrams, but haven't replaced it in runtime-view and C4 related files. 

PS not sure where I got that the issue number was 22 (used in branch name), but this actually closes #57


lwjohnst86 (approved):
Wonderful :stars:

PR #67, opened 2024-01-15
fix: Another fix to the test workflow

philter87:
## Description
- This PR should fix workflow

## Testing

- [ ] Yes

## Checklist

### DIFF
--- a/.github/workflows/docker-build-and-publish.yml
+++ b/.github/workflows/docker-build-and-publish.yml
@@ -17,9 +17,9 @@ env:
 
 jobs:
   push-store-image:
-    permissions: 
-    	contents: read
-    	packages: write
+    permissions:
+      contents: read
+      packages: write
     runs-on: ubuntu-latest
     steps:
       - name: Checkout repository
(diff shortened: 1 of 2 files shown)


.github/workflows/test.yml
```diff
@@ -15,20 +15,20 @@ jobs:
...
         with:
           python-version: '3.11'
           cache: 'poetry'
 
-      - name: Install poetry
-        run: pipx install poetry
-
       - name: Install dependencies with poetry
-        run: poetry install
+        run: poetry install --no-interaction --no-root
```
signekb, line 28: Just a question for my understanding: Why do we use `--no-root`?

.github/workflows/docker-build-and-publish.yml
signekb, line 1: Ignoring this file, as described in the PR

signekb (approved):
👍

.github/workflows/test.yml
philter87: That is a good question. Poetry handles packaging and dependencies. In this case, we are not worried about packaging (seedcase-sprout is currently not used by anyone else), so we just install dependencies

lwjohnst86: @signekb in this action, it doesn't need to be installed since we aren't building anything for others to use. In this action we are only running our tests, nothing else.

lwjohnst86: This also isn't a formal Python package, so we don't need to install the "seedcase_sprout" package itself. We are developing a Django project which is a bit different than a Python package.

lwjohnst86 (approved):
Wonderful :grin:

PR #92, opened 2024-01-15
fix: Deploy to fly.io is fixed

philter87:
## Description

- This PR fixes deployment to fly.io. The port was changed to the default fly.io port '8000' and the console_command was removed

This PR is mostly for Luke

## Testing

- [ ] Yes

We will se if it works in fly.io


lwjohnst86 (approved):
Great thanks!

PR #93, opened 2024-01-15
Fix/comment style in vscode settings json

signekb:
## Description

- This PR fixes the comment style (#) in the settings.json file since this is not a valid comment style for json (which does not accept comments at all, actually).
- I noticed that VS Code didn't like the comments in the settings.json and found that VS Code have a "json with comments" mode (.jsonc) which ´settings.json´ apparently is under the hood (see [vs code documentation](https://code.visualstudio.com/docs/languages/json#_json-with-comments) for reference)
- In addition, I have changed a semi colon to a comma 

## Testing

- [ ] Yes

Tests are not needed for the settings.json.  

## Reviewer Focus

Reviewers should focus on whether I have made a syntax error somewhere (but I have only changed `#` to `// //` and `;` to `.`, so there isn't be much to review)

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -20,6 +20,6 @@
     "refactor",//A code change that neither fixes a bug nor adds a feature//
     "style", //Changes that do not affect the meaning of the code (white-space, formatting, missing semi-colons, etc)//
     "test", //Adding missing tests or correcting existing tests//
-];
+],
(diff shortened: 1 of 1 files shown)


lwjohnst86 (approved):
Aaah, oops that was my fault for including those comment style. Thanks!

lwjohnst86:
I'll also update the `.github` config.

PR #98, opened 2024-01-16
Feat/models metadata

### LINKED ISSUE #69: Define data model for metadata for table and columns in Django

philter87:
## Description

- **This PR adds data models for the table and column metadata**. The most important changes are in the file 'models.py' where the models are defined

**For now, you should ignore the files in the migrations folder ('0001_initial.py' and '__init__.py') which are autogenerated by Django**

## Knowledge sharing notes
- **Models**. First we wrote the three classes in the file 'models.py' (every class will result in a table in the database). You should spend most of you energy validating the content of these. Are we missing anything?
- **Relation between models**. The tables/models are related to each other. The relation is established with "models.ForeignKey". For instance, a row in TableMetadata will likely be related to several rows in ColumnMetadata (an uploaded csv file will create a single row in TableMetadata and several related rows in ColumnMetadata). It is convenient to establish this relation between the models, because when we delete a row in TableMetadata, then the corresponding rows in ColumnMetadata should be deleted as well
- **Creating Migrations**. The files in the migration folder are created with the command 'python manage.py makemigrations app'. These files are automatically generated by Django based on the content of models.py. The migration files define the required database changes to support the models/classes defined in models.py
- **Applying the Migrations**. The changes defined in the migration folder are applied to the database when we run 'python manage.py migrate'
- **Tests**. I have added two tests in the app/tests.py
  - The first test 'test_create_metadata_for_a_table_and_columns_and_verify_creation' simply illustrates how to add stuff to tables in Django (see the method create_metadata_table_and_column)
  - The second test 'test_verify_foreign_key_constraint_by_deleting_table_which_should_delete_column', demonstrates the foreign key constraint. So when a row in TableMetadata is deleted, then the related rows in ColumnMetadata are deleted as well
- **settings.py**. I have made a small change to settings.py regarding ALLOWED_HOSTS. This property define a list of allowed host. The default option is ['127.0.0.1', 'localhost'], which is required when viewing the application locally in browser (we will get an error locally if it is missing). We also want to allow the fly.io host 'https://seedcase-sprout.fly.dev/', but instead of overwriting the ALLOWED_HOSTS property, we append 'https://seedcase-sprout.fly.dev/' to the list when the environment variable FLY_APP_NAME is available (which is only the case when running in the fly.io environment)

## Related Issues

Closes #69 

## Testing

- [X] Yes

## Reviewer Focus

You should start with the models.py file

## Checklist

For general documentation:

- [ ] Spell-check
    - [ ] US
    - [ ] UK
- [ ] Did the page(s) preview correctly on your machine without breaking
- [ ] New category words (keywords) (if any) added to the code snippet file


app/models.py
```diff
@@ -1,3 +1,23 @@
 from django.db import models
 
-# Create your models here.
+
+class TableMetadata(models.Model):
+    name = models.CharField(max_length=128)
+    original_file_name = models.CharField(max_length=200)
+    description = models.TextField()
+    created_at = models.DateTimeField(auto_now_add=True)
+    created_by = models.IntegerField()
```
signekb, line 9: Do we want the created_by to be an integer (`user_id`/PK, I'm guessing?) or could it be nice, if it was the user name instead?

app/tests.py
```diff
@@ -1,8 +1,47 @@
...
+
+        # Assert
+        self.assertEqual(0, TableMetadata.objects.count(), "Table should be deleted")
+        self.assertEqual(0, ColumnMetadata.objects.count(), "Column should be deleted")
+
+    @staticmethod
+    def create_metadata_table_and_column():
+        table_metadata = TableMetadata(name="TestTable",
+                                       original_file_name="test_file.csv",
+                                       description="Table create for testing",
```
signekb, line 34:
```suggestion
                                       description="Table created for testing",
```

app/tests.py
```diff
@@ -1,8 +1,47 @@
...
+
+        column_data_type = ColumnDataType(display_name="Text")
+        column_data_type.save()
+
+        column = ColumnMetadata(table_metadata=table_metadata, name="ColumnName",
+                                title="Friendly Column Name",
+                                description="Description of column",
+                                data_type=column_data_type,
+                                allow_missing_value=True,
+                                allow_duplicate_value=True)
```
signekb, line 46: Should we maybe include two columns in the test table? Then we can test that:
-  all columns are created (and not e.g., only the first column) in `test_create_metadata_for_a_table_and_columns_and_verify_creation` and 
- all columns in a table (and not just one) is deleted in `test_verify_foreign_key_constraint_by_deleting_table_which_should_delete_column`. 

We aren't testing that the deletion of a table only deletes all columns from that table either, actually. Should we do that by including two tables in the tests?

app/tests.py
signekb, line 1: Maybe we should add a test about deletion behaviour in one-to-many and many-to-many tables? 
- E.g., what happens to main table if a reference value is deleted? E.g., what happens in the column meta data, if we delete a data type? It should not be possible to delete a datatype while it's in use

app/models.py
lwjohnst86: user name is explicit and easier to read rather than integer, so stick with things that are clearer.

app/models.py
```diff
@@ -1,3 +1,23 @@
...
+class ColumnDataType(models.Model):
+    display_name = models.TextField()
+
+
+class ColumnMetadata(models.Model):
+    table_metadata = models.ForeignKey(TableMetadata, on_delete=models.CASCADE)
+    name = models.CharField(max_length=200)
+    title = models.CharField(max_length=200)
+    description = models.TextField()
+    data_type = models.ForeignKey(ColumnDataType, on_delete=models.PROTECT)
```
lwjohnst86, line 21: Hmm, I'm not sure I understand why we need to connect these this other `ColumnDataType` database, what's the rationale for that?

justfile
```diff
@@ -29,4 +32,4 @@ format-python:
 start-app:
   poetry run python ./manage.py runserver
   # Not sure if this works for others?
-  # xdg-open http://localhost:8000
\ No newline at end of file
+  # xdg-open http://localhost:8000
```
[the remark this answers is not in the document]
lwjohnst86: It's to automatically open the localhost when developing on your computer. But I haven't tested it, and I don't know if it works for non-Linux people. So I put it there for now to eventually get to testing it and seeing if it is useful or not.

app/tests.py
```diff
@@ -1,8 +1,47 @@
...
 
-# Note: Two types of TestCase: django.test or unittest
-# https://docs.djangoproject.com/en/5.0/topics/testing/overview/
-class SmokeTest(TestCase):
-    def test_smoke(self):
-        self.assertTrue(True)
 
+class MetadataTests(TestCase):
+
+    def test_create_metadata_for_a_table_and_columns_and_verify_creation(self):
```
lwjohnst86, line 7: I'm not super familiar with testing in Python. Is it common practice to have such long test function names? I know having them be descriptive helps, but just wasn't aware of the length of them. Not complaining, just curious :relaxed:

lwjohnst86 (commented):
No major issues, just some comments :)

app/tests.py
philter87: Naming of tests is a bit of a religion (I don't know if that is the case in python/Django). Here is a list https://dev.to/n_develop/how-do-you-name-your-tests-108c.  I have tried most of them at some point in my "career", but now I just use long descriptive names (last on the list in the link) because then I don't assume anything about the a future reader

app/models.py
K-Beicher: When you can use an ID then that is what you should do.  These fields are mainly for internal auditing purpose, it would be very rare for people to need to download this data.  There are a number of other reasons for using an integer, I'd be happy to elaborate if needed.

app/models.py
K-Beicher: We want them as controlled values that can potentially be used elsewhere in the system, thus a table makes more sense.

philter87: It is sort of a "relational-database-thinking".

The field data_type in ColumnMetadata (the model) will generate a field called 'data_type_id' in the database table 'app_columnmetadata'. The field 'data_type_id' will contain a integer that references a row in app_columndatatype. 

I know this seems 'over-engineered', but is convenient when we want to change the 'display_name' or the 'description' of a DataType. This change is then automatically 'propagated' to all rows in ColumnMetadata.

**Another solution** is to have the datatypes hardcoded as a list in the code. However, I would still like a id, so we can change the display_name and description at one location in the code



image

lwjohnst86 (dismissed):
Wonderful, I'm happy with these changes so far

PR #107, opened 2024-01-19
Feat/just generate single puml

signekb:
## Description

- This PR adds a command to the just file to create a png from a specific puml
- In addition, it renames the command to create pnga for all pumls from `generate-puml` to `generate-puml-all`

This is useful, when only one png needs to be updated (because we currently have pumls that are in progress which we don't include pngs for). The `generate-puml` command put out pngs for all pumls which does not fit what I am currently doing with updating the user flow diagrams (and don't want to include pngs for e.g., the runtime-view pumls).

Hope this makes sense :)


lwjohnst86 (approved):
Fantastic! 🤩 🎉

PR #108, opened 2024-01-19
build: :hammer: Add just recipe to install dependencies with poetry

lwjohnst86:
## Description

Updated justfile to run fixes via the check and format, while having another recipe that runs the linter only (shows things that need manual changes).

Also fix the port in the docker compose to be consistent with the other ports (use 8000).

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -3,4 +3,4 @@ services:
   app:
     build: .
     ports:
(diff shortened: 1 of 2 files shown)


signekb (approved):
Looks good 👍 🎉

philter87 (approved):

PR #110, opened 2024-01-22
Feat/file upload page

### LINKED ISSUE #78: Build html page for table extraction from "uploaded" csv
Image

philter87:
## Description

- This PR adds a page to upload a csv file "/file-upload/<table_id>"

Once again, I am sorry for creating a big PR. Here is a recap of the changes:
- **file_upload.py**. The method file_upload in **file_upload.py**. The method handles both GET and POST requests
   - **GET**. A get request is created when when the user visits '/upload-file/<table_id>' in the browser
   - **POST**. A POST request is called when the button "EXTRACT DATA STRUCTURE" is clicked and the form is submitted. This can result in two outcomes: 
     -  **Failed POST request**. The post request will fail if the file is not correct. This will cause the page to rerender but where upload_error is not empty. This wil cause a dialog to open (see second picture below). 
     -  **Successful POST request**. The post request is successful if the file is a valid csv and the user is redirected to '/edit-table-metadata/<table_id>' (which is currently not implemented). Furthermore, I have create a simple function to extract and persist column metadata. This needs to be redone in another Github issue
- **file-upload.html**. See the two pictures below
- **file-drop-zone.html and file-drop-zone.css**. The "file drop zone" is a bit complicated, so it is is defined in two separate files. We want to support the ability to drag a file into the "drop zone" which require some javascript (which is okay to skip)
- **tests_views.py**. I have added five tests to cover the different outcomes
- **db_test_utils.py**. I have create a file to contain testing related functionality which is likely used in several test suites.
- **views folder** is added instead of having a single views.py file

**WITHOUT UPLOAD ERROR (extract button is disabled)**
image

**WITH UPLOAD ERROR**
image

**AFTER UPLOAD (of file with long name)**
image

## Related Issues

Closes #78 

## Testing

- [X] Yes

## Reviewer Focus

The file_upload.py file is the most essential. It loads the template 'file-upload.html'.
I have created a folder called views instead of views.py

## Checklist


signekb:
@philter87 when the file is uploaded, should the original file name be saved to the db? There's a field for it in the TableMetadata model

app/tests/tests_views.py
```diff
@@ -0,0 +1,58 @@
...
+    def test_render_file_upload_view_and_verify_that_table_is_loaded(self):
+        # Arrange
+        table_name = "Table Name"
+        create_table(table_name).save()
+
+        # Act
+        response = Client().get("/file-upload/1")
+
+        # Assert. assertContains checks if status_code=200 and
+        # html should contain "TestTable"
```
signekb, line 19: Should be "Table Name"

app/templates/file-drop-zone.html
```diff
@@ -0,0 +1,91 @@
...
+  <p class="grey-text">Supported formats: .csv</p>
+
+  <div class="row center-align">
+    <button disabled id="submit_button" type="submit" value="OK">
+      EXTRACT DATA STRUCTURE
+    </button>
+  </div>
+</form>
+
+<script>
```
signekb, line 51: Maybe we can go over what happens in this script part on Monday :)

app/templates/file-upload.html
```diff
@@ -0,0 +1,25 @@
...
+  <h4>{{ table_name }}</h4>
+  <p>To create the structure for your new table, please upload a .csv data file to be
+    used as a template</p>
+
+  <!-- The file drop zone is included -->
+  {% include 'file-drop-zone.html' %}
+
+  <!-- Dialog is only showing if upload_error is not empty -->
+  {% if upload_error %}
+  <dialog class="modal active" id="dialog">
```
signekb, line 13: Should/can the dialog close when the user starts to upload a new file? It seems like the dialog is open until you click "OK"

app/tests/db_test_utils.py
```diff
@@ -0,0 +1,27 @@
...
+    )
+
+
+def create_column(name, table):
+    return ColumnMetadata(
+        table_metadata=table,
+        name=name,
+        title=name + "Title",
+        description=name + " description",
+        data_type_id=1,
```
signekb, line 16:
```suggestion
        data_type=1,
```
Is it possible to just set this ForeignKey as an int?

app/tests/tests_models.py
```diff
@@ -29,29 +29,13 @@ def test_verify_foreign_key_constraint_by_deleting_table_which_should_delete_col
...
-        column = ColumnMetadata(table_metadata=table_metadata, name="ColumnName",
-                                title="Friendly Column Name",
-                                description="Description of column",
-                                data_type=column_data_type,
-                                allow_missing_value=True,
-                                allow_duplicate_value=True)
-        column.save()
+        table_count = TableMetadata.objects.count()
+        self.assertEqual(1, table_count, "Table should not be deleted")
+        column_count = ColumnMetadata.objects.count()
```
signekb, line 40: Add that columns for that table specific is deleted (by adding `table_id` here) and not just all columns? 

(not sure the suggestion is correct)


```suggestion
        column_count = ColumnMetadata.objects.filter(table_id=1).count()
```

app/tests/tests_models.py
```diff
@@ -29,29 +29,13 @@ def test_verify_foreign_key_constraint_by_deleting_table_which_should_delete_col
...
-        column_data_type.save()
-
-        column = ColumnMetadata(table_metadata=table_metadata, name="ColumnName",
-                                title="Friendly Column Name",
-                                description="Description of column",
-                                data_type=column_data_type,
-                                allow_missing_value=True,
-                                allow_duplicate_value=True)
-        column.save()
+        table_count = TableMetadata.objects.count()
```
signekb, line 38: Add that we count whether that specific table exist and not just how many tables exists?

(not sure the suggestion is correct - this would also return `True` if table exists and not count)

```suggestion
        table_count = TableMetadata.objects.filter(table_id=1).exists()
```

app/views/__init__.py
signekb, line 1: Not sure why this is in this file? :)

app/tests/tests_views.py
signekb: ...or the other way around. Define table_name on L12 as TestTable (I think I prefer this - then we use the same name as in the other tests). I have added this as a suggestion above

app/tests/tests_views.py
```diff
@@ -0,0 +1,58 @@
...
+from django.test import Client
+
+from app.models import ColumnMetadata
+from app.tests.db_test_utils import create_table
+
+
+class FileUploadTests(TestCase):
+    def test_render_file_upload_view_and_verify_that_table_is_loaded(self):
+        # Arrange
+        table_name = "Table Name"
```
signekb, line 12:
```suggestion
        table_name = "TestTable"
```

app/tests/tests_views.py
```diff
@@ -0,0 +1,58 @@
+from django.core.files.uploadedfile import SimpleUploadedFile
+from django.test import TestCase
+from django.test import Client
+
+from app.models import ColumnMetadata
+from app.tests.db_test_utils import create_table
+
+
+class FileUploadTests(TestCase):
+    def test_render_file_upload_view_and_verify_that_table_is_loaded(self):
```
signekb, line 10:
```suggestion
    def test_render_file_upload_view_and_verify_that_table_metadata_is_loaded(self):
```
Added "metadata"

app/tests/tests_views.py
```diff
@@ -0,0 +1,58 @@
...
+        # Act
+        response = Client().get("/file-upload/1")
+
+        # Assert. assertContains checks if status_code=200 and
+        # html should contain "TestTable"
+        self.assertContains(response, table_name)
+
+    def test_upload_of_file_should_create_columns_in_database(self):
+        # Arrange
+        table_name = "Table Name"
```
signekb, line 24:
```suggestion
        table_name = "TestTable"
```

app/tests/tests_views.py
```diff
@@ -0,0 +1,58 @@
...
+        create_table(table_name).save()
+        file = SimpleUploadedFile("file.csv", b"first_name,second_name,age")
```
signekb, line 26: What's the b in `b"[...]"` doing?

app/tests/tests_views.py
```diff
@@ -0,0 +1,58 @@
...
+
+    def test_upload_failed_with_missing_file_error(self):
+        create_table("Table Name").save()
+
+        response = Client().post("/file-upload/1")
+
+        self.assertContains(response, "The file is missing!")
+
+    def test_upload_failed_with_invalid_csv_header(self):
+        create_table("Table Name").save()
```
signekb, line 53:
```suggestion
        create_table("TestTable").save()
```

signekb: above, e.g., in L24-25, you have first created the table name variable and then used it. Maybe align so the creation of the table is identical in the tests? I don't have a preference to either method :)

app/tests/tests_views.py
```diff
@@ -0,0 +1,58 @@
...
+        file = SimpleUploadedFile("file-with-bad-headers.csv", b"no valid headers")
```
signekb, line 54: Is the headers not valid here bc there's spaces? Or why are they not valid?

app/views/file_upload.py
```diff
@@ -0,0 +1,52 @@
...
+        return "The file is missing!"
+
+    if not uploaded_file.name.endswith(".csv"):
+        return "Unsupported file format: ." + uploaded_file.name.split(".")[-1]
+
+    return extract_and_persist_column_metadata(table_id, uploaded_file)
+
+
+def extract_and_persist_column_metadata(table_id, uploaded_file):
+    # A more complicated function needs to be added
```
signekb, line 36: Do you have any considerations on what a "more complicated function" should include already?

app/views/file_upload.py
```diff
@@ -0,0 +1,52 @@
...
+    table_metadata = TableMetadata.objects.get(pk=table_id)
+    file_upload_data = {"table_name": table_metadata.name, "upload_error": upload_error}
+    return render(request, "file-upload.html", file_upload_data)
+
+
+def extract_columns_and_get_error(table_id, files):
+    # File is missing
+    uploaded_file = files.get("uploaded_file", None)
+    if not uploaded_file:
+        return "The file is missing!"
```
signekb, line 27: When will this happen? I doesn't seem to be possible to click "Extract" without a file?

app/views/file_upload.py
```diff
@@ -0,0 +1,52 @@
...
+
+from app.models import ColumnMetadata, TableMetadata
+
+
+def file_upload(request, table_id):
+    if request.method == "GET":
+        return render_file_upload_page(request, table_id, "")
+
+    upload_error = extract_columns_and_get_error(table_id, request.FILES)
+    if upload_error:
```
signekb, line 11:
```suggestion
    if upload_error is not None:
```

signekb (commented):
Nice! I have some questions (who is surprised). We'll discuss them on Monday, where I'll update my comments with the answers. 

When I try to upload a .csv file, it still fails with the overlay, saying "Unable to extract headers". I think this might be bc it's a semicolon separated csv (also tried with a tab separated csv - also failed). 
EDIT: This is handled in another PR. #123

app/views/file_upload.py
```diff
@@ -0,0 +1,52 @@
+from django.shortcuts import render, redirect
+
+from app.models import ColumnMetadata, TableMetadata
+
+
+def file_upload(request, table_id):
```
lwjohnst86, line 6: I've been going through and adding docstrings and type hints, just so I can also understand what this does. But I'm a bit confused about where the `table_id` comes from?

app/views/file_upload.py
```diff
@@ -0,0 +1,52 @@
...
+    if request.method == "GET":
```
lwjohnst86, line 7: You mention that there is a POST request, but I only see a GET request. Does it come from somewhere else?

app/views/file_upload.py
```diff
@@ -0,0 +1,52 @@
...
+        return render_file_upload_page(request, table_id, "")
+
+    upload_error = extract_columns_and_get_error(table_id, request.FILES)
+    if upload_error:
+        return render_file_upload_page(request, table_id, upload_error)
+
+    return redirect("/edit-table-columns/" + str(table_id))
```
lwjohnst86, line 14: The way I understand the function is that the other `return` statements return an `HttpResponse`object, but this return is a `HttpResponseRedirect` right? Is there a way to write the code to be more predictable? If we can follow philosophies of more strict functional programming (predictable inputs and outputs), that would make the code easier to reason about and understand. Plus the return type hint will be easier to set, then it can be ` -> HttpResponse:`.

app/views/file_upload.py
```diff
@@ -0,0 +1,52 @@
...
+
+
+def extract_columns_and_get_error(table_id, files):
+    # File is missing
+    uploaded_file = files.get("uploaded_file", None)
+    if not uploaded_file:
+        return "The file is missing!"
+
+    if not uploaded_file.name.endswith(".csv"):
+        return "Unsupported file format: ." + uploaded_file.name.split(".")[-1]
```
lwjohnst86, line 30:
```suggestion
    if not uploaded_file.name.endswith(".csv"):
        raise StopUpload(True)
```

Same as above. Unless there is a really strong reason to use `return` instead.

app/views/file_upload.py
lwjohnst86, line 27:
```suggestion
        raise StopUpload(connection_reset=True)
```

Use a `raise` rather than `return` (same as below). `connection_reset=True` says to stop the upload and restart the connection.

lwjohnst86 (changes requested):

lwjohnst86:
Oops, forgot to write a bit more. It would help a lot if the PRs were a bit small 🥺 🙏🏻 🙇🏻 Maybe split future PRs into code vs html?

Could we name `utils` to something more meaningful? Like `uploading`, so that `csv` and other future folders for extensions could be within it?

app/templates/file-drop-zone.html
```diff
@@ -0,0 +1,91 @@
...
+      <h5 class="center-align bold">DRAG YOUR .CSV FILE HERE</h5>
+      <div class="medium-space"></div>
+      <i class="extra">attach_file</i>
+      <div class="medium-space"></div>
+      <h5 class="center-align bold">OR CLICK TO OPEN YOUR FILE EXPLORER</h5>
+    </div>
+
+    <!-- Drop zone with file name and size. Visible when file is added -->
+    <div hidden id="drag_zone_with_file" class="file-upload-info-box"
+         onclick="event.stopPropagation()">
```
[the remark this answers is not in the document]
signekb: It makes sure that the click event doesn't hit the dropzone that opens the file directory pop-up

app/templates/file-drop-zone.html
```diff
@@ -0,0 +1,91 @@
...
+    <div class="center-align" id="drag_zone_empty">
+      <h5 class="center-align bold">DRAG YOUR .CSV FILE HERE</h5>
+      <div class="medium-space"></div>
+      <i class="extra">attach_file</i>
+      <div class="medium-space"></div>
+      <h5 class="center-align bold">OR CLICK TO OPEN YOUR FILE EXPLORER</h5>
+    </div>
+
+    <!-- Drop zone with file name and size. Visible when file is added -->
+    <div hidden id="drag_zone_with_file" class="file-upload-info-box"
```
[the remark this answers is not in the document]
signekb: Box with file name and info is hidden before we have chosen a file to upload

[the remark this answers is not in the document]
signekb: @philter87 will look into using dropzone.dev for this instead. Hopefully, this will simplify the code a bit (and it's pretty).

app/templates/file-drop-zone.html
signekb: this might be removed with the use of dropzone.dev

app/tests/db_test_utils.py
signekb: Yes, it is.

app/views/__init__.py
signekb: they should be moved out :)

app/tests/tests_views.py
signekb: Because the validator checks that there's more than one column. Here, there's only one.

app/views/file_upload.py
philter87: This is done in another issue #123

app/views/file_upload.py
philter87: You are right. The code that disables the button when there are not files was created later

app/views/file_upload.py
```diff
@@ -0,0 +1,52 @@
...
+
+    return extract_and_persist_column_metadata(table_id, uploaded_file)
+
+
+def extract_and_persist_column_metadata(table_id, uploaded_file):
+    # A more complicated function needs to be added
+    column_names = uploaded_file.readline().decode("utf-8").split(",")
```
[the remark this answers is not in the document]
philter87: Yes, exactly. That is the "complicated function", I want to put here, which might be the pandas.read_csv

app/views/file_upload.py
philter87: Yes, I see why that is confusing. table_id is defined in the url pattern: 
image

app/views/file_upload.py
philter87: Yes, I can make this a bit clearer. The get-method is the simplest to process, so I want to return quickly (to avoid too much nesting). The rest of the function is assumed to be handling "POST" requests, but I will change it a bit since it is unclear

app/views/file_upload.py
philter87: I will need to think about this one. In my head, it is the obvious think to do, but if it is unclear then I need to iterate a bit on it

app/views/file_upload.py
philter87: I need to return a error message to the user, but doing this by rasing an error is a great idea

app/views/file_upload.py
philter87: I think the StopUpload exception is required when you create a custom FileUploadHandler. I am not doing that, but it might be a create way to propagate the error up

philter87:
> Oops, forgot to write a bit more. It would help a lot if the PRs were a bit small 🥺 🙏🏻 🙇🏻 Maybe split future PRs into code vs html?
> 
> Could we name `utils` to something more meaningful? Like `uploading`, so that `csv` and other future folders for extensions could be within it?

Yes, I am sorry about the size of the pull request. This is even after I tried to make it smaller, but I will promise not to do it again

signekb:
@philter87 @K-Beicher and I decided to merge this so we can start putting the pieces from other PRs together with this. We will raise issues if there's something that should be optimised somehow.

PR #123, opened 2024-01-26
Feat/analyse and extract column types 

### LINKED ISSUE #112: Build function to analyse/extract column data types

philter87:
## Description

- This PR adds the method **derive_csv_column_types(file)**. The input is a file and it returns the column types
- These changes are done because of we need to automatically derive the types in a CSV column

### **Logic**
We need to extract the data types for a csv file. The data type for a given column should be the most "narrow" type possible. This is best described by an example. Imagine a single column (called "Chapter") with these values:

| "Chapter" (Column Values) | Derived Type |
| ----------------------------- | ------------ |
| 1                                         | --> bool     |
| 2                                         | --> int        |
| 2.2                                      | --> float     |
| 2.3 Bird singing: Chiffchaff | --> str        |
| 2.4                                      | --> str        |
| 3                                         | --> str        |

For every row, we need to "broaden" the type for the column, see "Derived Type". Notice, that the last rows are stuck on "str" after the 4th row.

**Type Hierarchy?**
Based on this you can imagine a "type hierarchy" of how narrow a type is. I have defined, the type hierarchy like this (from most narrow type to broadest type):

[bool, int, float, time, date, datetime, str]

### **Explanation of the code**
The most essential method is **derive_csv_column_types(file)** which takes a file as input and returns a list of CsvColumnStats. Every CsvColumnStats contains the stats for a column: the type, a TypeParser (to convert a string the relevant type), the number of rows read and if an empty value have been found.

A cell in a column is analysed by calling "CsvColumnStats.analyse(value)".

## Should we use pandas for extracting types?
This is a bit tricky question. It seems like an obvious choice but:
- **con**. pandas supports datetime64 and time_delta[ns], which is not exactly the same as datetime, time and date (this might be okay, because time_delta[ns] is similar to time and dates can be stored as datetime64)
- **con**. pandas is not able to automatically detect which columns are datetime64 or time_delta[ns]. To convert columns, you need to either add a date_format to csv_read or you need to "manually" detect if a column can be converted to datetime or time_delta (with pandas.to_datetime() and pandas.to_timedelta).
- **small con**. Maybe memory inefficient. We need to convert entire columns at the time (and not iterate through the dataset row by row).
- **con**. pandas does not support true-ish and false-ish values like: 'True', 'true', 'yes', 'y', 'ja', 0. In this case, we would need our own parser anyway

- **huge pro**. If we know, that we will use pandas later when analysing the data, then we need to somehow align with the pandas types anyway. So this is a very obvious reason to stick to pandas

## Related Issues

Closes #112 

## Testing

- [X] Yes

## Reviewer Focus

The most important thing is **derive_csv_column_types(file)** in csv_utils.py. 

Also we need to consider how "forgiving" the date parser should be. Currently, it is very forgiving. 

## Checklist


lwjohnst86 (changes requested):
We discussed this a bit via Discord, but this would be a great place for using pandas. There's a function in there that takes finds the data types of the columns: https://pandas.pydata.org/docs/user_guide/io.html#specifying-column-data-types via `pandas.read_csv("file.csv").dtypes`. This would simplify the code a lot more, rather than have to implement yourself from scratch. Could you try refactoring to use Pandas instead? (we can run performance checks later after we've made a more complete prototype).

philter87:
> Reference in new issue
> 
> ### Reference in new issue
> Repository
> seedcase-sprout
> Title
> 
> Body

Indeed, the code is complex, and this is why I also wanted to use pandas.read_csv (as we discussed in Discord). But I tried pandas, but I lacked a bit of control. It works well for identifying ints and floats, but I was lacking a bit of control regarding support for datetime, date and time. 

But I will give it another try

lwjohnst86:
@philter87 hmm, I didn't know. That's odd. Well, if it isn't powerful enough, maybe there are other packages? I found something like this, would that be useful? https://github.com/edcrewe/django-csvimport or https://pypi.org/project/dataclass-csv/?

philter87:
> @philter87 hmm, I didn't know. That's odd. Well, if it isn't powerful enough, maybe there are other packages? I found something like this, would that be useful? https://github.com/edcrewe/django-csvimport or https://pypi.org/project/dataclass-csv/?

Thanks, I will take a look

PR #126, opened 2024-01-30
fix: using pandas to parse column types

### LINKED ISSUE #72: Function to upload (at least part of) file to server via Django to use for column metadata extraction

### LINKED ISSUE #112: Build function to analyse/extract column data types

philter87:
## Description

**WAIT! TRY TO USE POLARS INSTEAD**

- This PR adds the method **read_csv_file(file)**. The input is a file and it returns a panda dataframe with derived types in the property called df.dtypes
- These changes are done because of we need to automatically derive the types in a CSV column

### **Logic**
The main method is read_csv_file(file). It is based on pandas.read_csv(file), but with some added features.

Features:
- **dialect detection**. It detects the csv dialect (fx. is the delimiter a comma or a semicolon)
- **remove whitespace**. It removes whitespace in the header
- **boolean-ish**. Columns with boolean-ish values ('yes', 'y', '1', 'True') are converted to booleans. All values in a column needs to be boolean-ish or they remains as strings
- **datetime**. Columns with date-ish values ('2023-01-03 14:32:00') are converted to datetime64. All values needs to be datetime-ish or they remain strings
- **time_delta**. Columns with time-ish values ('12:00:01') are converted to time_delta64. All values needs to be time-ish or they or they reamain as strings

Problems:
- **date**. Pandas does not have a 'date' type, so columns will be converted to datetime (and not date). We will try to address this in another PR 
- **json**. Pandas does not have a 'json' type, so columns will remain string objects. We might address this later

## Related Issues

Closes #112 
Closes #123 

## Testing

- [X] Yes

## Reviewer Focus

The most important thing is **read_csv_file(file)** in csv_parser.py. The tests might be a great way to understand how it works

## Checklist


signekb:
@philter87 Maybe add docstrings and typehints to make it clearer what each function does? 🚀

lwjohnst86:
Must nicer and smaller, thank you!! 🤩 🤗 🙌🏻

lwjohnst86:
For the problem of Pandas and dates, there [this explanation](https://www.geeksforgeeks.org/how-to-convert-datetime-to-date-in-pandas/) about using `.dt.date` to extract only the date from the datatime object.

app/csv_parser.py
```diff
@@ -0,0 +1,140 @@
...
+        A dataframe where column types are derived and converted
+    """
+    df.columns.str.strip()
+    for col in df.columns:
+        _convert_column_if_possible(df, col)
+
+    return df
+
+
+def _convert_column_if_possible(df: DataFrame, column_name: str) -> None:
```
lwjohnst86, line 54: From a functional perspective, I don't know how I feel about an output being `None`. What's the rationale here? (related to the comment above)

app/csv_parser.py
```diff
@@ -0,0 +1,140 @@
...
+    be converted to a different type
+    Args:
+        df: the dataframe with the csv data
+
+    Returns:
+        A dataframe where column types are derived and converted
+    """
+    df.columns.str.strip()
+    for col in df.columns:
+        _convert_column_if_possible(df, col)
```
lwjohnst86, line 49: I'm confused by this. Are DataFrame columns mutable? So you don't need to reassign changes to columns, it will already change them for you? My experience in R might be confusing my understanding here 😝

app/csv_parser.py
philter87: Valid! It will be easier to read if it returns something and it assigns it (even if the column will be unchanged)

app/csv_parser.py
philter87: You are right, I will reassign because it is more clear

philter87:
> For the problem of Pandas and dates, there [this explanation](https://www.geeksforgeeks.org/how-to-convert-datetime-to-date-in-pandas/) about using `.dt.date` to extract only the date from the datatime object.

Okay, now I am a bit confused about how types work in pandas.

So I was just checking the official types supported and date is not part of it (but 'object' is)

When I do the conversion as described in the link. It converts datetime to datetime.date, but the dtype is just an object (which can be everything). 

That might be okay, but then I should not rely too much on dtype

app/csv_parser.py
```diff
@@ -0,0 +1,140 @@
...
+        if val == 0:
+            return False
+        if val == 1:
+            return True
+        raise ValueError
+
+    if val.lower() in ["true", "yes", "1", "y"]:
+        return True
+    if val.lower() in ["false", "no", "0", "n"]:
+        return False
```
lwjohnst86, line 139: This is tricky here. What we want to check is if `val` only contains either "true" or "false" (or "yes"/"no"). But it doesn't do that. For instance:

```
print(["true", "yes", "maybe", "why not"] in ["true", "yes", "1", "y"])
print(["true", "true", "false", "yes"] in ["true", "yes", "1", "y"])
```

returns `false` in both cases. Is there a way to check if that values in `val` *only* contain either True or False, Yes or No, etc?

lwjohnst86: Mmm, I missed the `.apply()` above. So then this is applied to each row in the column? Ok, that makes more sense., nevermind.

app/tests/tests_csv_parser.py
```diff
@@ -0,0 +1,128 @@
+import io
+from typing import Any
+from unittest import TestCase
+import numpy
+from pandas import DataFrame, Series, Timedelta
+
+from app.csv_parser import read_csv_file
+
+
+class CsvTests(TestCase):
```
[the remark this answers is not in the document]
lwjohnst86: I think to the developer one, but also to make sure our assumptions are correct about what the user gives us as CSV files. How does the functions handle when things *aren't* what we expect to be given? I looked into it more, it seems we use `try: ... raise:` within test functions?

app/tests/tests_csv_parser.py
```diff
@@ -0,0 +1,128 @@
...
+        """
+        Testing boolean-ish values are derived as booleans and that the values are
+        converted correct
+        """
+        csv_file = io.StringIO(
+            "b1,b2,b3,b4\n"
+            "0,true,yes,y\n"
+            "1,false,no,y\n"
+            "0,false,yes,n"
+        )
```
lwjohnst86, line 59: For instance, could you create a test that includes a row or values that are something like "true,false,no", to confirm that an error gets thrown when that happens.

philter87: That would actually work in my current implementation. It will only fail if is is not a boolean-ish value: for example: "true,false,valid" would fail. If we need to make sure that the entire column uses the same pattern, I will need to change the implementation. Should I do that? Or should just remove the boolean-ish feature?

philter87: I have added a test with (0, YEAH, 0) and (true, REALLY TRUE!, false) to assert that the columns are derived as strings

signekb (commented):
Very nice! Just have some questions and minor comments 🌻

app/csv_reader.py
```diff
@@ -0,0 +1,100 @@
...
+    Returns:
+        DataFrame: A Pandas DataFrame with the derived data types (`dtypes`).
+    """
+    dialect = _derive_csv_dialect(csv_file, row_number)
+
+    df = pandas.read_csv(csv_file,
+                         nrows=row_number,
+                         dialect=dialect,
+                         true_values=["yes", "Yes", "YES", "y"],
+                         false_values=["no", "No", "NO", "n"])
```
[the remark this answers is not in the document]
philter87: Yes, that is true. They will be recognized and stored as booleans

app/csv_reader.py
```diff
@@ -0,0 +1,100 @@
...
+        row_number (int, optional): Number of rows to scan. Defaults to 1000.
+
+    Raises:
+        csv.Error: if unable to parse CSV dialect
+
+    Returns:
+        Dialect: A Dialect object containing information on the separator for the columns in the CSV file.
+    """
+    dialect = csv.Sniffer().sniff(csv_file.read(row_number))
+    csv_file.seek(0)
```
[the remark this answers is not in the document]
philter87: The "file pointer" is moved back to the start of the file. Otherwise, when we use pandas.read_csv, then it would start from the middle of the file

Issue #91, opened 2024-01-15
Function to report progress of column metadata extraction


lwjohnst86:
Creating a function to be used for the progress bar showing how far Sprout is in extracting the csv column metadata. 
This function, therefore, relies on the function created in #17 that actually scans and extracts the metadata.

lwjohnst86:
Closed from #134

PR #134, opened 2024-02-05
Feat/progress bar for file upload

### LINKED ISSUE #90: Build html progress bar for after uploading file for check
Build html progress bar, as shown in wireframe below, after clicking "Extract data structure" on the page for uploading csv files (#74):

Image

To make this progress bar, we need to know how far we are in the process of extracting metadata (#91)

[beercss](https://www.beercss.com/):
`<progress value="0" max="100"></progress>`

signekb:
## Description

- This PR adds a progress bar after the user clicks "Extract Data Structure" at the file upload page. The progress bar is shown for two seconds just to show it off for now. Currently, the extraction progress is so fast that it would otherwise be shown for a very short time, which would be confusing (since you can't even read the title before it's gone again). The plan is to remove the two second wait, when the extraction is finalised.
    - Alternatively, a spinner could be added instead, if the extraction of column metadata is so fast that it doesn't make sense to have a progress bar.  
- I have left out a function that tracks how far we are in the extraction progress that should be shown in the progress bar for now, since the extraction isn't finished yet - and maybe we don't need it when the extraction process is pretty fast? 

Screenshot showing the progress bar:
image

## Related Issues

Closes #90. 

See also #91 (which I have omitted for now, and I'm not sure we need it). 

## Testing

- [ ] Yes

I don't think tests are needed for this. It's just showing the progress bar for two seconds before going to the next page. 

## Reviewer Focus

Discussion point: Should we omit the function to track how far we are in the extraction progress, when it doesn't take that long? Might this change/be slower later, so we would actually need to track the progress?  

## Checklist

- [X] Relevant documentation has been updated (nothing to update)

### DIFF
--- a/app/templates/file-drop-zone.html
+++ b/app/templates/file-drop-zone.html
@@ -48,6 +48,12 @@
   </div>
 </form>
 
+<!-- Progress bar dialog -->
+<dialog id="progress-dialog">
+  <h5>Extracting Column Metadata</h5>
+  <progress></progress>
+</dialog>
+
 <script>
   function query(id) {
     return document.querySelector("#" + id);
--- a/app/templates/file-drop-zone.html
+++ b/app/templates/file-drop-zone.html
@@ -42,7 +42,7 @@
   <p class="grey-text">Supported formats: .csv</p>
 
   <div class="row center-align">
-    <button disabled id="submit_button" type="submit" value="OK">
+    <button disabled id="submit_button" type="submit" value="OK" onclick="showDialog()">
       EXTRACT DATA STRUCTURE
     </button>
   </div>
@@ -61,6 +61,8 @@
 
   const highlightClass = "file-upload-box-highlight"
   const dragZone = query("drag_zone")
+  const dialog = document.getElementById("progress-dialog"); 
+
 
   function selectedFiles(files) {
     const isFilesListEmpty = files.length === 0;
@@ -94,4 +96,9 @@
     e.preventDefault()
     dragZone.classList.add(highlightClass)
   }, false)
+
+
+  function showDialog() { 
+  dialog.show(); 
+} 
 </script>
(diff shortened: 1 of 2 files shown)


app/templates/file-drop-zone.html
```diff
@@ -88,4 +96,9 @@ <h5 class="center-align bold">OR CLICK TO OPEN YOUR FILE EXPLORER</h5>
     e.preventDefault()
     dragZone.classList.add(highlightClass)
   }, false)
+
+
+  function showDialog() { 
+  dialog.show();
```
philter87, line 102: Maybe the statement "dialog=document.getElementById("progress-dialog")" should be added here instead, as dialog is only used here

app/views/file_upload.py
```diff
@@ -50,6 +51,9 @@ def handle_post_request_with_file(request: HttpRequest,
     except StopUpload as upload_error:
         return render_file_upload_page(request, table_id, upload_error.args[0])
 
+    # wait to sec (to show the progress bar)
+    time.sleep(2)
```
philter87, line 55: I still feel a bit weird about this one, but the plan was to remove it when the actual "extraction-column-logic" is in place, right? 

If it never takes above 0.5 seconds, then maybe we should consider adding a small spinner to the button instead.

But in either case, we can reuse this when we do the validation

philter87 (dismissed):
Very pretty loading bar :)!

app/templates/file-drop-zone.html
signekb: 👍

app/views/file_upload.py
philter87: Lets just keep it for now. We might also decide to actually upload the entire file to the server, and then it could potentially take a long time

philter87 (approved):

lwjohnst86 (approved):
Super neat, awesome!

PR #141, opened 2024-02-06
fix: :bug: A function was forgotten to be renamed after it was changed.

lwjohnst86:
There was a bug caused by a function that had been forgotten to be renamed. Fixed.


philter87 (approved):
Ups! Thanks :)

signekb (approved):
👍

Issue #56, opened 2024-01-09
Discussion: Which data types will we want in Sprout


K-Beicher:
Create a list of the data types we want to offer to the users in the drop down when creating a new table structure in Sprout.

K-Beicher:
Overall we'll want one of more values within the following categories:

Numeric
Date
Boolean
Character (text)

Optional would be:
JSON
XML

K-Beicher:
Numbers
In addition to the most common data type of Integer, we should have a value which will allow for precision calculation, which would either be the values Numeric (aka Decimal) or double precision (aka float8).  The difference is in the way some values are stored, compared and used in calculations.  It would be rare for people to use it, but if you are working with very small values it may be appreciated.

Date
There are three values typically associated with dates.  Date, time and Date+Time.  All three would be useful to have as options for users of Sprout, and all three are fully supported by Postgres as Date, Time and Timestamp.

Boolean
An option for a 0/1 value field should be included.

Text
There are a couple of options in postgres for text fields.  The best options in this case would be Varchar or Text.  Text is not a part of standard SQL, but many other systems have something similar.  A Varchar can be limited to a certain amount of characters, but without a limitation it can contain just over 10M characters.  A Varchar option would seem as the best solution for Sprout at this point in time.

K-Beicher:
JSON
There could be a need (maybe not a first) for an option that will allow users to store columns with JSON data.  Postgres offers two types of JSON data storage, json and jsonb.  If we are to offer a JSON option my suggestion would be the standard json format, as this is the format where the text a user uploads is also the text that is stored.  The second JSON format (jsonb) will optimise the text for quicker processing, which I don't think we will need to do in any significant way in Seedcase.

XML
The XML format in postgres will check for 'well-formedness' of the XML as it is being written to the database.  This may create some issues for the validation of data, which means that a standard text field would be a better option.  I think we should probably keep the XML format as a future possibility, but not implement it at present.

K-Beicher:
@lwjohnst86, @signekb, @philter87, @henrik-goul this is ready for you guys to read and comment on.

signekb:
I think that the data types you include as required here seem very reasonable!

lwjohnst86:
This I think has been done. Closing.

PR #145, opened 2024-02-07
feat: central location for ColumnDataTypes

### LINKED ISSUE #133: Discussion: Should we represent the data types as a table or as a choice (enum-like value)?
Currently, we store data types in the database table app_columndatatypes. We add new data types by adding new migrations. 
We could solve this in another way by using "choices" https://docs.djangoproject.com/en/5.0/ref/models/fields/#django.db.models.Field.choices

Pros and cons of using a "table" vs "choices"
1. **Pro for choices**. It is easy to update the data types, as we don't need a migration to do it
2. **Pro for choices**. The data types are defined a single place in the code and not in several potential migrations
3. **Con for choices**. We will lack data type information when we join tables in the database (when using raw sql and bypassing the Django ORM). If we create data type ids as strings, it would get some information about the individual types, but we would never be able to change these ids.
4. **Con for choices**. Foreign key constraints?

philter87:
## Description

This PR does two things
- It splits the models.py file into several files and puts them in the models folder. 
- The COLUMN_DATA_TYPES are applied after a 'migrate', and not as part of the migration. This will simplify changes and all types will be located at a single location (and not in scattered over several migration files)

## Related Issues

Closes #133

Related #56

## Testing

- [X] Yes

## Reviewer Focus

- The file column_data_type.py and apps.py are the most important files. 
- The "ready()"-method is triggered automatically by Django on application startup. I have added post_migration "signal", which is a way to execute some code after the migrations have been applied. 
- The migration file "0001_initial.py" is no longer adding ColumnDataType as this is handled in a different way
- Ignore the files in the folder "models" (except column_data_type.py)

## Checklist


lwjohnst86 (approved):
Had a look over (@philter87 also went over it with me today) and added some small docstring edits and other formatting fixes. Looks great 🤩

app/models/column_data_type.py
signekb, line 1: Nice! Just a small comment: The first two datatypes have examples of usage in their description. Should the others have that as well?

app/models/column_data_type.py
```diff
@@ -0,0 +1,67 @@
...
+    ),
+    ColumnDataType(
+        id=2,
+        display_name="Text",
+        description="A text field without a limit to how many characters can be stored",
+    ),
+    ColumnDataType(
+        id=3,
+        display_name="Yes/No",
+        description="Also known as Yes/No, Boolean, or Logical. This stores data as either 0/1 or True/False. If you need/want to store actual words, use the Text field",
```
signekb, line 38:
```suggestion
        description="Also known as Boolean, or Logical. This stores data as either 0/1 or True/False. If you need/want to store actual words, use the Text field",
```
"Yes/No also known as "Yes/No"

app/models/column_data_type.py
```diff
@@ -0,0 +1,67 @@
+from django.db import models
+
+
+class ColumnDataType(models.Model):
+    display_name = models.TextField()
+    description = models.TextField()
+
+
+"""
+COLUMN_DATA_TYPES contains all the "sprout" types.
```
signekb, line 10:
```suggestion
COLUMN_DATA_TYPES contains all the data types in Sprout.
```

app/models/column_data_type.py
```diff
@@ -0,0 +1,67 @@
...
+
+Changing the properties of the content can be risky. See the list below
```
signekb, line 12:
```suggestion
Be cautious of changing the properties of the content of COLUMN_DATA_TYPES. See the list below
```

app/models/column_metadata.py
```diff
@@ -0,0 +1,20 @@
...
+
+from app.models.column_data_type import ColumnDataType
+from app.models.table_metadata import TableMetadata
+
+
+class ColumnMetadata(models.Model):
+    """Object to create table with types of metadata from columns in imported data.
+
+    Args:
+        models: Values to store for metadata.
```
signekb, line 11: Not sure I understand this sentence? :)
We haven't names the arg `models` above, right? But just specified the type?

signekb (approved):
Very nice. I deleted my `db.sqlite3` and used the `just update-migration` command. It worked perfectly 👌  
I really like that we have moved the column datatypes away from the migrations!

signekb:
Ah, I approved and now it's merged. Sorry, about that @philter87

philter87:
> Ah, I approved and now it's merged. Sorry, about that @philter87

Okay, I will create a new PR with you suggested changes

PR #154, opened 2024-02-12
feat: read csv file with polars

philter87:
## Description

The third attempt to read and derive csv column types. I have implemented a method 'read_csv_file' which builds upon "polars.read_csv". The polars.read_csv is missing a few features, which I have implemented:
- A series with only 0 and 1 should be converted to a boolean series.
- A series with boolean-ish values ("Yes", "No") should be converted to a boolean series
- The polars.read_csv is unable to parse some CSV dialects, so the csv content is converted to a certain dialect before it is passed to polars.read_csv. This could possibly

- This PR adds a feature to read csv file and derive columns

## Related Issues

Closes #123

## Testing

- [X] Yes

## Reviewer Focus

- Should we support the boolean-ish values like "Yes" and "No"? Or should I remove it?
- polars.read_csv is unable to parse certain date formats automatically: e.g. "12. Jan 2024". Is this okay?
- The memory consumption is high, because we need to convert to a different CSV format and keep the converted format in memory. Alternative is to store in a temporary file and then delete afterwards

## Checklist


app/tests/tests_csv_reader.py
```diff
@@ -0,0 +1,186 @@
+"""File with tests for csv_read_file()."""
```
lwjohnst86, line 1:
```suggestion
"""File with tests for `csv_read_file()`."""
```

app/tests/tests_csv_reader.py
```diff
@@ -0,0 +1,186 @@
...
+        self.assert_values(df["s1"], 0, 1, 2)
+        self.assert_values(df["s2"], "true", "REALLY TRUE!", "false")
+
+    def test_datetime(self):
+        """Testing different date and time formats.
+
+        The columns (d1, d2, d3) should convert to dates and (t1) should
+        convert to time
+
+        NOTICE: d3 (ex: '27. Oct 1987') is not recognized as date
```
lwjohnst86, line 108: Yea, I think we'll have to figure out how to solve that problem, because there are dozens of ways people enter date data and we should be able to parse that.

app/tests/tests_csv_reader.py
```diff
@@ -0,0 +1,186 @@
...
+        """
+        csv_file = io.StringIO(
+            "s1,s2\n"
+            "Hello, World, Seedcase"
+        )
+
+        self.assertRaises(csv.Error, read_csv_file, csv_file)
+
+    def assert_types(self, df: DataFrame, *expected_types: str):
+        """A method test expected_types in a DataFrame object.
```
lwjohnst86, line 157:
```suggestion
        """A method test `expected_types` in a DataFrame object.
```

lwjohnst86 (changes requested):
Looks great! Just some small suggested changes, mostly formatting and things to look into in the future. But wonderful 🎉 🥳

app/csv_reader.py
```diff
@@ -0,0 +1,106 @@
+"""File with read_csv_file and related functions."""
+import csv
+import io
+from typing import Any, TextIO
+
+from polars import Boolean, DataFrame, Series, read_csv
+
+
+def read_csv_file(csv_file: TextIO, row_number: int = 1000) -> DataFrame:
```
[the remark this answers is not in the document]
philter87: This will mainly be a problem when we need to validate or parse the entire file, but there are other solutions. Either we will divide the file in chunks or we can create a temporary file to store the data (instead of in-memory)

app/csv_reader.py
```diff
@@ -0,0 +1,106 @@
...
+    Args:
+        df: Polars Dataframe with series
+
+    Returns:
+        DataFrame: A DataFrame with series converted to
+    """
+    boolean_mapping = {"yes": True, "Yes": True, "y": True, "Y": True,
+                       "no": False, "No": False, "n": False, "N": False}
+
+    for column in df.columns:
```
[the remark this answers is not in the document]
philter87: Valid point! Very observant!

And this was my initial implementation (provided by chatgpt), but I thought the implementation with selectors is less readable (even though it is short):

df = df.select([_convert_to_booleans(df[col]) for col in df.columns])

What do you think @signekb @K-Beicher?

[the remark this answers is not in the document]
philter87: Nevermind, I will just implement it with selectors

app/tests/tests_csv_reader.py
philter87: Should I try to fix in this PR?

app/csv_reader.py
```diff
@@ -0,0 +1,111 @@
...
+    "y": True,
+    "Y": True,
+    "no": False,
+    "No": False,
+    "n": False,
+    "N": False,
+}
+
+
+def _convert_to_booleans_if_possible(series: Series) -> Series:
```
signekb, line 69: Is it necessary to actually convert the column to booleans when we only want to extract the column metadata and not load the file itself at this point?

signekb (changes requested):
Looks good! Mainly have some suggestions/edits for docstrings :)

app/csv_reader.py
```diff
@@ -0,0 +1,111 @@
...
+    Returns:
+        TextIO: An in-memory CSV file which is suitable for `polars.read_csv()`
+    """
+    # Read part of csv_file to detect dialect and return to beginning of file
+    dialect = csv.Sniffer().sniff(csv_file.read(row_number))
+    csv_file.seek(0)
+
+    csv_content = ""
+    for line in csv.reader(csv_file, dialect=dialect):
+        csv_content = csv_content + ";".join(line) + "\n"
```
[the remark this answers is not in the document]
philter87: In this context I prioritized readability, but yes the code is not very efficient. And the memory consumption is an even bigger problem

app/csv_reader.py
philter87: My plan is to use the dtypes on the returned DataFrame to create the ColumnMetaData objects, but I could have just returned a dictionary of the column-to-type mapping instead of the DataFrame.

Furthermore, I hope that read_csv_file will be used for validation (and etc) as well, but lets see if it is possible.

signekb: Ahhhh, I see! Lets just keep it then :)

app/csv_reader.py
```diff
@@ -15,11 +15,12 @@ def read_csv_file(csv_file: TextIO, row_number: int = 500) -> DataFrame:
...
     - Finds CSV dialect of file and converts to a csv format suitable for
-      `polars.csv_read()`
+      `polars.csv_read()`, i.e., transform dialect to semicolon remove and
+      whitespaces and quotes
 
     Args:
-        csv_file: The CSV file to read
-        row_number: The number of rows to scan from the file
+        csv_file (TextIO): The CSV file to read
+        row_number (int, optional): The number of rows to scan from the file
```
signekb, line 24: Oh, sorry! The types shouldn't be added. I made this edit before my docstring deep dive yesterday and forgot to delete them again. Sorry!
Same for L77 and L96+97

lwjohnst86 (approved):
Wooohooo!! 🎉

app/csv_reader.py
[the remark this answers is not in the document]
lwjohnst86: I also tend to avoid using loops for similar reasons @signekb. One solution could be to detect the dialect and change the `separator` argument in the `read_csv()` to use the dialect there, rather than change the whole file itself. I also found the `scan_csv()` which might be useful for these purposes.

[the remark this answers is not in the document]
philter87: I also tried, but failed because there is more to a dialect. You can read more in the issue you created: #187

PR #173, opened 2024-02-13
fix: centre align icon in dialog and change it to "new window"

signekb:
## Description

- This PR centre aligns the icon in the create new table dialog. 
- It also changes the icon from "edit" to "new_window" 
    - I think it fits better with the "create new table" icon. (which is a "+").
    - It also ties nicely into the "Add data to existing table" icon, which is the same as "new window", but with "data" in it already. 
    - Could it be confusing since the new icon is called "new_window"? 
    - What do you think? ☺️

New window icon:
Screenshot 2024-02-13 at 11 17 15

Edit icon: 
Screenshot 2024-02-13 at 11 17 28
 

## Testing

- [ ] Yes

## Reviewer Focus

## Checklist

No tests nor documentation has been added. 
Didn't run linter, since this is solely edits in html.


philter87 (dismissed):

lwjohnst86:
Hmm, I moved the icon to be beside the text intentionally, because it feels a bit weird to be off on its own on the top. If you look at examples of Material Design dialog boxes (and Google's [own documentation](https://m2.material.io/components/dialogs#behavior) on it), I don't see the use of icons often, unless it is to draw attention.

In this case, I think we're only using it for aesthetics, which, if we follow Material Design, we shouldn't do. So I am leaning more towards removing the icon entirely since it doesn't add anything extra that the text doesn't already say. @henrik-goul thoughts?

signekb:
Ah, yeah, we could also just remove it! 
I just didn't like the look of it being on the same line as the header (sorry ☺️ )

K-Beicher:
I hadn't even noticed that there was an icon, so I guess I'm for removing it :-)

signekb:
@lwjohnst86 I have removed the icon now. I think that makes the most sense, actually :)

lwjohnst86 (approved):

PR #149, opened 2024-02-09
Feat/review column data

K-Beicher:
## Description

- This PR adds the view of the column meta data for Sprout.  It also adds the ability to edit the information and to save it back to the metadata table.  

## Related Issues

Closes #113, closes #81, closes #83, closes #71 

See also Issues #...

## Testing

- [x] Yes
- [ ] No, not needed (give a reason below)
- [ ] No, I need help writing them

## Reviewer Focus

This PR requires an in-depth review.

First time adding to Django so would appreciate if everything is checked to see if code is placed in the correct files.
 

I'm especially interested in comments on the testing and how the PR has been sorted, I'm not entirely sure I've done everything the right way.

## Checklist

For all PRs that are not general documentation

- [x] Tests accompany or reflect changes to the code
- [x] Tests ran and passed locally
- [x] Ran the linter and formatter
- [x] Build has passed locally
- [ ] Relevant documentation has been updated

PR #183, opened 2024-02-16
Feat/add original column name

signekb:
## Description

- This PR adds the original column name, `original_name` to the `ColumnMetadata` model and adds it to the `file_upload` view. 
- The user will be able to edit the name and title of the column (machine-friendly and human-readable names), but not the original column name. This ensures that we have a reference when users want to add data to an existing table. 
- To be able to add a new field in the `ColumnMetadata` model without a default, I deleted the  `0001_initial` migration and built it again.

## Reviewer Focus

- Have I missed somewhere where `original_name` needs to be added? I'm thinking that @K-Beicher will add this to the column review PR #149 
- I haven't added any tests for this. Should I? 

## Checklist

- [X] Tests accompany or reflect changes to the code (no tests added)


philter87 (approved):
Looks good! At some point, we might need to add default values, but I see why you wanted to redo the initial migration

signekb:
@philter87 Should we though? I asked @K-Beicher about this. If there's missing a header in the csv, wouldn't it make sense for us to throw an error?

lwjohnst86 (approved):
Simple enough, awesome!

philter87:
> @philter87 Should we though? I asked @K-Beicher about this. If there's missing a header in the csv, wouldn't it make sense for us to throw an error?

Yes, we should throw an error if a header is missing, but my point was something different.

In the future, we should not change migrations after they have already been deployed/applied, because Django only applies a migration once. So if our application was running somewhere for real, then this new column would not be added. 
Therefore we would need to either allow nulls or add default values somehow when adding columns. In this case, however, it would be a little difficult to fix existing rows (except we could use the maybe name or title as default values)

But for now it is perfectly okay. It was just a small note (which was a bit hard to understand)

signekb:
Ahh, yes, I see! Thanks for elaborating!

Issue #160, opened 2024-02-12
Investigate where raw data files will be stored (file path)


lwjohnst86:
How are raw data files saved? On Fly.io, it makes sense that they are saved on their servers. But what about for self-hosted installs? Where should they be saved? How does that work and how can users at least see them (but not edit)?

philter87:
In my current (very simple) implementation, the files are stored in the folder called 'persistent_storage'. 

This works well with:
- fly.io, because we can just mount a fly volume
- docker, because we can also just mount a volume
- locally, because it is just stored in the persistent_storage folder

**How to download?**
However, we cannot just serve these assets in the statics folder. We need to have more control. 

Therefore, we also persist some metadata for every file upload. Every file is tied to a table and the table defines who have access to its data.

PR #191, opened 2024-02-20
feat: persistent storage

philter87:
## Description

- This PR adds persistent storage to files located in persistent_storage. The sqlite file is now created here and eventually uploaded files will be created here. This will also be the case when doing local development, so we need to ignore these files when pushing to git (.gitignore) or when creating a docker image (.dockerignore)

This PR touches many files but the changes are small
- settings.py. PERSISTENT_STORAGE_PATH is defined
- fly.toml. We define a volume that preserves the content of persistent_storage. 
- docker-compose.yml is changed to use persistent_storage as volume (sqllite and CSV will be preserved docker-compose deployments)
- Two markdown files are added in persistent_storage and persistent_storage/raw to explain the purpose of the folders
- .dockerignore. We have added persistent_storage as we done want to include the content of persistent_storage to the docker image
- .gitignore. The content of persistent_storage should go to git (but we should allow markdown)
- docker-compose 

## Related Issues

See also Issues #160 

## Testing

- [ ] Yes

I have tested for seedcase-sprout.fly.dev and it seems to work, but it is a manual test

## Reviewer Focus

WARNING: In this implementation, the fly app is only allowed to have one machine! A volume cannot be shared between machines

## Checklist


lwjohnst86 (approved):
Not sure how much I intuitively like keeping the folders here in the repo vs having them installed on deployment/install, but I don't know enough at this point to have a strong opinion. So for now, very excited to see how this works in Fly! 🌠

K-Beicher (approved):
Looks good as far as I'm able to see

PR #195, opened 2024-02-20
build: :rewind: I accidentally changed this to the wrong value, should be app

lwjohnst86:
## Description

- This PR reverts a change I made in the `fly.toml` file, which made deployment break. The `processes` should be `"app"`, I had accidentally renamed it to `"sprout"` (thinking it related to the app folder.


K-Beicher (commented):
Happens ;-)

PR #199, opened 2024-02-23
feat: upload file to persistent storage

### LINKED ISSUE #161: Function to upload raw data file
where will the raw files be stored? See #160 for more details.

philter87:
## Description

- This PR adds persistence of a file to the persistent_storage folder on the server. This happens on the "file-upload" page after the file have been validated (I know that the actual data upload should have been done later, but this was to test it out and we might need this anyway). Every uploaded file should also create a FileMetadata object in the database. 

## Related Issues

Closes #161 

## Testing

- [X] Yes

## Reviewer Focus
- FileMetaData.persist_file(). The method persists a file and stores some metadata so we can find it again
- FileMetaData.delete(). I have also added a delete method, which is used in the tests, but will likely be used later as well

## Checklist


config/settings.py
```diff
@@ -84,7 +84,7 @@
 WSGI_APPLICATION = "config.wsgi.application"
 
 # Files are preserved in the "data" folder
-PERSISTENT_STORAGE_PATH = os.path.join(BASE_DIR, "persistent_storage")
+PERSISTENT_STORAGE_PATH = "persistent_storage"
```
lwjohnst86, line 87: Small question, why remove `os.path.join()`?

lwjohnst86 (dismissed):
Really cool!

config/settings.py
philter87: Yes, that is a very good question (it is usually advised to use os.path.join). 

When I store the files on the server, I also save FileMetaData with a property server_file_path. Part of the server_file_path is PERSISTENT_STORAGE_PATH. If I use os.path.join it looks weird on my Windows machine: "C:\\repos\\seedcase-sprout\\persistent_storage\\raw\\asdasdasd.csv", and it would look different on another machine. So instead I use "persistent_storage/raw/asdasdasd.csv".

Both things seems to work, but personally I liked the latter

signekb: So the difference is an absolute path vs a relative path? :)

sprout/migrations/0002_filemetadata.py
signekb, line 1: Will there be a conflict/an issue when this is merged bc there already are a 0002 migration on main?

sprout/models/file_metadata.py
```diff
@@ -0,0 +1,75 @@
...
+class FileMetaData(models.Model):
+    """Model for a persisted file."""
+
+    original_file_name = models.TextField()
+    server_file_path = models.TextField()
+    file_extension = models.CharField(max_length=10)
+    file_size_bytes = models.BigIntegerField()
+
+    table_metadata = models.ForeignKey(TableMetadata, on_delete=models.CASCADE)
+    created_at = models.DateTimeField(auto_now_add=True)
```
signekb, line 21: do we want a `created_by` here?

sprout/models/file_metadata.py
```diff
@@ -0,0 +1,75 @@
...
+            FileMetaData: The relative path on the server
+        """
+        file_extension = file.name.split(".")[-1]
+
+        raw_folder = f"{PERSISTENT_STORAGE_PATH}/raw"
+        if not os.path.exists(raw_folder):
+            os.makedirs(raw_folder)
+
+        # Unique file path in the raw
+        server_file_path = f"{raw_folder}/{uuid.uuid4().hex}.{file_extension}"
```
signekb, line 42: would it be more intuitive to include the original file name in `server_file_path` so the csv name in the `persistent_storage/raw` folder still contains the original file name? 
Otherwise, the user could lose track of which files are in the raw folder? Not sure whether they would go into this folder, but in case they do?

sprout/tests/tests_models_file_metadata.py
```diff
@@ -0,0 +1,51 @@
...
+        file_meta = FileMetaData.persist_raw_file(file, test_table.id)
+
+        # Act
+        file_meta.delete()
+
+        # Assert
+        self.assertEqual(FileMetaData.objects.count(), 0)
+        self.assertFalse(os.path.exists(file_meta.server_file_path))
+        # And we change that the table is not deleted
+        self.assertEqual(TableMetadata.objects.count(), 1)
```
signekb, line 51: So when `FileMetadata` is deleted, the raw file is deleted too, but the `TableMetadata` should still be there? Why? :)

signekb (commented):
Nice! Just have some questions, before I will approve :)

config/settings.py
philter87: Yes, and I don't want to store OS dependent path in the database

sprout/models/file_metadata.py
philter87: Yes, I can add that

sprout/models/file_metadata.py
philter87: It is to avoid naming conflicts. The same filename could be uploaded multiple times. But I could preserve the original name and then append some random Id at the end. I will do that.

The user will never see this name, but for debugging purposes it might make sense

sprout/models/file_metadata.py
signekb: @K-Beicher what say you? :)

sprout/models/file_metadata.py
signekb: Yess, I think I got that during the status meeting today as well. But nice! Thanks!

sprout/tests/tests_models_file_metadata.py
signekb: Ah, yessss, you are completely right! I had the feeling that I was missing something. Thanks! 

But what if there's only one file referring to a table? Should the deletion of that file then delete the `TableMetadata`? 🤔

philter87: Maybe, but from my perspective it is a unexpected side effect that the schema is deleted when the last file is deleted. But maybe @lwjohnst86 has an opinion?

PR #200, opened 2024-02-26
Feat/file download example

### LINKED ISSUE #160: Investigate where raw data files will be stored (file path)

philter87:
## Description

File overview for a table and download links, see picture below. You can download by clicking the title. Eventually, we can add some logic in "table_file_download()" that will allow/prevent a user from downloading

- This PR adds overview of files for a given table. The overview is based on FileMetaData objects in the database and not just a folder of files. This will allow us to prevent users from downloading files if they are not allowed to

image

## Related Issues

Closes #160.

## Testing

- [X] Yes
- [] No, not needed (give a reason below)

## Reviewer Focus

- The url "table-files/<table_id>" renders the html "table-files.html". It displays a list/table of files for a table_id
- The url "table-files/<table_id>/download/<file_id>" download the relevant csv file. Eventually, we can put some logic in "table_files.table_file_download()" to prevent unauthorized users from downloading a file

## Checklist


K-Beicher (dismissed):
Looks good

sprout/templates/table-files.html
```diff
@@ -0,0 +1,32 @@
+{% extends 'base.html' %}
+
+{% block content %}
+  <h2>Files for&nbsp;<b>{{ table.name }}</b></h2>
+  <p>Here you can view and download the files relevant for a certain table.</p>
```
lwjohnst86, line 5:
```suggestion
  <p>Here you can view and download the files relevant for this table.</p>
```

sprout/tests/tests_views_table_files.py
```diff
@@ -0,0 +1,29 @@
...
+class TableFilesTests(TestCase):
+    """Tests for table_files."""
+
+    def test_overview_should_show_files_for_table(self):
+        """The rendered overview should show file.csv on the page."""
+        # Arrange
+        table = create_table("TestTable")
+        table.save()
+        file = io.BytesIO(b"File content")
+        file.name = "file.csv"
```
lwjohnst86, line 20: any particular reason this uses the `.` in the name? Or you are setting the `name` property of `file` to be `file.csv`?

sprout/views/table_files.py
```diff
@@ -0,0 +1,40 @@
...
+    Args:
+        request: The request from the client
+        table_id: Table id
+        file_id: The id of the file to download
+
+    Returns:
+        FileResponse: A file download response
+    """
+    file_metadata = FileMetadata.objects.get(id=file_id, table_metadata_id=table_id)
+    print(file_metadata.server_file_path)
```
lwjohnst86, line 38: Should this print statement be here?

lwjohnst86 (changes requested):
Only a small edit fix. Very neat! Not sure if we should provide a way to download files, but being able to view what files are stored might be useful, so I am fine keeping this as is for now. How might they get to this page though?

sprout/views/table_files.py
philter87: No, thanks :)

sprout/tests/tests_views_table_files.py
philter87: It is the file name, and I am expected a file name with an extension. Is it wrong to assume an extension to a file? In the upload view, I am also enforcing a ".csv" extension

philter87:
> Only a small edit fix. Very neat! Not sure if we should provide a way to download files, but being able to view what files are stored might be useful, so I am fine keeping this as is for now. How might they get to this page though?

I thought it could be part of 2.1 (figma view 2.1) or the expanded view (figma view 2.2)

And the idea was also to illustrate a simple version of seedcase, where people "just" share CSVs (and maybe validated parquet files with the potential of converting to CSVs on the fly). 

But it seems we are heading towards storing the data in a database, and then the download part is less useful.

sprout/tests/tests_views_table_files.py
lwjohnst86: No, not at all. I didn't understand what the `.` did in the `file.name`, but I understand now ☺️

PR #207, opened 2024-02-28
feat: read_csv uses file path and ColumnMetaData created

### LINKED ISSUE #89: Function to check if file is a valid file (in this case csv)

### LINKED ISSUE #112: Build function to analyse/extract column data types

philter87:
## Description

- This PR add the feature to derive and persist ColumnMetaData based on a CSV. Instead of using a file-object, we use a file path to save memory (and potentially use scan_csv)

## Related Issues

Closes #112, Closes #89 

## Testing

- [X] Yes

## Reviewer Focus
- The main thing in file_upload.py is that "read_csv_file()" is called. It creates a DataFrame which is used to create and persist ColumnMetaData objects.
- I changed the method "read_csv_file" to use a file path instead of a file object. This makes the PR a ugly as a lot of tests had to be changed
- The internals of "read_csv_file" is changed, so polars is used to remove whitespaces and quotes. 
- I created tests_views_file_upload.py to isolate tests for file_upload view.

## Checklist


signekb (commented):
Nice! 
When I run it locally, it get the error `django.db.utils.OperationalError: table sprout_filemetadata has no column named created_by_id` after I upload a test.csv?

sprout/tests/tests_views.py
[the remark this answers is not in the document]
signekb: Makes sense. I'll make a PR with that 👍

K-Beicher (dismissed):
Looks like it's been tidied.

Just curious, why has it changed the csv_reader.py from a tidy version to a less tidy one?

philter87:
> Looks like it's been tidied.
> 
> Just curious, why has it changed the csv_reader.py from a tidy version to a less tidy one?

I am not sure I understand, so now I might say a lot of random things. The polars.read_csv only understands certain CSV formats, so I am trying to transform to a suitable format. 

The transformation was done with a for-loop before, but now it is done in a "polars-way". I think this is faster

I also changed from a file object to using file paths. File objects might require a lot of memory (if everything is moved to memory), so it is better to use disk space

philter87:
> Nice! When I run it locally, it get the error `django.db.utils.OperationalError: table sprout_filemetadata has no column named created_by_id` after I upload a test.csv?

It should work now

sprout/csv_reader.py
```diff
@@ -1,74 +1,77 @@
...
 
-    csv_content = ""
-    for line in csv.reader(csv_file, dialect=dialect):
-        csv_content = csv_content + ";".join(line) + "\n"
-    return io.StringIO(csv_content)
+    df = df.select(pl.all().str.strip_chars())
+    df = df.select(pl.all().name.map(lambda n: n.strip().strip('"')))
+    df = df.select(pl.all().str.strip_chars('"'))
+    cleaned_path = csv_path + "cleaned"
+    df.write_csv(cleaned_path)
```
lwjohnst86, line 73: Niiiiccceee, love it!

sprout/views/file_upload.py
```diff
@@ -86,73 +87,37 @@ def render_file_upload_page(
...
-    altered_string = sub("[^a-zA-Z0-9\s_-]+", "", altered_string)
-
-    # Replace spaces and hyphens with underscores
-    altered_string = sub(r"[\s-]+", "_", altered_string)
+    df = read_csv_file(file.server_file_path)
 
-    # Convert camelCase to snake_case
-    altered_string = sub(r"([a-z0-9])([A-Z])", r"\1_\2", altered_string)
+    if len(df) == 0:
+        raise StopUpload("Unable to extract column types. No rows where found")
```
lwjohnst86, line 114:
```suggestion
        raise StopUpload("Unable to extract column types. No rows were found.")
```

lwjohnst86 (changes requested):
I like the polars-way change! Only a small change.

philter87 (commented):
Spelling mistake fixed

lwjohnst86 (approved):

Issue #196, opened 2024-02-20
Discussion: Naming scheme for HTML files and paths within app


lwjohnst86:
I don't feel we have a strong scheme for how we name the Django HTML templates. What are some good naming schemes here? How should the file path "API" be structured. This I think should have a design to it as well.

signekb:
I like it! With sprout instead of import 👌

K-Beicher:
I like the sprout version as well.

K-Beicher:
So probably go with the /sprout/ solution.  @lwjohnst86 Should we convert this into a decision writing issue, and/or implement it?

signekb:
Maybe we should actually change `column_review`to `review_columns` so it fits with the rest with verbs first, like `derive_columns` and `upload_data`? Aligning with that, should `existing_tables` be `review_existing_tables`, or am I taking it too far?

Alternatives could be to lose the verb like so: 
- **derive-table-columns** -->  **column-extraction**
- **column-review**** --> **column-review** (same)
- **existing-tables** --> **existing-tables-review**
- **upload-data** --> **data-upload**

K-Beicher:
I suspect we'd be grateful down the line if we keep the option for a verb, I'd be happy to change mine to comply and use the 'verb first' approach going forward.

philter87:
I actually do like the "verb-first" - I don't know why 🤔

signekb:
Alrighty. I'm fine with either, as long as we are consistent 🚀

lwjohnst86:
Thanks for all the input! I've been thinking about this a lot too, and going with the "verb" approach and in line with CRUD (Create Read Update Delete) of the core "language" of connecting with databases as well as the equivalent HTTP verbs (GET, POST, PUT, PATCH, and DELETE), I thought of a couple of "actions" and "objects" we have/work with:

- Objects: data and metadata
- Actions: create, view (or get), edit, upload

So something like:

- `/upload/data/` and `/upload/metadata/` for landing pages, and `/upload/data/<table_id>` and `/upload/metadata/<table_id>` for individual tables that already exist
- `/create/data/` and `/create/metadata/`
- `/view/metadata/` (to show list of all tables to select) and `/view/metadata/<table_id>`
- `/edit/data/<table_id>` and `/edit/metadata/<table_id>`

Thoughts?

philter87:
`/projects` shows all projects for the user
`/projects/<id>/tables` shows all tables for a project
[...]

PR #220, opened 2024-03-01
Feat: Add utility function to convert original column names to snake case

### LINKED ISSUE #159: Function to convert extracted metadata column names into appropriate snake_case or similar
During the metadata extract phase, having an (default) option to convert column names to `snake_case` (one of the best for readability) would be great. So that we can enforce/encourage following better practices.

signekb:
## Description

- This PR adds the feature of converting original column names to snake case and adding them into the `ColumnMetaData` field `name`.
- Additionally, it adds tests of the new function `_convert_to_snake_case`

## Related Issues

Closes #159

## Testing

- [X] Yes

## Reviewer Focus

I am not sure about the location of the `_convert_to_snake_case` function. Do you have any inputs? 

## Checklist

- [X] Relevant documentation has been updated (no documentation updated here - but see #user flow


sprout/views/file_upload.py
lwjohnst86, line 1: Very nice! A potential place to put the function might be in a set of `tidy_`, `convert_`, or `clean_` file names? So for this could be `clean_column_names.py`? Though I'm not so certain. The hardest things in programming... naming things.. 😆

lwjohnst86 (dismissed):
Very nice! I only have a small comment about the location of the function, but it isn't vital. We can very easily move it around, especially based on the related discussion in #196

sprout/views/file_upload.py
signekb: Yes, I also mentioned it in the PR that I'm unsure of the location of this function. 
I was considering a `utils` module within `sprout`, but `utils` might be too vague?

lwjohnst86 (approved):
Fixed the test case.

PR #223, opened 2024-03-01
Docs: remove option to add column during table completeness validation

signekb:
## Description

- This PR removes the option to add columns during the validation check "table completeness" in the user flow diagrams. 
- These changes are done based on feedback from @K-Beicher. Having the option to add a column at this point will complicate the field `original_column_name`, since we currently want this field to be non-blank and extracted from a file (what should the `original_column_name` be for an added column?). Additionally, the (some of this) functionality will be covered in the option to edit existing tables. 

Thoughts?


lwjohnst86 (approved):
Hm, I don't even recall reading that part. Good idea in removing it!

PR #232, opened 2024-03-05
Fix/table name suggestions

### LINKED ISSUE #231: [bug]: Rename "name" to "table name" in Create New Table dialog

philter87:
## Description

- This PR fixes a bug. We don't want any suggestions when finding a name for a table (before it would show many different names)

This solution is a bit weird, but Chrome is acting weird on Windows. You need to add the attribute 'autocomplete: "new-password"'. You can read more here (but you really don't need to): https://stackoverflow.com/questions/43132693/how-to-turn-off-html-input-form-field-suggestions

## Related Issues

Closes #231 

## Testing

- [ ] Yes

Was tested manually on Chrome Windows. 
We should also test on:
- Safari Mac
- Chrome Mac
- Chrome Linux
- Firefox Linux

## Reviewer Focus

## Checklist

### DIFF
--- a/sprout/forms.py
+++ b/sprout/forms.py
@@ -22,7 +22,7 @@ class TableMetadataForm(ModelForm):
 
         # Adding 'autocomplete: new-password' to disable suggestions on input field
         widgets = {
-            'name': Textarea(attrs={"autocomplete": "new-password"}),
+            "name": Textarea(attrs={"autocomplete": "new-password"}),
         }
 
     def clean_name(self) -> str:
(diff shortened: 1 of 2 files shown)


signekb (dismissed):
Nice! Yeah, it's a weird fix. But it seems to solve the issue on Mac Chrome and Safari 👍

sprout/forms.py
```diff
@@ -20,6 +20,11 @@ class Meta:
         model = TableMetadata
         fields = ["name", "description"]
 
+        # Adding 'autocomplete: new-password' to disable suggestions on input field
```
lwjohnst86, line 23:
```suggestion
        # Adding 'autocomplete: new-password' to disable suggestions on input field
        # TODO: Look into other solutions for this? (later)
```

lwjohnst86 (approved):
Very odd... But great to see this fix! I added a TODO item to it to remind us to look into another solution later (it feels weird to have to set a thing related to passwords here). But LOW priority.

PR #236, opened 2024-03-05
feat: change from ID to Original name

### LINKED ISSUE #211: Change rowID to Original Name in column-review.html
The row ID should be removed, and the Original name should be shown in the table overview in column-review.  

NB NB NB
The original name should be the only column that can't be edited.

K-Beicher:
## Description

- This PR adds original name and removes ID from the page column-review.
- These changes are done because we wish to offer more clarity to the user, the fact that the row ID was incrementing for the table itself meant that it could start at 57, which we feel is confusing.

## Related Issues

Closes #211 

## Testing

- [x] Yes

## Reviewer Focus

There are two changes, first the easy change from ID to original_name, and second the change to prevent editing of the original_name.  Especially the last item would be good to have a look at.

## Checklist


K-Beicher:
@philter87 I'm running into an error again setting up the correct values to enter for the test.  Could you have a look at the test, please? I'm away until Monday, so nothing urgent.

K-Beicher:
The tests are now running as expected, so this is ready to be looked at.

philter87 (approved):
Looks good :) 

I did the weird part with the missing original_name. But maybe another solution is to not include original_name it in the form 🤔 

Maybe I will take a look again, when it is merged in.

lwjohnst86 (approved):
Very nice 🙌🏻

PR #256, opened 2024-03-11
fix: save method is overridden and a test is added (SMALL)

### LINKED ISSUE #122: [bug]: should modified_at in TableMetadata be null until it is modified?
### Steps to reproduce

In `models.py`, the field `modified_at` is set to `auto_now=True`. I would argue that this field should be Null until it is actually modified. 

### Expected result

I would expect `modified_at` to be Null when it hasn't been modified, but only created.

### Actual result (the bug)

Currently, when you create a new row in `TableMetadata`, `modified_at` is equal to `created_at`. 

_No response_

### Severity/Priority

3: Minor

### Contact Details

_No response_

### Code of Conduct

philter87:
## Description

- This PR fixes the bug that modified_at is not null when created but equal to created

This is a very small pr.

## Related Issues

Closes #122 

## Testing

- [X] Yes

## Reviewer Focus

## Checklist

### DIFF
--- a/sprout/models/table_metadata.py
+++ b/sprout/models/table_metadata.py
@@ -39,5 +39,3 @@ class TableMetadata(models.Model):
             self.modified_at = datetime.datetime.now(datetime.UTC)
 
         super().save(*args, **kwargs)
-
-
--- a/sprout/models/table_metadata.py
+++ b/sprout/models/table_metadata.py
@@ -1,4 +1,6 @@
 """Module defining the TableMetaData model."""
+import datetime
+
 from django.conf import settings
 from django.db import models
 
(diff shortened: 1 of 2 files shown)


signekb (commented):
Hmmm, I think it was me that pushed this issue forward. I think it makes the most sense that `modified_at` is `NULL` when the table has not been modified, but only created. However, I can see the logic in having `modified_at` = `created_at` when it hasn't been modified after creation. 
In your PR description, it sounds like it's something we should avoid. Can you elaborate?

philter87:
> Hmmm, I think it was me that pushed this issue forward. I think it makes the most sense that `modified_at` is `NULL` when the table has not been modified, but only created. However, I can see the logic in having `modified_at` = `created_at` when it hasn't been modified after creation. In your PR description, it sounds like it's something we should avoid. Can you elaborate?

I now realize that the capital letters in the description could be interpreted as "angry", undskyld. It was just to highlight that the code should not necessarily be merged in. 

It is only a few lines of code, so if you prefer it to be NULL then we should just merge it in

K-Beicher (approved):
Seems sensible.

lwjohnst86 (approved):

Issue #224, opened 2024-03-02
Discussion: Naming practices of functions


lwjohnst86:
The old saying that "the hardest thing in programming is naming things"... which I definitely feel is true! I don't know about the others, but it would help me if we had some type of structure or "guide" on how we name functions and how we name files and folders. This is related to #196 but for Python code.

signekb:
This is tied to the discussion of #196 and closed with that.

lwjohnst86:
Reviewing the code more and thinking about this more, this isn't actually closed by #196, though it is a part of it. I'm going to reopen, change the title of the issue, and maybe split this out into other issues, for functions, for files, and for folders.

Issue #171, opened 2024-02-13
Develop the upload page for wireframes 2.2.1-2.2.3


lwjohnst86:
Wireframes 2.2.1-2.2.3 are very similar and could probably be implemented at the same time.

PR #259, opened 2024-03-11
refactor: :recycle: Updated naming to use `metadata/create/` instead of `file-upload`

lwjohnst86:
## Description

This PR refactors the code, paths, and names to follow the `metadata/create/` style rather than `file-upload`.

I made these changes because as I was digging into #171, I was getting anxious and stressed because there are already files and functions called `file-upload`, but that don't actually upload the data and rather are used for extracting the metadata. It is part of the reason I started the issues #196 and #224. I get very confused when trying to reuse existing functions (that I think are to upload file), but do a lot of secondary things too (like creating and storing metadata). 

So this PR starts changing the names to match what they actually are doing internally, and so I can create HTML templates for the data upload part.

I think we'll also need to split out some of the functionality of the renamed `metadata_create` function to focus only on extracting the metadata from the file, and have another function that uploads and saves the file (which I will need for #171). But I'll do another PR for that.

## Related Issues

Related to #196 and #224

## Testing

- [x] Yes

## Reviewer Focus

## Checklist


philter87:
I understand the confusion, and maybe it originates from more than just the name. 

1. The metadata for the table is created in a dialog on the previous page and then the metadata for the columns/properties are created on this page based on the uploaded file. Maybe we should move the input fields with table name and table description to this page.
2. And yes the file-dropzone.html part is confusing. It calls the url where it is included

We should quickly let you merge it in, so you can relax a bit again

philter87 (approved):

signekb (approved):
👍

PR #275, opened 2024-03-14
Feat/tabindex added to description

### LINKED ISSUE #274: On column-review page, User should automatically go to the next description field when hitting 'tab'

philter87:
## Description

- This PR adds a tabindex to the description fields. When the user hits 'tab' the cursor moves between description fields.

## Related Issues

Closes #274

## Testing

- [ ] Yes

I have only done manual testing.

## Reviewer Focus

This is a quick review, but if you like you can read about django custom templatetags, which is a way to add python code to django templates. See more here: https://docs.djangoproject.com/en/4.2/howto/custom-template-tags/

## Checklist

### DIFF
--- a/sprout/templates/column-review-list.html
+++ b/sprout/templates/column-review-list.html
@@ -12,10 +12,10 @@
 
   <h5>Extracted field name: {{ form.instance.original_name }}</h5>
   <article class="border">
-    <p>Column name: {{ form.name }}</p>
-    <p>Display name: {{ form.title }}</p>
-    <p>Description: {{ form.description|tabindex:forloop.counter }}</p>
-    <p>Data Type: {{ form.data_type }}</p>
+    <p>Column name: {{ form.name|tabindex:1 }}</p>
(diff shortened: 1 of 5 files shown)


signekb (commented):
Nice! 🎉 
Only thing: If I'm in another field than description and I use tab, `description` is skipped, which might be confusing?

philter87:
> Nice! 🎉 Only thing: If I'm in another field than description and I use tab, `description` is skipped, which might be confusing?

Nice spotted! I noticed as well, but I do not have a great fix. 

So I guess @lwjohnst86 needs to decide if this is a "keeper"?

(I could also try some weird javascript and decide the next field based on the previous field)

lwjohnst86:
@philter87 hmm, if you could try to implement it so it tabs correctly, that would be great! If it's too much work, maybe we create an issue to deal with it later.

lwjohnst86 (approved):
Super neat!!

philter87:
I changed it a bit, so it works in the same way for all of the fields

PR #280, opened 2024-03-15
feat: show tooltips for headers

### LINKED ISSUE #204: Add (?) buttons to explain each field in the column review table
I just realised that the column_review page is missing the (?) at each field, see wireframe below.

Image

I have assigned it to you @K-Beicher, since you created the page and made it a separate issue to avoid that #149 gets larger. But if you would prefer to work something else by now, you can just assign it to me.

K-Beicher:
## Description

- This PR add a tooltip button to the column review page.
- These changes are done because we need a place to tell users about the different headings and how they should be used.  We opted for one central button for two reasons, first it is nice only to have to click once, second, it hasn't been possible to get the button to work if added to the individual heading.

## Related Issues

Closes #204 

## Testing

- [x] Yes

## Reviewer Focus

The functionality has been lifted from the data-import page without any major modification.

## Checklist

### DIFF
--- a/sprout/static/css/style.css
+++ b/sprout/static/css/style.css
@@ -11,3 +11,11 @@
   width: 400px;
   height: 400px;
 }
+
+dt {
+  font-weight: bold;
+}
+
+dd {
+  margin: 0;
+}
--- a/sprout/templates/column-review.html
+++ b/sprout/templates/column-review.html
@@ -3,20 +3,21 @@
 
 {% block content %}
 
-<h2>Headers and Definitions extracted for {{ table_metadata.name }}</h2>
-
+<h2>Headers and Definitions extracted for {{ table_metadata.name }}
+  <button class="circle small fill" data-ui="#help-ext_name"><i>question_mark</i></button>
+</h2>
 <form method="post" action="">
   {% csrf_token %}
   <table>
     <thead>
       <tr>
-        <th>Extracted Name</th>
+        <th>Extracted Name
+
+        </th>
         <th>Column Name</th>
         <th>Display Title</th>
         <th>Description</th>
         <th>Data Type</th>
-        <th>Allow Missing</th>
-        <th>Allow Duplicate</th>
       </tr>
     </thead>
     <tbody>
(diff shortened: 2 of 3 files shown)


philter87 (dismissed):
Yay!

sprout/static/css/style.css
```diff
@@ -11,3 +11,11 @@
   width: 400px;
   height: 400px;
 }
+
+dt {
```
lwjohnst86, line 15: @philter87 why does this not change the `dt` style in the HTML output? I tried a few different approaches, but the style of dt in the dialog box doesn't change.

sprout/templates/column-review.html
```diff
@@ -2,43 +2,58 @@
...
+            <td>{{ form.title }}</td>
+            <td style="overflow:hidden;">{{ form.description }}</td>
+            <td>{{ form.data_type }}</td>
+            <input type="hidden" name="{{ form.prefix }}-id" value="{{ form.instance.id }}" />
+          </tr>
+        {% endfor %}
+      </tbody>
+    </table>
+    <button type="submit">Update All</button>
+  </form>
```
lwjohnst86, line 35: This was just reformatted with `prettier` (vscode plugin I am trying out)

sprout/templates/column-review.html
```diff
@@ -2,43 +2,58 @@
...
-</tbody>
-</table>
-</form>
-
-{% endblock content %}
\ No newline at end of file
+  <!-- Help dialog -->
+  <dialog class="dialog" id="help-ext_name">
+    <h5 class="center-align">Fields</h5>
+    <dl>
```
lwjohnst86, line 40: I converted to definition lists, since that is what it should be structured as (easier to style these). Right now doesn't look too pretty but with the CSS style (if it works) it should look nicer...

sprout/static/css/style.css
philter87: I tried to run it locally. It works on my computer (but maybe I misunderstood your question)
image

philter87 (dismissed):

sprout/static/css/style.css
lwjohnst86: hmmmmmm, I wonder if it is a firefox thing again... I'll compare with another browser.

lwjohnst86 (approved):

PR #282, opened 2024-03-15
docs: :memo: A draft of the naming scheme

### LINKED ISSUE #196: Discussion: Naming scheme for HTML files and paths within app

lwjohnst86:
## Description

- This PR adds documentation on how we name things. This is based on our extended discussions in #196.
- These changes are done because of having a strong naming scheme will help us develop as well as help the user navigate Sprout.

## Related Issues

Closes #196

## Reviewer Focus

This PR only needs a quick review, since we already discussed and decided on this during our meeting.

### DIFF
--- a/docs/design/naming.qmd
+++ b/docs/design/naming.qmd
@@ -18,16 +18,16 @@ How we name things, both internally and user-facing, will be:
 ## Composing names
 
 We'll compose names based on the objects we and our users interact with as well
-as the actions taken on those objects. These objects and actions are inspired by 
-the "core" language of working with data called *CRUD* (Create Read Update Delete) as
+as the actions taken on those objects. These objects and actions are inspired by and align with
+the "core" language of working with data called *CRUD* (Create Read Update Delete), common naming schemes used in REST API applications or services (such as Github), as
 well as by the vocabulary of the HTTP verbs (GET, POST, PUT, PATCH, and DELETE):
 
 - Types of objects Sprout interacts with: projects, data, and metadata.
-- Types of actions Sprout takes on those objects: create, view (or list), edit (or update), delete, and upload.
-- Types of identifiers for specific items of objects (as numbers or text): data, metadata, and project identifiers.
+- Types of actions Sprout takes on those objects: create, view, edit, delete, and upload.
+- Types of identifiers for specific items of objects (as numbers): data, metadata, and project identifiers.
 
 To name things, we'll combine the above with the object name first, followed by the
-action, with a natural heirarchy and "ownership" to the objects. Based on these
+action, with a natural heirarchy of "offspring" to the objects. Based on these
 principles, we can derive a naming scheme. In the scheme, objects and actions 
 are ordered, with names occurring first acting as a "parent" to later
 names. The names are separated by a symbol based on its context:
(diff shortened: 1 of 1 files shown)


philter87:
### Remove <id> on "create" urls
When we create a project or "data", we don't know the <id>, because it is not created yet. 

### Projects metadata
I think we should skip the 'metadata' part in projects, so we just have:
- projects <id> view
- projects <id> edit
- projects <id> create
- projects <id> upload

(This is also based on the part of equivalent views in you last section)

### Skip the 'view'
(I had a longer comment until I saw the last section with equivalent urls)

I guess, you like the consistency with an action at the end, but I prefer without

docs/design/naming.qmd
```diff
@@ -0,0 +1,108 @@
...
+
+## Composing names
+
+We'll compose names based on the objects we and our users interact with as well
+as the actions taken on those objects. These objects and actions are inspired from 
+the "core" language of working with data (CRUD: Create Read Update Delete) as
+well as with HTTP verbs (GET, POST, PUT, PATCH, and DELETE).
+
+- Types of objects Sprout interacts with: projects, data, and metadata.
+- Types of actions Sprout takes on those objects: create, view (or list), edit (or update), delete (?), and upload.
```
signekb, line 26: I still have some reservations about using the word `view`. However, I think it is more intuitive than the other words we have discussed (like overview, show, or list), so maybe we should just give it a go and see if we (or maybe mostly I) get confused. 

I think it would make sense to decide on either edit *or* update. I have no preference, but it seems like you're using edit further down in the example section?

docs/design/naming.qmd
```diff
@@ -0,0 +1,108 @@
...
+
+# View all data objects of a specific project
+projects <id> data view
+
+# View, create, edit, or upload metadata of a specific data 
+# object in a specific project
+projects <id> data <id> metadata view
+projects <id> data <id> metadata create
+projects <id> data <id> metadata edit
+projects <id> data <id> metadata upload
```
signekb, line 64: I'm just thinking of which pages these would be equivalent to: 
- create:  ?
- edit: "column-review" / wireframes 1.5 - 1.12?
- upload: "file-upload" (with dropzone) / wireframes 1.2-1.4?

docs/design/naming.qmd
```diff
@@ -0,0 +1,108 @@
...
+
+# View landing page of a specific data object in 
+# a specific project
```
signekb, line 67: What kind of "landing page" would this be? A view showing what's in the "table"/"data" currently?

docs/design/naming.qmd
```diff
@@ -0,0 +1,108 @@
...
+projects <id> data <id> view
+
+# Create, edit, or upload data of a specific data 
+# object in a specific project
+projects <id> data <id> edit
```
signekb, line 72:
```suggestion
projects <id> data <id> create
projects <id> data <id> edit
```

docs/design/naming.qmd
```diff
@@ -0,0 +1,108 @@
...
+projects <id> data <id> upload
+```
+
+Where `<id>` is a placeholder for a number or text for the identifier of it's
```
signekb, line 76:
```suggestion
Where `<id>` is a placeholder for a number for the identifier of it's
```
If you agree with only using numbers, following GitHub, as suggested above :)

docs/design/naming.qmd
```diff
@@ -0,0 +1,108 @@
...
+parent object. `view` could be the default action when an object is given
```
signekb, line 77: If view is default, would that mean that we would actually never use it explicitly in the naming scheme? That might be nice :)

docs/design/naming.qmd
```diff
@@ -0,0 +1,108 @@
...
+projects <id> create
+projects <id> metadata create
+
+# These could be equivalent
+projects <id> edit
+projects <id> metadata edit
+
+# These could be equivalent
+projects <id> data <id> view
+projects <id> data <id> metadata view
```
signekb, line 103: Not sure how these are equivalent? Wouldn't there be a difference in viewing the data object and viewing the metadata of the data object? 
Same for create below?

signekb (commented):
Very nice that you've written it out! 🎉  I have added some suggestions and thoughts

docs/design/naming.qmd
philter87: I would also prefer to skip 'view'

docs/design/naming.qmd
lwjohnst86: Yea, I think edit is nicer because it is shorter 😛 

I think if we say the default action is view, then we don't actually need to include it anywhere.

docs/design/naming.qmd
```diff
@@ -0,0 +1,108 @@
...
+as the actions taken on those objects. These objects and actions are inspired from 
+the "core" language of working with data (CRUD: Create Read Update Delete) as
+well as with HTTP verbs (GET, POST, PUT, PATCH, and DELETE).
+
+- Types of objects Sprout interacts with: projects, data, and metadata.
+- Types of actions Sprout takes on those objects: create, view (or list), edit (or update), delete (?), and upload.
+- Types of identifiers for sub-items of objects (as numbers or text): data, metadata, and project identifiers.
+
+To name things, we'll combine them first with the object name, followed by the
+action, with a natural heirarchy and "ownership" to the objects. Based on these
```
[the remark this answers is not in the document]
lwjohnst86: Maybe descendant might be a better word

docs/design/naming.qmd
lwjohnst86: This is partly why I have made this such a high priority. Right now it many of them do multiple things, so having this naming scheme can help organize what we're doing.

docs/design/naming.qmd
lwjohnst86: That's right! Probably something like the top half of what Henrik has done for the data upload side of things.

docs/design/naming.qmd
lwjohnst86: See @philter87 comment. We can't create data without an id.

docs/design/naming.qmd
lwjohnst86: No, we don't want them able to view the data (unless it's the admin, which case, they have direct access to the database anyway). So viewing the data would actually be viewing the metadata about that data object/table.

lwjohnst86:
I updated the docs based on your comments.

philter87 (approved):
Yay!

PR #283, opened 2024-03-15
refactor: :recycle: True/False seems more appropriate than Yes/No.

### LINKED ISSUE #148: Rename ColumnDataType "True/False"
I have thought about us naming Booleans `Yes/No`, and I think that it might be a bit misleading, since what this data type actually stores a not the words "Yes" and "No", but `True` and `False`. 

I personally like "Boolean" the most, since that's what I know this datatype as and this naming is quite commonly used. If we expect Data Administrators to set up the table and column metadata, I would expect these users to know the boolean data type as well. 

But, if `Boolean` is too technical a term, then maybe, we could use `True/False`, since this is what will be stored in columns with this datatype? 

What do you think?

lwjohnst86:
## Description

- This PR updates the Data Type Yes/No to True/False, since that seems more appropriate. 

## Related Issues

Closes #148

## Testing

- [x] Yes

## Reviewer Focus

## Checklist

### DIFF
--- a/sprout/models/column_data_type.py
+++ b/sprout/models/column_data_type.py
@@ -1,4 +1,5 @@
 """Module defining the ColumnDataType model and COLUMN_DATA_TYPES."""
+
 from django.db import models
 from polars import Series
 
@@ -64,10 +65,11 @@ COLUMN_DATA_TYPES = [
     ),
     ColumnDataType(
         id=3,
-        display_name="Yes/No",
+        display_name="True/False",
         polars_types="Boolean",
-        description="Also known as boolean, or logical. This stores data as either 0/1 "
-        "or True/False. If you need/want to store actual words, use the Text field",
+        description="Also known as boolean or logical. This stores data as either a 0 "
+        "or 1, True or False, or Yes or No. If you need or want to store actual words, "
+        "use the Text field.",
     ),
     ColumnDataType(
         id=4,
--- a/sprout/models/column_data_type.py
+++ b/sprout/models/column_data_type.py
@@ -67,9 +67,9 @@ COLUMN_DATA_TYPES = [
         id=3,
         display_name="True/False",
         polars_types="Boolean",
-        description="Also known as boolean or logical. Sprout stores this data type as True "
-        "or False. If you need or want to store actual words, "
-        "use the Text field.",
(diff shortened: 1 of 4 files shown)


philter87 (commented):
It looks good and I prefer true/false (even though, I suggested yes/no originally). 

But Alisa preferred Yes/no? Or am I not remembering this correctly?

signekb:
@philter87 I can't remember if she preferred it over True/False, but she at least thought it made sense as well

sprout/models/column_data_type.py
```diff
@@ -64,10 +65,11 @@ def __str__(self):  # noqa: D105
     ),
     ColumnDataType(
         id=3,
-        display_name="Yes/No",
+        display_name="True/False",
         polars_types="Boolean",
-        description="Also known as boolean, or logical. This stores data as either 0/1 "
-        "or True/False. If you need/want to store actual words, use the Text field",
+        description="Also known as boolean or logical. This stores data as either a 0 "
+        "or 1, True or False, or Yes or No. If you need or want to store actual words, "
```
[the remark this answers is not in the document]
philter87: Yes, it just converts them to True/False. There is something called categorical data which is a sort of enum. This could be "Yes" or "No", but I don't know if this is something we want

lwjohnst86, line 71:
```suggestion
        description="Also known as boolean or logical. Sprout stores this data type as True "
        "or False. If you need or want to store actual words, "
```

sprout/models/column_data_type.py
```diff
@@ -64,10 +65,11 @@ def __str__(self):  # noqa: D105
...
     ColumnDataType(
         id=3,
-        display_name="Yes/No",
+        display_name="True/False",
         polars_types="Boolean",
-        description="Also known as boolean, or logical. This stores data as either 0/1 "
-        "or True/False. If you need/want to store actual words, use the Text field",
+        description="Also known as boolean or logical. Sprout stores this data type as True "
+        "or False. If you need or want to store actual words, "
+        "use the Text field.",
```
lwjohnst86, line 72:
```suggestion
        description="Also known as boolean or logical. Sprout stores this data type as "
        "True or False. If you need or want to store actual "
        "words, use the Text field.",
```

philter87 (approved):

PR #289, opened 2024-03-17
refactor: :recycle: consolidate duplicated files and code

lwjohnst86:
## Description

- This PR consolidates the `metadata_review`, `column_review`, and `column_review_list` into one function/page, since they all do the same thing and it would be better if they were in the same location.

I've hidden the table and list views for the column review for now, since we should have some type of trigger to show one or the other later.

## Testing

- [x] Yes

## Reviewer Focus

This PR requires an in-depth review. When I run the app and try it out, it works fine. But when I run the tests, it fails for some reason and I can't debug why.

## Checklist


philter87 (dismissed):
Nice!

PR #295, opened 2024-03-18
bug: Remove bfcache on back button click in firefox

### LINKED ISSUE #201: [bug]: "extracting metadata" progress bar in file-upload remains after clicking back from 404 page
### Steps to reproduce

1. Create new table and fill in dialog box details.
2. Upload file. It will pop up with a "extracting" popup progress bar.
3. It will move into extracting metadat.
4. Click back. The popup progress bar is still there. It should restart.

### Expected result

The popup progress bar for extracting metadata should not be there after clicking back.

### Actual result (the bug)

But the popup is still there after clicking back from the next page.

_No response_

### Severity/Priority

None

### Contact Details

_No response_

### Code of Conduct

philter87:
## Description

- This PR fixes bug. The loading bar in 'metadata-create' reappears when hitting 'back' button in chrome. This is solved by removing bgcache (back-forward cache) in firefox. (I did not know this was a thing)

## Related Issues

Closes #201

## Testing

- [ ] Yes

Hard to create a unit test for this, but I have tested manually.

## Reviewer Focus

## Checklist

- [] Tests accompany or reflect changes to the code

### DIFF
--- a/sprout/templates/file-drop-zone.html
+++ b/sprout/templates/file-drop-zone.html
@@ -97,8 +97,13 @@
   }, false)
 
 
-  function showDialog() { 
+  function showDialog() {
     const dialog = document.getElementById("progress-dialog")
-    dialog.show(); 
-} 
+    dialog.show();
+  }
+
+  // This will disable the bfcache (back-forward cache) in firefox to ensure the page
+  // is reloaded, when user clicks the 'back button' in the browser
+  window.addEventListener('unload', function(){});
+  window.addEventListener('beforeunload', function(){});
 </script>


K-Beicher:
@philter87 Are we worried that when googling 'unload event' the developer.mozilla website states that it should no longer be used, should there be a note about it being deprecated, or are they referring to something different?

https://developer.mozilla.org/en-US/docs/Web/API/Window/unload_event

philter87:
> @philter87 Are we worried that when googling 'unload event' the developer.mozilla website states that it should no longer be used, should there be a note about it being deprecated, or are they referring to something different?
> 
> https://developer.mozilla.org/en-US/docs/Web/API/Window/unload_event

Nice catch! I will investigate a bit more. It might be a problem

philter87:
@K-Beicher. You are right, that it is not an optimal solution, but I am unable to find a better solution for clearing the bfcache in Firefox. I have found a long discussion about it [here](implemented yet: https://github.com/whatwg/html/issues/5744.) but it seems that there is no simple way to opt-out of the bfcache, yet. 

But there could be another solution. Maybe we can remove the page in the browser history as returning to it might cause some weird behaviour - like uploading the columns twice

K-Beicher:
@lwjohnst86 is it possible to make an issue that will appear in a years time to revisit this and see if a solution has shown up?

lwjohnst86 (approved):

PR #293, opened 2024-03-17
feat: ✨ upload data file page

lwjohnst86:
## Description

This PR finally adds the page for uploading the data files for #171.

- Please ignore the TODO items, since they will eventually need to be implemented but I can't do it all in this PR (plus depends on others work).
- Right now, the "feature" simply uploads the CSV file to the storage, which was already done by @philter87. The aim is that it joins existing data and updates the database with that data.
- The changes to the JavaScript is because of the formatter, not from me!!

## Related Issues

Closes #171 

## Testing

- [ ] Yes
- [ ] No, not needed (give a reason below)
- [x] No, I need help writing them

I haven't included any tests yet because right now it doesn't do anything that isn't already done. But once other functions have been implemented, those tests will need to be added.

## Reviewer Focus

This PR requires an in-depth review. 

## Checklist

For all PRs that are not general documentation

- [ ] Tests accompany or reflect changes to the code
- [ ] Tests ran and passed locally
- [x] Ran the linter and formatter
- [x] Build has passed locally
- [x] Relevant documentation has been updated

PR #300, opened 2024-03-18
Feat: Wireframe 2.1 - View existing metadata

### LINKED ISSUE #162: Create page for wireframe 2.1 and 2.2: Select Existing Table to add Data to storage (file and database)
We haven't gone this route yet too much, but since it connects well with the uploading files, we can start working on it.

@signekb:
There are these parts to this issue:
+ [X] Show metadata as rows in table with three buttons to `create new`, `edit existing`, or `add data`. #300 
+ [X] When selecting a row, expand to show the column metadata for the selected metadata. #307 
+ [X] Only show first 10 rows when expanding, and if more than 10 rows, have option to expand further. #359 
+ [X] Data Rows #379 
+ [X] Last Upload #379

signekb:
## Description

- This PR adds a simplified version of wireframe 2.1 (overview of existing data) where users can select a row and get redirected.
- This will enable users to choose to create new metadata, edit existing metadata, and upload metadata (there pages are not completed yet, so currently all buttons redirect to the same randomly chosen page.
- If a user clicks "create" they will go to `data-import` 
- If a user clicks "edit" or "upload" they will be redirected to `column-review` (all redirects should be changed later)
- Note: "Data Rows" and "Last Upload" contain the placeholder value "Cell". 

## Related Issues

See Issue #162 (this PR does not close this issue - #307 continues the work towards closing it)

## Testing

- [X] Yes

## Reviewer Focus

- Could the code be clearer/improved on?
- Do we need more tests to cover what this view does?
- Should I remove `-view` at the end of paths/files/functions (since view is default)?
 

## Checklist

- [X] Relevant documentation has been updated (none)


signekb:
@lwjohnst86 How do I update the active deployment? It's outdated? :)

lwjohnst86:
@signekb it seems to have built now. Or is it outdated still?

signekb:
@lwjohnst86 Ah, maybe I was just too impatient

sprout/templates/project-id-metadata-view.html
```diff
@@ -0,0 +1,98 @@
+{% extends 'base.html' %}
+
+{% block content %}
+
+<style>
```
philter87, line 5: Nice! I assumed that styles had to be in a different file (but that was just silly of me). I will need to redo some templates

sprout/templates/project-id-metadata-view.html
```diff
@@ -0,0 +1,98 @@
...
+    background-color: #D9D9D9;
+    margin-bottom: 10px;
+  }
+
+  .table-cell {
+    flex: 1;
+    padding: 8px;
+  }
+
+  .selectable:hover {
```
philter87, line 27: Nice with the hover effect!

sprout/templates/project-id-metadata-view.html
```diff
@@ -0,0 +1,98 @@
...
+      {% empty %}
+      <p>No existing data to show.</p>
+      {% endfor %}
+    </div>
+  </div>
+
+  <!-- three buttons will submit this form: create, edit, or upload -->
+  <nav class="right-align">
+    <button type="submit" name="button_create"><i>add</i>Create Metadata</button>
+    <button type="submit" name="button_edit"><i>edit</i>Edit Metadata</button>
```
philter87, line 78: Potentially you could add the disabled attribute to the buttons based on the query parameter
```
<nav class="right-align">
  <button {% if not request.GET.selected_metadata_id %}disabled{% endif %} type="submit" name="button_create"><i>add</i>Create Metadata</button>
  <button {% if not request.GET.selected_metadata_id %}disabled{% endif %} type="submit" name="button_edit"><i>edit</i>Edit Metadata</button>
  <button type="submit" name="button_upload"><i>upload</i>Upload Data</button>
</nav>
```

image

But the dialog gives a bit more information. 

What do you think @lwjohnst86

sprout/templates/view-existing-tables.html
philter87, line 1: Should this file be deleted? Maybe a result of a rename or is it a different version you want to preserve?

philter87 (commented):
Great work! 
Only one 'big' thing. I was confused about 'view-existing-tables.html'

philter87 (dismissed):
Looks good now :)

sprout/templates/project-id-metadata-view.html
signekb: I actually like your version better, @philter87. I'll change it 👍

philter87 (dismissed):

signekb:
I'm done editing now, I promise 😁

sprout/static/css/style.css
```diff
@@ -28,4 +28,35 @@ dd {
   white-space: nowrap;
   display: inline-block;
   width: 85%;
+}
```
lwjohnst86, line 31: Keep CSS styling within it's own sheet.

sprout/templates/project-id-metadata-view.html
```diff
@@ -0,0 +1,53 @@
+{% extends 'base.html' %}
+
+{% block content %}
+
+<form id="selectedElementsForm" method="post">
+  {% csrf_token %}
+  <div class="table">
+    <div class="table-header">
```
lwjohnst86, line 8: Hmm, what was the reason to use `div` rather than `table` HTML tag? This feels more complicated. Plus my formatter/linter is complaining of an unexpected closing div, but I can't debug it.

sprout/templates/project-id-metadata-view.html
```diff
@@ -0,0 +1,53 @@
...
+      <div class="table-cell">Description</div>
+      <div class="table-cell">Data Rows</div>
+      <div class="table-cell">Date Created</div>
+      <div class="table-cell">Last Upload</div>
+    </div>
+    <div class="table-body">
+      {% for metadata in existing_metadata %}
+      <!-- if statement that changes url depending on selection and deselection of rows -->
+      {% if request.GET.selected_metadata_id|add:'0' == metadata.id %}
+      <a style="display:block" href="{% url 'project-id-metadata-view' %}">
```
lwjohnst86, line 19: Neat way of doing this. Though the big disadvantage here is that whenever you select a new row, the page has to be rendered to a whole other page because it is a redirect. Let's keep everything within this page and not move to other pages.

sprout/templates/project-id-metadata-view.html
```diff
@@ -0,0 +1,53 @@
...
+        {% else %}
+        <a style="display:block" href="{% url 'project-id-metadata-view' %}?selected_metadata_id={{metadata.id}}">
+          {% endif %}
+          <div>
+            <div
+              class="table-row selectable{% if request.GET.selected_metadata_id|add:'0' == metadata.id %} selected{% endif %}"
```
lwjohnst86, line 25: Using the `table` HTML you can use CSS to highlight on hover, rather than using `div` classes for this. Unless you have really strong reasons for using `div` rather than `table`, could you switch to `table`? It will be easier to style if it is a `table` element.

sprout/templates/project-id-metadata-view.html
```diff
@@ -0,0 +1,53 @@
...
+    </div>
+  </div>
+
+  <!-- three buttons will submit this form: create, edit, or upload -->
+  <nav class="right-align">
+    <button type="submit" name="button_create"><i>add</i>Create Metadata</button>
+    <button {% if not request.GET.selected_metadata_id %}disabled{% endif %} type="submit"
+      name="button_edit"><i>edit</i>Edit Metadata</button>
+    <button {% if not request.GET.selected_metadata_id %}disabled{% endif %} type="submit"
+      name="button_upload"><i>upload</i>Upload Data</button>
```
lwjohnst86, line 48: I'd prefer that redirect happens here, not on the Django side (see how I did it in #293). So it would be an `a` element with `href` to other pages. The less we have to send to Django for small things like this, the less traffic between server and client there has to be. 

Of course, these would only be clickable after selecting a page so will need some JavaScript.

lwjohnst86 (changes requested):

sprout/templates/project-id-metadata-view.html
signekb: @philter87 Luke prefers to keep style in a separate file, so I'll change it here instead

sprout/templates/project-id-metadata-view.html
signekb: I ran into some issues when I added the expand feature showing the column metadata. But I'll just change it again and see if I can figure out the way to expand using the `table` tag.

sprout/templates/project-id-metadata-view.html
signekb: I changed i now, but use quite a lot of js. I can probably be optimised.

sprout/templates/project-id-metadata-view.html
signekb: I have changed it back to `table`

sprout/templates/project-id-metadata-view.html
signekb: As described in another comment above, I have changed it now and use js, but I think that the js code could/should be optimised

signekb:
@philter87 and I have now reviewed the code together and simplified the scripts. 
Note two changes in particular:
1. It is now not possible to deselect a row when it has been selected (we discussed this, and it might not be necessary to deselect a row after all)
2. We have removed four tests that check that the redirects when the different buttons have been clicked. This is not possible anymore with the js solution.

lwjohnst86 (approved):
Super nice and clean! Great job!

PR #304, opened 2024-03-19
docs: :memo: update naming scheme after extensive discussions about it

### LINKED ISSUE #303: Discussion: For the naming scheme, should metadata come first of data?
### Skipping metadata
projects/{id}/data/create
projects/{id}/data/{id}
projects/{id}/data/{id}/edit
projects/{id}/data/{id}/update
projects/{id}/data/{id}/add
projects/{id}/data/{id}/delete

### metadata as parent
projects/{id}/metadata
projects/{id}/metadata/create
projects/{id}/metadata/{id}/index.json
projects/{id}/metadata/{id}/data/index.sql
projects/{id}/metadata/{id}/data/update
projects/{id}/metadata/{id}/data/edit
projects/{id}/metadata/{id}/data

lwjohnst86:
## Description

I thought about this quite a bit today after our discussion. I think moving the metadata before data removes the "equivalent actions" problem with my original scheme, and also matches better with how the user walks through the process. So I've updated the naming scheme to have metadata first and then data :tada: 

## Related Issues

Closes #303

## Reviewer Focus

This PR only needs a quick review. I've mostly revised the scheme to match what we discussed during the meeting.

### DIFF
--- a/docs/design/naming.qmd
+++ b/docs/design/naming.qmd
@@ -23,7 +23,7 @@ We'll compose names based on the objects we and our users interact with as well
 as the actions taken on those objects. These objects and actions are inspired by
 and align with the "core" language of working with data called *CRUD* (Create
 Read Update Delete), common naming schemes used in REST API applications or
-services (such as Github), as well as by the vocabulary of the HTTP verbs (GET,
+services (such as GitHub), as well as by the vocabulary of the HTTP verbs (GET,
 POST, PUT, PATCH, and DELETE):
 
 - Types of objects Sprout interacts with: projects, data, and metadata.
--- a/docs/design/naming.qmd
+++ b/docs/design/naming.qmd
@@ -15,15 +15,19 @@ How we name things, both internally and user-facing, will be:
 - Consistent: Remains the same across different contexts
 - Align with existing conventions: Follow widely-established standards and conventions
 
-## Composing names
+## User-facing content
+
+### Composing names
 
 We'll compose names based on the objects we and our users interact with as well
-as the actions taken on those objects. These objects and actions are inspired by and align with
-the "core" language of working with data called *CRUD* (Create Read Update Delete), common naming schemes used in REST API applications or services (such as Github), as
-well as by the vocabulary of the HTTP verbs (GET, POST, PUT, PATCH, and DELETE):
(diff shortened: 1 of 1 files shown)


docs/design/naming.qmd
```diff
@@ -40,42 +44,46 @@ names. The names are separated by a symbol based on its context:
 Additional "rules" to the naming include:
 
 - When a naming scheme ends without an action, the default action is `view`.
-- When there are naming schemes that have a duplicate final result, the shorter name is preferred.
+- Data objects do not have any `view` action, since we want to limit access to looking at raw data.
```
K-Beicher, line 47: Do we want to end this sentence with 'at raw data in Sprout.'

K-Beicher (dismissed):
I like it. Just the one comment, I'm more thinking of not boxing ourselves in by saying we don't want to make it possible to look at data in Seedcase, we may want to add this functionality to a different Seedcase component in future.

docs/design/naming.qmd
lwjohnst86: Since this doc and naming scheme are for Sprout and in the Sprout repo, that isn't necessary ☺️

lwjohnst86:
> I like it. Just the one comment, I'm more thinking of not boxing ourselves in by saying we don't want to make it possible to look at data in Seedcase, we may want to add this functionality to a different Seedcase component in future.

@K-Beicher agreed! Though I think that would be well into the future. When we reach that point, we can revise the naming scheme then. So for now, we assume (non-admin) users can't see data contained within Sprout (aside from of course the data they upload).

lwjohnst86:
> I like it. Just the one comment, I'm more thinking of not boxing ourselves in by saying we don't want to make it possible to look at data in Seedcase, we may want to add this functionality to a different Seedcase component in future.

@K-Beicher agreed! Though I think that would be well into the future. When we reach that point, we can revise the naming scheme then. So for now, we assume (non-admin) users can't see data contained within Sprout (aside from of course the data they upload).

philter87 (dismissed):
I also like it. Let's see how it goes or we can always change to the data-first perspective

signekb (commented):
Yes, this also makes more sense to me. Lets go 🚀

PR #305, opened 2024-03-19
docs: :memo: naming scheme for internal Python code 

### LINKED ISSUE #224: Discussion: Naming practices of functions

lwjohnst86:
## Description

Related to general discussions on naming things. I know for myself, having some type of agreed on way of naming things makes it easier for me to create, review, and understand code. So consider this PR as a starting point for a discussion on this topic for *internal* facing content.

## Related Issues

Closes #224


philter87 (dismissed):

docs/design/naming.qmd
```diff
@@ -8,50 +8,64 @@ title: "Naming of Things"
...
+objects. Based on these principles, we can derive a naming scheme. In
+the scheme, objects and actions are ordered, with names occurring first
+acting as a "parent" to later names. The names are separated by a symbol
+based on its context:
+
+-   `_` for Python function names.
+-   `-` for file path names.
+-   `/` for URLs and web API endpoint names (a `/` always begins each
+    URL to act as a root, but never ends a URL).
+-   for command line interface names.
```
lwjohnst86, line 54: Note that this is only showing reformatting and changes in whitespace. You can ignore this. It is styling only.

docs/design/naming.qmd
```diff
@@ -87,3 +101,52 @@ projects <id> metadata <id> delete
...
+-   Functions:
+    -   Use `snake_case()`.
+    -   Prefer a single action verb for functions: `action()`.
+    -   Use the arguments to specify the object: `action(object=)`.
+    -   If necessary, include the object in the name following the
+        pattern `action_object()`.
+    -   If the object is in the name, prefer the plural form of the
+        object if the function could return multiple items of an object.
+    -   If the object is in the name, prefer a single object name rather
+        than multiple.
```
K-Beicher, line 131: Could we do an example here?

K-Beicher (dismissed):
Other than the one comment it looks fine to me.

docs/design/naming.qmd
```diff
@@ -87,3 +101,52 @@ projects <id> metadata <id> delete
...
+
+### Python functions
+
+Following the same naming schemes for CRUD, HTTP verbs, and common REST
+API endpoints, these are some guidelines on the general words to use for
+actions and objects:
+
+-   Actions: get/list, update/edit, delete/remove, create,
+    convert/transform/modify, extract, validate/check/verify/is, read,
+    write, as/to
```
signekb, line 115: Should we limit some of the synonym actions here?
```suggestion
-   Actions: list, edit, delete, create, extract, validate, read,
    write, to
```

docs/design/naming.qmd
```diff
@@ -87,3 +101,52 @@ projects <id> metadata <id> delete
...
+-   Objects: array (analogous to non-nested Lists, Sets, and Series (in
+    Pandas/Polars) in Python), lists (analogous to nested Lists, Tuples,
+    and Dictionaries in Python or Tables in SQL), files, classes
```
signekb, line 118:
```suggestion
-   Objects: array (analogous to non-nested Lists, Sets, and Series in
    Pandas/Polars in Python), lists (analogous to nested Lists, Tuples,
    and Dictionaries in Python or Tables in SQL), files, classes
```

docs/design/naming.qmd
```diff
@@ -87,3 +101,52 @@ projects <id> metadata <id> delete
 projects <id> metadata <id> data update
 projects <id> metadata <id> data delete
 ```
+
+## Developer-facing (internal) content
+
+### Python functions
```
signekb, line 107: This section includes naming schemes for more than functions - also classes,  methods, variables, args, etc.

```suggestion
### Python
```

docs/design/naming.qmd
signekb: I would also like some examples here :)

docs/design/naming.qmd
```diff
@@ -87,3 +101,52 @@ projects <id> metadata <id> delete
...
+    -   Prefer a single action verb for functions: `action()`.
+    -   Use the arguments to specify the object: `action(object=)`.
+    -   If necessary, include the object in the name following the
+        pattern `action_object()`.
+    -   If the object is in the name, prefer the plural form of the
+        object if the function could return multiple items of an object.
+    -   If the object is in the name, prefer a single object name rather
+        than multiple.
+-   Methods (includes everything from the functions):
+    -   Prefer action verbs only: `ClassName.action()`.
```
signekb, line 133: is this different from functions? Or why only this one point for methods?

docs/design/naming.qmd
```diff
@@ -87,3 +101,52 @@ projects <id> metadata <id> delete
...
+-   Classes:
+    -   Use `PascalCase`.
```
signekb, line 135: Oh, new case for me! I know it as upper camel case.

docs/design/naming.qmd
```diff
@@ -87,3 +101,52 @@ projects <id> metadata <id> delete
...
+        (`TwoObjects`).
+-   Variables (of any type):
+    -   Use `snake_case`.
+    -   Prefer a plural form for objects if they *could* contain more
+        than one item.
+    -   Use singular form *only* if the object *must* only contain one
+        item.
+    -   If an action name is necessary, prefix action verb in the past
+        tense: `actioned_object`.
+-   Arguments (includes everything from the variables above):
```
signekb, line 148:
```suggestion
-   Arguments (includes everything from variables above):
```

docs/design/naming.qmd
lwjohnst86: different because the action happens to the object (which is the class, already defined). I'll add an item saying "avoid object names in methods".

docs/design/naming.qmd
lwjohnst86: @signekb yes for sure!!! I included them all to get your thoughts and opinions. My only disagreement with what you've changed is `validate`. I'll add another guideline, to choose shorter action verbs over longer ones. So I think `verify` or `check` would be preferable. I like `verify` over `check`. What about the `modify` action verbs? I guess `edit` could cover that... 🤔

docs/design/naming.qmd
lwjohnst86, line 131:
```suggestion
    -   If the object is in the name, prefer a single generic object name rather
        than multiple specific names: `read_csv()` rather than `read_csv_file()` or `upload_file()` rather than `upload_csv_file()` (uploading isn't depending on file extension, while reading is)
```

docs/design/naming.qmd
lwjohnst86, line 133:
```suggestion
    -   Prefer action verbs only: `ClassName.action()`.
    -   Avoid using object names, unless it makes the action clearer.
```

PR #307, opened 2024-03-20
Feat: Wireframe 2.2 - Show column metadata when row is selected

signekb:
## Description

- This PR builds upon #300  and adds the feature that a row is expanded to show the column metadata of all columns when selected.
- These changes are done so users can see the column metadata for the selected metadata. 
- Note: Again, this is a simplified version of the original wireframe. Ideally, viewing the column metadata of a row should be separate of selecting it. 

## Related Issues

Closes #...

See also Issues #162

## Testing

- [ ] Yes
- [X] No, I need help writing them

I'm not sure which tests I can/should include. Help? :) 

## Reviewer Focus

- Can this implementation be optimised/simplified? 
- Which tests do we need for this solution? 

## Checklist


K-Beicher:
Screenshot 2024-04-03 at 14 59 33
This looks very confusing to me, but that may just be because some functionality is missing.  Could we repeat the headers for the table metadata, so they don't follow on from the field metadata?

signekb:
@philter87 Seems like the css styles doesn't work as expected when Sprout is deployed? I swear it looked fine when I ran it locally 😅

philter87:
> @philter87 Seems like the css styles doesn't work as expected for Kris? I swear it looked fine when I ran it locally and it also looks fine in the deployment for this PR 😅

It also looks great on my windows computer in chrome, edge and firefox :)

sprout/templates/project-id-metadata-view.html
```diff
@@ -41,18 +68,32 @@
...
+    // Loop through all the metadata columns and show the ones that matches the selected row
+    for (let column_number = 0; column_number < metadata_columns.length; column_number++) {
+      // If the metadata column's id matches the selected row's id, remove the 'hide' class
+      if (metadata_columns[column_number].id === row.id) {
+        metadata_columns[column_number].classList.remove("hide");
+      } else {
+        // Otherwise, add the "hide" class
+        metadata_columns[column_number].classList.add("hide");
+      }
+    }
```
philter87, line 96: Small thing. Maybe you can use this for-loop instead (This is not supported in IE, but that is okay. I have already done many things not supported by IE )

```suggestion
    // Loop through all the metadata columns and show the ones that matches the selected row
    for (const metadata_column of metadata_columns){
      // If the metadata column's id matches the selected row's id, remove the 'hide' class
      if(metadata_column.id === row.id){
        metadata_column.classList.remove("hide")
      } else {
        // Otherwise, add the "hide" class
        metadata_column.classList.add("hide")
      }
    }
```

philter87 (dismissed):
Yay! Looks great! Maybe @K-Beicher can share her screen, so we can figure out what is up

lwjohnst86 (dismissed):
Niiiccceee. I don't really understand the JavaScript, but seems @philter87 has a handle on that 🦾

sprout/templates/project-id-metadata-view.html
signekb: @philter87 Ah, alright! Is the one I implemented an older version of a for-loop or how come there's different kinds?

philter87:
@K-Beicher and @signekb, I have used a hidden attribute instead which might do the trick (But I am unable to test)

K-Beicher:
I can confirm that this now works fine in Safari and Chrome on my Mac.

signekb (commented):
@philter87's changes works for me in Chrome and Safari too 🎉  (but I can't approve my own PR)

lwjohnst86 (approved):

PR #308, opened 2024-03-20
feat: Exclude column checkbox

### LINKED ISSUE #243: [feature-request]: Option to select which columns to upload/keep during the metadata extraction phase
### What feature would you like to request?

During the demo, we realized that we should have an option to select which columns get uploaded and which don't. Either that or we inform the user to delete the column themselves before uploading..?

### Optional: Suggest a solution

_No response_

### Code of Conduct

philter87:
## Description
- This PR adds the feature to exclude columns on the tempåla`column-review.html` (both in 'grid-view' and 'table-view')

### Grid overview

**I HAVE BEEN PLAYING A BIT. IT COULD ALSO BE A CHECKBOX BELOW DESCRIPTION FIELD.**

In the grid overview, it is displayed with a clickable bin-icon.
image

When excluded it will show like this (The action can be undone by clicking):
image

### Table overview
In the table overview, I have just added a column to the table:
image

## Related Issues

Closes #243 

See also Issues #...

## Testing
- [X] Yes

## Reviewer Focus

- When excluding a column, the description field should not be required. This is handled with a bit of javascript and it is handled in the backend.
- I add a overlay to a card when it is excluded. This also requires a bit of javascript.

## Checklist


philter87:
> Maybe a warning on the one with the Bin, as once you click it is very much gone!

It is only gone, when you click the button "Update All". Until clicking "Update All", you can revert with the "undo"-button at the top. 

But I can also add warning, if we want to be sure?

lwjohnst86:
@philter87 I think the icon is enough, but we'll have to see during user testing (probably not achievable in this iteration)

sprout/templates/column-review.html
signekb, line 1: I easily lose the overview of changes in templates when it includes JavaScript. But it seems to work as I would expect it to. Maybe you could add some explanatory comments (for my sake? 😄 )?

signekb (dismissed):
Nice! I have asked for some more comments for my understanding of the javascript part, but otherwise it seems to work as expected 👍

lwjohnst86 (approved):
Niiicceee! I only changed it so longer variables get truncated with `...` so the bin doesn't fall off the card.

PR #324, opened 2024-03-21
fix: add that no spaces are allowed in error message

### LINKED ISSUE #314: fix: In table name error message, add that space cannot be included
https://github.com/seedcase-project/seedcase-sprout/blob/9131df78d179b2c8b785e98a4ed7d693436fca39/sprout/validators.py#L29-L30

signekb:
## Description

- This PR adds the that no spaces are allowed when naming a table/metadata.
- These changes are done because after the user test today, where we noticed that this information could be included explicitly for clarity. 

## Related Issues

Closes #314 

## Testing

Already added.

## Reviewer Focus

- Maybe the error message sentence is a bit too long now? Or is it ok?
- Why does deploy fail here?

## Checklist

### DIFF
--- a/sprout/tests/tests_forms_tablemetadata.py
+++ b/sprout/tests/tests_forms_tablemetadata.py
@@ -1,4 +1,5 @@
 """Tests for forms."""
+
 from django.test import TestCase
 
 from sprout.forms import TableMetadataForm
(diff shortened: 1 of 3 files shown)


philter87 (approved):
Nice! 

Should we also have a display name and a machine readable name for tables as well? Sorry for potentially creating a huge discussion, but less restrictions could improve the user experience? 

(In this case the user would never see the machine readable name, except if the look in the database)

signekb:
@philter87 Oh no 😅 I guess the table name as it is now is the machine readable, right? But you're right, we could do a similar thing as with the column names - it just depends on where we land with that.
What say you @lwjohnst86 and @K-Beicher ?

lwjohnst86 (approved):
💪🏻

PR #334, opened 2024-03-25
feat: Support for decimals with comma is added

### LINKED ISSUE #268: Include tests to check for `;` separators with `,` decimals
I couldn't see if there are tests available for that, but it would be good to include tests to assess that

philter87:
## Description

- This PR adds a feature. We want to support ';' as separator and ',' in decimals. This is not supported by polars by default (I think?)

## Related Issues

Closes #268

## Testing

- [X] Yes

## Reviewer Focus

Notes:
- I am solving this with regular expressions as part of the CSV preparation step. I don't know if there is a better way to do this? It seems a bit hacky. 
- All values in a column needs to match `r"(\d+),(\d+)"` or be empty, otherwise, the conversion is not applied.
- A value is only matched if there are numbers before and after the comma. Examples:
  - '12,345' is matched by the regex pattern and becomes '12.345'
  - 123 is not matched by the regex pattern but that is okay because I don't need to replace anything
  - ',12' is NOT matched by the regex! Should we support add support for this? I want to make the regular expression as 'tight' as possible. We can add support by using `r"(\d*),(\d+)"` instead. 

## Checklist

### DIFF
--- a/sprout/csv/csv_reader.py
+++ b/sprout/csv/csv_reader.py
@@ -84,6 +84,7 @@ def _transform_to_suitable_csv_format(csv_path: str, row_count: int | None) -> s
     df = df.select(pl.all().name.map(lambda n: n.strip().strip('"')))
     df = df.select(pl.all().str.strip_chars('"'))
 
+    df = df.select([_convert_decimal_with_comma_to_dot(column) for column in df])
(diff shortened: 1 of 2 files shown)


lwjohnst86 (approved):
🎉 🌮

PR #335, opened 2024-03-25
chore: :technologist: quality of life improvement for VS Code to find the virtual environment

lwjohnst86:
## Description

This PR moves the default virtual environment created by poetry from the default path to the workspace folder `.venv` location. This makes it so VS Code can more easily find it and also provide access to better detection of function documentation, import checks that the package is installed, etc.

You'll all probably need to run `poetry install` followed by maybe restarting VS Code to trigger that it can find this new location and make use of it's features when developing in Django/Python.


philter87 (dismissed):
Nice!

signekb (dismissed):
Very nice 🥳

PR #336, opened 2024-03-25
feat: Values preserved between view switches

### LINKED ISSUE #317: Saving input when switching views (related to update all without needing required fields)
Connected to a few other issues, about saving things to return to it later.

philter87:
## Description

- This PR adds a feature. On 'column-review.html', the input values are preserved when switching between table view and grid view.

## Related Issues

Closes #317 

## Testing

- [ ] Yes

I was unable to write tests for this as it is javascript

## Reviewer Focus
**This is a small PR, but I need some help to figure out if this is the correct solution**. 

I decided to move the input values with Javascript and not persist the values when switching views. From my perspective, this is what the user expects... But I can easily be persuaded to change this?

## Checklist


signekb (commented):
When I test it by adding a description in the card view and switch to table view, the description is not shown in the table view. However, the value is preserved in the sense that when I return to card view, it's still there. 

So, I'm thinking that this isn't working as expected? :)

philter87:
> When I test it by adding a description in the card view and switch to table view, the description is not shown in the table view. However, the value is preserved in the sense that when I return to card view, it's still there.
> 
> So, I'm thinking that this isn't working as expected? :)

You are right. I failed to correctly rename a variable. It works now

signekb (changes requested):
Ah, nice! It seems that column names and descriptions update as expected when I switch view, but the data types doesn't? So, something's still not quite right? :)

lwjohnst86 (changes requested):
Both the display name and description get changed but the data type doesn't.

philter87:
Fixed: DataType and Excluded flag are now preserved when switching views

K-Beicher (approved):
All fields now seem to work as expected, looks great!

lwjohnst86 (approved):

Issue #229, opened 2024-03-04
Wireframes for log in


henrik-goul:

lwjohnst86:
Related to the discussion, use an unique username, email, and password, rather than using email as username.

lwjohnst86:
We decided to start the web app from scratch after the major design/architecture decisions and changes we made. No longer relevant.

PR #347, opened 2024-04-03
feat: authentication and password reset is created

### LINKED ISSUE #330: Implement login page based on wireframe
For now, assume a "create new user" with email and password, and not have external logins (e.g. MitID).

### LINKED ISSUE #331: Create user details Django Model
I guess before we can have a log-in page, we need a model (plus form) to store user details.

philter87:
## Description

- This PR adds authentication to seedcase. The changes are mostly done in a new "django application" called "accounts". It provides a page for login, registration (signup) and a password reset flow. See images below:

**Login** (accounts/login/)
image

**Registration** (accounts/register/)
image

**Password reset** (accounts/password_reset/)
image

## Related Issues

Closes #331, Closes #330 

See also Issues #229, #2 

## Testing

- [X] Yes

## Reviewer Focus

This PR is big, but there is a lot of boilerplate as I have created a new "django app" to encapsulate the authentication. 

I have used the default Django authentication system as much as possible, but it required a bit of customization because I want to use **email for login** and **not the username**. This is what I have done to make the authentication work:
  - To include the default authentication system in Django, you need to include the relevant urls in the urlpattern variable. I have done this in `accounts/urls.py` with the line `path("", include("django.contrib.auth.urls"))`. This will include these patterns:
```
accounts/login/ [name='login']
accounts/logout/ [name='logout']
accounts/password_change/ [name='password_change']
accounts/password_change/done/ [name='password_change_done']
accounts/password_reset/ [name='password_reset']
accounts/password_reset/done/ [name='password_reset_done']
accounts/reset/<uidb64>/<token>/ [name='password_reset_confirm']
accounts/reset/done/ [name='password_reset_complete']
```
- **Password reset pages**. Django provides templates and views for the password reset pages. This would have been tricky to do, so this is a great temporary solution. We will need to customize this later.
- **Login page**. Django does not provide a default template for the login page, so I have created a login template called: "registration/login.html". The location is a bit weird, but this is required by Django. The login request is handled automatically by the Django framework, so I don't need to create a view function.
- **Register user page**. The registration of a new user is done in the template: "registration/register.html". Weirdly, the registration of users is not provided by Django so I have created both a template (registration/register.html) and view for the registration page (in views.py).  
- **Email instead of username**. By default, Django uses **username** for login, but I want to use an email. I have solved this by creating a custom user model called CustomUser.

## Checklist


K-Beicher (dismissed):
After deletion of the db file, and running 'just start-app' it works nicely.  Didn't quite figure out how to check the reset function, but the account creation and login seems to work.

sprout/templates/base.html
signekb, line 1: Why are there changes to this file under Sprout in this PR? :)

config/settings.py
```diff
@@ -64,6 +65,20 @@
     "django.middleware.clickjacking.XFrameOptionsMiddleware",
 ]
 
+LOGIN_REDIRECT_URL = "home"
+AUTH_USER_MODEL = "accounts.CustomUser"
+
+# The EMAIL_BACKEND is responsible for sending password-reset emails.
+# We use a mock email backend (print to console) until we get a real email account.
```
signekb, line 72: When you say "we get a real email account", what do you mean? A seedcase email orrr?

accounts/models.py
```diff
@@ -0,0 +1,10 @@
+from django.contrib.auth.models import AbstractUser
+from django.db import models
+
+
+class CustomUser(AbstractUser):
```
signekb, line 5: Should this have a more descriptive name than `CustomUser`? Or is it simply what it's usually called in Django?

accounts/templates/registration/login.html
```diff
@@ -0,0 +1,71 @@
...
+          </div>
+        </div>
+
+        <!-- LOGIN BUTTON -->
+        <div class="content-center margin">
+          <button>Login</button>
+        </div>
+
+        <div class="error-msg">
+          {{ form.non_field_errors|striptags }}
```
signekb, line 50: Is there a simple way to customise some of the error messages? 
I played around with writing incorrect emails and passwords and got the quite unspecific error message that "both fields *may* be case sensitive".

accounts/tests.py
```diff
@@ -0,0 +1,72 @@
...
+
+        # Act
+        response = self.client.post(url, form_data)
+
+        # Assert
+        self.assertContains(response, "This password is too short")
+        self.assertContains(response, "This password is too common")
+        self.assertContains(response, "This password is entirely numeric")
+
+    def test_create_user_and_login(self):
```
signekb, line 51:
```suggestion
    def test_create_account_and_login(self):
```
Just to keep the wording similar to the other tests

accounts/tests.py
```diff
@@ -0,0 +1,72 @@
...
+        """Register and login with account."""
+        # Arrange
+        password = "as123dsSf"
+        email = "philter@philter.com"
```
signekb, line 55: Could this be moved up in a `setUp` function to avoid having to repeat it in multiple tests?

signekb (commented):
Looks good! I have added some suggestions and questions :) 
A also have a few questions to the description of this PR: 
1. Under **password reset pages**, you wrote that we will "customise this later". What do you mean by that specifically? :) 
2. About the use of email as username. I think we discussed at an update meeting that this might not be a great idea after all, but I can't remember the conclusion. Can you?

philter87:
> Looks good! I have added some suggestions and questions :) A also have a few questions to the description of this PR:
> 
> 1. Under **password reset pages**, you wrote that we will "customise this later". What do you mean by that specifically? :)
> 2. About the use of email as username. I think we discussed at an update meeting that this might not be a great idea after all, but I can't remember the conclusion. Can you?

1. We can customize these pages by adding django templates in the folder templates/registration
2. @K-Beicher asked if the email could be changed later. I initially said 'no', but now I do think it is possible to change the email. The users are assigned an id (integer) which is used to reference the user

accounts/admin.py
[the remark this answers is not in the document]
philter87: In Django, they use the word "group" for a "role". So people belong to a group instead of a having a role.

The model is called "models.Group" and is defined by Django.

accounts/apps.py
```diff
@@ -0,0 +1,8 @@
+from django.apps import AppConfig
+
+
+class AccountsConfig(AppConfig):
+    """AccountsConfig class."""
+
+    default_auto_field = "django.db.models.BigAutoField"
```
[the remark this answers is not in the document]
philter87: Django automatically adds a "id"-column to our models. This field defines which data type should be used for the id-columns. This is a Big integer, which automatically increments when new rows are added

accounts/static/css/accounts.css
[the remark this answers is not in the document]
philter87: We can reuse the styles from "sprout" app, but in this case I decided not to. I don't think I created any CSS "class" which are used by both, but eventually there will be shared CSS classes

config/settings.py
philter87: We need a "real" email account to be able to send mails. It could be just a gmail (seecase123@gmail), but people might expect something like info@seedcase.com

accounts/models.py
philter87: Maybe we should have a better name, but I dont know what

accounts/templates/registration/login.html
philter87: Yes, that would be possible, but I don't know for sure how difficult/easy it is at the moment. But maybe it is possible to just override one of the "clean" methods

sprout/templates/base.html
philter87: Yes, that requires a comment. 

I wanted to be able to add headers as well in the base template, so we wan import CSS and other related assets based on what is needed on a page

K-Beicher (dismissed):
In that case I'm happy to add my review to this

sprout/templates/includes/base.html
```diff
@@ -14,6 +14,10 @@
     <!-- Global css -->
     <link rel="stylesheet" href="{% static 'css/style.css' %}" />
     <link rel="stylesheet" href="{% static 'css/file-drop-zone.css' %}" />
+
+    <!-- Headers required for a page is injected by using a block called 'headers' -->
+    {% block headers %}
+    {% endblock headers %}
```
signekb, line 20: Why is it called "headers" here but "content" in `sprout?`

signekb (dismissed):
I have lost track a bit of this PR, I must admit. But I have played around with it and it seems to work as expected. I'll approve 👍

sprout/templates/includes/base.html
philter87: I am using the base.html from sprout, so the block called "content" still exists, I just added another "block" called "headers". 

A django template can have multiple blocks, so when you "extend" the template you can "inject" multiple blocks. I wanted a place in base.html to inject header-tags because I wanted to add the stylesheet "accounts.css". But I only want to include accounts.css one accounts-related pages.

Or are you just referring to the name: "headers". It could also be "header" or "header-tags".

signekb: Thanks for the elaboration. Yes, I was mainly thinking of the naming :)

PR #359, opened 2024-04-05
feat: ✨ existing metadata view expanded to show first 5 cols

signekb:
## Description

- This PR adds the feature of showing 5 metadata columns initially and then showing more than 5 metadata columns when clicking the "show more" button
- It's set to 5 now for testing purposes, even though we decided on showing 10 initially (since the test data/from the fixtures has less than 10 columns)

## Related Issues

See also Issues #162 

## Testing

- [ ] Yes
- [X] No, I need help writing them

I have not added any tests since this is mostly done with JS. Help?

## Reviewer Focus


signekb:
@philter87 I don't understand why the deployment failed. Can you help?

philter87:
> @philter87 I don't understand why the deployment failed. Can you help?

I don't know. I just tried again and it seemed to work

philter87 (dismissed):

sprout/templates/projects-id-metadata-view.html
```diff
@@ -97,6 +100,40 @@
           metadata_column.hidden = true
         }
       }
+      // for tags with "shown" class, remove the class and add hidden attribute
```
lwjohnst86, line 103: I'm not loving all that the JavaScript is growing... and intuitively I feel there should be a better approach to this that uses less code, considering this is such a common thing to do... But I don't yet have the knowledge of frontend stuff to know how it should look. I've been looking at Alpine.js in addition to HTMX since many in Django community suggest these tools, and it seems like that has the tools to more easily do this selection and showing and listing stuff.

sprout/views/projects_id_metadata_view.py
```diff
@@ -21,7 +21,14 @@ def projects_id_metadata_view(request: HttpRequest) -> HttpResponse:
...
     existing_metadata = Tables.objects.all()
-    existing_metadata_columns = Columns.objects.all()
+
+    # create a dictionary of existing metadata columns to filter in the template
+    dict_existing_metadata_columns = {}
+
+    for column in Columns.objects.all():
+        if column.tables.pk not in dict_existing_metadata_columns:
+            dict_existing_metadata_columns[column.tables.pk] = []
+        dict_existing_metadata_columns[column.tables.pk].append(column)
```
lwjohnst86, line 31: I really don't understand the logic going on here, nor why it is needed. @signekb could you explain a bit and maybe convert this into a function with clearer input and output type hints?

lwjohnst86 (commented):
I've been looking through this to see if I can simplify it, but its a steep learning curve 😆 😭 I'll maybe return to this later.

sprout/templates/projects-id-metadata-view.html
signekb: Yeahh, I know that it isn't too good that the JS keeps growing. My thought process was that we'd just keep with JS for now instead of mixing HTMX and JS - which we ofc could. I'll look into Alpine.js next week

sprout/views/projects_id_metadata_view.py
signekb: It's converting the columns object into a dict with tables.pk (i.e., the primary key, which is `id`) as keys and the columns as values, so I can use the id in the template to filter. I'll look into making this clearer/more elegant next week.

sprout/templates/projects-id-metadata-view.html
philter87: The problem with frontend stuff is there are so many ways to solve this. I can maybe imagine 4 different approaches. It seems Alpine.js might solve some of these issues. The new guy might also suggest "React", but it depends on how much "dynamic" behavior we need on the client

lwjohnst86: React is way too excessive and complex for us though, plus requires we use Django as only the backend. We only need simple things, not too complicated ☺️ and something so we can keep using Django.

philter87: I think you are right, that React is too complicated, but I think we should merge this in and potentially refactor later

lwjohnst86: Definitely agree. Just waiting on @signekb to address my other comment first ☺️

signekb: @lwjohnst86 Should I try to change the current implementation to alpine.js? Or in what way do you mean I need to address your comment?

lwjohnst86: Sorry, I realized I hadn't answered your question. Leave the JavaScript as is, that is fine. We will refactor to Alpine or something else later. I was meaning the still unresolved comment below ☺️

sprout/views/projects_id_metadata_view.py
philter87: I have simplified it a bit by calling `prefetch_related("columns_set")`. This automatically populates the tables with the corresponding columns in a property called `columns_set`

philter87 (dismissed):

signekb (commented):
@philter87 's changes looks very nice! I tried it locally and it seems to work just like before 🎉 
I can't approve bc it's my PR.

lwjohnst86 (dismissed):

PR #368, opened 2024-04-07
feat: create database from uploaded data

lwjohnst86:
## Description

This is a stacked PR based on two others (#294 and #293), so looks like it has a LOT of files but that's from #293.

This PR adds the page for `metadata id data create` that has been missing. Made a connection to continue from the `metadata id update` (which should actually be `metadata create`).

There is still a few things to do, but just wanted to put it up to get some eyes.

Specifically, I don't know how the page should look, it's very basic right now. And I don't know how it fits with the wireframes, since we haven't really gotten to it (sorta, but not really).

Still need to create some tests.

Thoughts on how the page should look? In the process of doing this page, I realized we also need to have a separate page for metadata creation and metadata update (since we don't want to upload data to a database if it's only updating the metadata). And right now, this data create page links from metadata update... which isn't right.

It probably won't build because it depends on those other in progress PRs, which also need updates from `main`.

## Related Issues

Related to #171

See also Issues #...

## Testing

- [ ] Yes

Still need to create them.

## Reviewer Focus

This PR requires an in-depth review, especially around how the page should look.

## Checklist


lwjohnst86:
We've done a lot of updates and refactoring, so basically all of this is no longer relevant. Closing.

PR #378, opened 2024-04-09
Discussion/Suggestion: edit metadata or add data for each row?

philter87:
## Description

- This PR moves the "edit metadata" and "add data" buttons to a "more"-button on each row. See picture below. Something confused Omar about the buttons, but I am not sure what. This is just a suggestion based on the "User management" wireframes, where actions are added to a "more"-button.

### Suggestion 1:
image

### Suggestion 2:
image

### Suggestion 3:
image

### Suggestion 4
image

## Testing

- [ ] Yes

Not possible to test and it is just a suggestion 

## Reviewer Focus

The code is not so difficult, but from a UX perspective is this okay?

## Checklist

### DIFF
--- a/sprout/templates/projects-id-metadata-view.html
+++ b/sprout/templates/projects-id-metadata-view.html
@@ -22,7 +22,8 @@
           <td>{{ metadata.created_at }}</td>
           <td>Cell</td>
           <td style="text-align: right">
-            <button class="circle fill" style="z-index: 10" onclick="event.stopPropagation()">
+            <!-- Actions button. z-index decreasing so the next is not covering this -->
+            <button class="circle transparent" style="z-index: {{ forloop.revcounter }}"  onclick="event.stopPropagation()">
               <i>more_horiz</i>
               <menu class="left no-wrap">
                 <a href="{% url 'projects-id-metadata-id-data-update' table_id=metadata.id  %}"><i>add</i>Add data</a>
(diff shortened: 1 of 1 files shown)


signekb (commented):
Ah, that was the vision you mentioned yesterday! In some ways I like it, in others I'm not so sure. Like: wouldn't it look a bit "empty" for a user who can only import data (where the "create metadata" button won't be showed)?

philter87:
> Ah, that was the vision you mentioned yesterday! In some ways I like it, in others I'm not so sure. Like: wouldn't it look a bit "empty" for a user who can only import data (where the "create metadata" button won't be showed)?

I think we should aim for a simple/empty look, but you are right in the sense that the "import data" button is more hidden and might be hard to find.

But maybe there is a better solution. I just sensed that Omar was confused about something with the buttons, but it could have just been the "wording" in the buttons.

henrik-goul:
My gut feeling is that "add data" should have its own button, but "edit metadata" could fit well in the "more"-menu :)

signekb:
Ah, so the two "adding" actions are shown as buttons and editing the metadata could be seen as a "more" action? 
That also makes sense to me :)

signekb:
@philter87 and I just talked about that the current implementation is missing the "unfold" icon that's in the wireframes 
Untitled

lwjohnst86:
We'll keep thinking about this. Could be the buttons could be less "bold" or softer to not be too noisy. We'll look at this again after implementing more of the look and feel applied.

signekb:
@henrik-goul : It's a balance between the buttons looking active and inactive. I would like the upload icon more than the plus icon

lwjohnst86:
Given that we're doing a decent amount of refactoring and redesigning, we'll close this for now since it would require a lot of fixing to be relevant now.

signekb:
@lwjohnst86 Wouldn't this still be relevant when we start looking at the UI again?

PR #398, opened 2024-04-19
style: 💄 add theme colours based on ui wireframes and homepage look

### LINKED ISSUE #375: Start implementing bits of UI "look and feel"
Using the designs in Figma

signekb:
## Description

- This PR initiates the UI look and feel based on the homepage / `metadata-view` page. I have looked through all pages and I believe everything looks fine at this point 🎉 💄 🚀 

- There are still some details that differ from the wireframe ui (noticeably the green area behind the expanded column metadata tables that @philter87 addresses in #408), but I think this is a great starting point.  

- In the templates there are still inline styling here and there, which I think we should move out in to the `.css` files. But I haven't addressed that here. I have only moved one part of the card styling into `card.css`. 

## Related Issues

Closes #375 

## Testing

Only styling. 

## Reviewer Focus


lwjohnst86:
Ah still working on it? I was starting to review it. I notice some of the color combinations, like when in a dialog box and filling in a textarea, that they are extremely difficult for me to see.

signekb:
@lwjohnst86 Yes, sorry, I was still working on it. I had a meeting with @philter87 about how to use the colour themes in a better way than I had done previously. I force-pushed and, thereby, ignored your changes. I think I have fixed the things you addressed in the refactor 🤞

lwjohnst86:
@signekb hmm, you almost never should force push changes. I see you didn't fix what I had done, but thankfully I have my local branch so I will make a PR against this PR ☺️

signekb:
Yeah, I know it's bad practice. Sorry about that!

lwjohnst86 (approved):
Amazing!! I only made a few modifications, just formatting/styling.

PR #400, opened 2024-04-19
docs: :memo: start including a glossary of terms

lwjohnst86:
## Description

Will use this yaml file to include within Sprouts docs website.

We've talked a few times about what these terms mean, so this is the start of putting them down to eventually put them into the docs website


philter87 (approved):
Great idea with a glossary! 

I know that the vision is to include any type of data (csv, image, video, etc), but maybe it would be nice to mention a CSV file as a concrete example of 'data' and the metadata is information about the columns.

PR #403, opened 2024-04-22
feat: connect Postgres into Docker

### LINKED ISSUE #373: Save data into Postgres database (currently it's a separate SQLite database)
In `project id metadata id data create` (in draft in PR #368), the csv data is stored as a SQLite database because I didn't know enough on how to save it as a Postgres database. So we'll need to convert that code to implement a Postgres database instead.

K-Beicher:
## Description
This PR makes it possible to switch between a postgres database and a sqlite database.  The switch is done in the settings file under the section Databases.

The postgres database requires Docker Desktop to be up and running and the just command to run docker should be run prior to using 'just start-app' the first time.  Afterwards it should run as expected when the localhost is spun up.

## Related Issues

Closes #373 

## Testing

- [ ] Yes
- [x] No, I need help writing them
I'm not sure how one would go about writing tests for this.

## Reviewer Focus

There should be a check to see that localhost can still write metadata to the database and retain it.

If testing the db connection first make sure that the sqlight db is removed from your environment (use just reset-local), then run just start-docker, and just start-app.

## Checklist


config/settings.py
```diff
@@ -12,6 +12,7 @@
 import os
 from pathlib import Path
 
+# this import should be commented out if moving to postgres db
```
philter87, line 15: Remove this comment

config/settings.py
```diff
@@ -94,8 +95,21 @@
 sqlite_url = "sqlite:///db.sqlite3"
 database_url = os.environ.get("DATABASE_URL", sqlite_url)
 
+# The commented out db is the one we are running while using sqlite3
```
philter87, line 98: And remove this comment

config/settings.py
```diff
@@ -94,8 +95,21 @@
...
 DATABASES = {"default": dj_database_url.parse(database_url, conn_max_age=600)}
-
+""" DATABASES = {
```
philter87, line 100: And remove this comment

philter87 (commented):
Yay! Very good. The only thing is that we forgot to remove some comments in the settings.py file which might confuse a bit

signekb:
We have tested this on my computer and even though we start docker, `sprout` still seems connected to `sqlite` instead of a postgres db. @K-Beicher will look into it ⚡

K-Beicher:
This should now be ready to test.  I updated based on comments that I had overlooked in the first instance.  Also added a shorthand for how to change over to the postgres db.

docker-compose.yml
```diff
@@ -1,10 +1,22 @@
...
     volumes:
       - persistent_storage:/code/persistent_storage
+    environment:
+      - DATABASE_URL=postgres://primary:Create20221014@database:5432/SproutData
+  database:
+    container_name: seedcase_sprout
+    image: 'postgres'
+    restart: always
+    ports:
+      - '5432:5432'
```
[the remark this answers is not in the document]
K-Beicher: Sure, but it is failing at deployment, it is working locally.

philter87 (approved):

lwjohnst86 (approved):
We'll eventually need to actually connect the data import to go into the postgres backend, but this is a fantastic first step!! 🎉

PR #408, opened 2024-04-24
style: using divs instead of table to allow more flexibility in design

philter87:
## Description

The table is recreated with div-tags instead of table-tags. 

I was unable recreate the design in figma with table-tags, so I tried with div-tags and that worked.

Furthermore, I have reformatted the javascript code a bit

image


lwjohnst86 (commented):
@signekb originally used divs and I switched them to tables. I don't quite understand why we'd want to use divs over tables when we're displaying tabular data here. Especially if we want to sort and filter these information later. And I'm not the only one confused, there seems to be lots of posts about this: https://softwareengineering.stackexchange.com/questions/277778/why-are-people-making-tables-with-divs and https://stackoverflow.com/questions/4990819/table-based-layout-vs-div-based-layout and https://stackoverflow.com/questions/819237/div-element-instead-of-table and https://www.reddit.com/r/html5/comments/wlwaal/table_vs_div/

If we can style just fine with CSS and tables, what's the actual advantage in our use case for divs over tables? Especially when in our case we aren't using the table for just layout, but also to display tabular data (list of metadata)? Unless we go the route of `article` tags and cards like in your other PR.

philter87:
> @signekb originally used divs and I switched them to tables. I don't quite understand why we'd want to use divs over tables when we're displaying tabular data here. Especially if we want to sort and filter these information later. And I'm not the only one confused, there seems to be lots of posts about this: https://softwareengineering.stackexchange.com/questions/277778/why-are-people-making-tables-with-divs and https://stackoverflow.com/questions/4990819/table-based-layout-vs-div-based-layout and https://stackoverflow.com/questions/819237/div-element-instead-of-table and https://www.reddit.com/r/html5/comments/wlwaal/table_vs_div/
> 
> If we can style just fine with CSS and tables, what's the actual advantage in our use case for divs over tables? Especially when in our case we aren't using the table for just layout, but also to display tabular data (list of metadata)? Unless we go the route of `article` tags and cards like in your other PR.

It is a good question, and I know it seems weird, but I am just unable to replicate the design from figma with the table-tags. When I add styles to divs it does what I expect, but changing the style in table-tags just seems totally random and indeterministic.

philter87:
@lwjohnst86 I will give it another try with table-tags, because @signekb was close (and it might be close-enough if we prefer table-tags instead of pixel-perfect-design). But styling tables just seems different that styling everything else.

The main issue is the "expand"-functionality which makes it diverge from "just tabular data". I should also note that table-tags do not support sorting and filtering by default (and third-party libraries will not work out-of-the-box because the expanded content is actually a new row a the moment)

lwjohnst86:
Ooooo good point. Hmmm, do you know how easy it is to sort and filter divs? If its too difficult to implement we could use dialogs instead to show the metadata summary like you did in the cards PR.

philter87:
> Ooooo good point. Hmmm, do you know how easy it is to sort and filter divs? If its too difficult to implement we could use dialogs instead to show the metadata summary like you did in the cards PR.

I have two comments: 
1. A dialog would help a lot with the styling and we would be able to remove some javascript code (and the "show more"-button would not less important as we could just add scrolling in the dialog)
2. The sorting and filtering could be done in two ways: with javascript and with Django. I would prefer the doing it with Django. This would require a new page load for every sorting/filtering-option, but this is common when doing using a fullstack framework like Django

sprout/static/css/table.css
```diff
@@ -1,63 +1,82 @@
...
+  background-color: white;
 }
-
-.primary-table tbody {
-  background-color: var(--surface);
-  border-radius: 0px;
+.table-row {
+  display: flex;
+  flex-direction: row;
+  background-color: rgb(242, 242, 242);
```
lwjohnst86, line 18: Should this not use the color variables @signekb made in her PR?

lwjohnst86 (commented):
Small comment for now, I'll review this in more detail later.

sprout/static/css/table.css
philter87: Yes! :)

lwjohnst86 (approved):

PR #415, opened 2024-04-26
feat: ✨ single page for metadata creation steps

### LINKED ISSUE #366: Need a separate `metadata create`, rather than a redirect to `metadata update`
We need a page that is separate from updating, since we need some way of saying "ready to upload data?" or something like that. Right now it ends at "update all" (metadata). The flow stops.

philter87:
## Description

- This PR adds the stepper functionality for creating metadata. This is totally overwhelming! Sorry!

Many pages have been renamed and moved, because it is now steps instead of separate pages. You should start reading in this order: 
1. projects_id_metadata_id/create.py is the main view of the stepper, see picture below. This will call other "views" depending on a query parameter called ´step´. The first part of the stepper is rendered if step is missing or equal to one. This will call the "view" defined in "step_name_and_description.py".
2. projects_id_metadata_id/create.html is the main template, that conditionally shows the different templates depending on the "step" query parameter

image

## Related Issues

Closes #366 

See also Issues #...

## Testing

- [ ] Yes

## Reviewer Focus

## Checklist

For general documentation:

- [ ] Spell-check
    - [ ] US
    - [ ] UK
- [ ] Did the page(s) preview correctly on your machine without breaking
- [ ] New category words (keywords) (if any) added to the code snippet file


sprout/views/projects_id_metadata_id/create.py
signekb, line 1: Same as with the templatdes: I would keep the main view out of the subfolder and name it `prokects_id_metadata_create` and rename this subfolder to `porjects_id_metadata` (without `_id`)

sprout/views/projects_id_metadata_id/step_columns_update.py
```diff
@@ -1,29 +1,17 @@
...
-    database.
-
-    Args:
-        request: The HTTP request sent from the server (by the user).
-        table_id: The ``table_id`` from Tables.
-
-    Returns:
-        HttpResponse: A response given back to the server.
-    """
+def step_columns_update(request: HttpRequest, table_id: int) -> HttpResponse:
```
signekb, line 13:
```suggestion
def step_columns(request: HttpRequest, table_id: int) -> HttpResponse:
```
remove update bc this will be used for both creating and updating column metadata, right?

sprout/urls.py
```diff
@@ -13,15 +13,10 @@
     ),
     path("view", views.projects_id_view, name="projects-id-view"),
     path(
-        "metadata/<int:table_id>/create",
+        "metadata/create/",
```
signekb, line 16:
```suggestion
        "metadata/create",
```

signekb (commented):
Very, very nice, Philip! 🎉 
I have added some suggestions for refactoring (naming of files and subfolders + file placement). I have ignored docstrings for now and will add some more thorough ones later. I also did lose overview with the tests - but I think it was mostly refactoring/moving stuff around 👍

sprout/templates/projects_id_metadata_id/step-columns-update.html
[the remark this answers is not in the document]
philter87: Yes, that is better!

lwjohnst86 (approved):
This is fantastic! Like @signekb there are a few things I'd like to dig into a bit more, but we'll do that in a later refactoring stage. Let's get this going for the demo tomorrow!!

PR #469, opened 2024-06-24
build: :bricks: fix the Quarto build to follow Git root = website root

lwjohnst86:
## Description

- This PR fixes some issues with the Quarto website building. The issue was I was hoping the website root would be in the `docs/` folder. But Quarto strongly assumes and is opinionated on the location of the website root being the Git root. I was trying to fight it for a while by seeing if there were other ways to change that behaviour, but it isn't worth it. So this PR is to follow the behaviour of Git root = website root.

## Reviewer Focus

This PR only needs a quick review. Mostly moving files around.


_extensions/debruine/glossary/_extension.yml
signekb, line 1: What's debruine here?

_quarto.yml
```diff
@@ -8,7 +11,7 @@ website:
   page-footer:
     center:
       - text: "License: CC-BY 4.0"
-        href: LICENSE.md
+        href: https://github.com/seedcase-project/seedcase-sprout/blob/main/docs/LICENSE.md
```
signekb, line 14: The licence description shouldn't be rendered and to be found on the website instead of linking to the raw file on GitHub?

_extensions/debruine/glossary/_extension.yml
lwjohnst86: I believe that is to help with creating a glossary automatically. I was in the middle of testing it out, but didn't finish with incorporating it into actually creating a glossary.

_quarto.yml
lwjohnst86: The rendered license looks really weird because it isn't actually written following Markdown syntax, even though the file is a `.md` file.

signekb: Ah, alrighty!

_extensions/debruine/glossary/_extension.yml
signekb: Hmmm, alright, so you want to keep this here for now? :)

lwjohnst86: I honestly don't even remember how this was added, probably when moved over from design repo. I'll remove it from here for now.

signekb (approved):

PR #471, opened 2024-06-24
feat: :sparkles: add seedcase theme to pumls

signekb:
## Description

- This PR adds the seedcase theme to pumls (without addressing the outdated information in some of them)
- It also regenerates the pumls as svgs instead if pngs. 
- Note that I haven't edited or regenerated the C4 container and context diagrams since they have been edited in seedcase-project/design#87


lwjohnst86 (approved):
Not sure how the stepper files got included here, but we'll be refactoring everything anyway after this iteration.

PR #487, opened 2024-06-28
docs: remove naming scheme post

signekb:
## Description

- Remove the naming scheme post which will be added to the Design docs instead


K-Beicher (approved):

lwjohnst86 (changes requested):
Ummm, wait, the naming scheme specific to sprout should stay. But the general principles shouldn't

PR #490, opened 2024-07-01
docs: :memo: remove general naming style guide content, move to design repo

lwjohnst86:
## Description

This is removing the general naming guide content and keeping only naming relevant to Sprout.


martonvago (approved):
Looks good to me! 👍 👍

PR #498, opened 2024-07-08
build: :wrench: set Fly option so it doesn't turn off too fast

### LINKED ISSUE #423: Figure out the slow load times and the timeout issue with Fly
Really this is only for making demos easier. So important, but not critical. More annoyance.

lwjohnst86:
## Description

These changes are done because Fly would turn off too soon (default is 5 seconds!!), because during demos, users take some time to look around and it ends up restarting... So this hopefully should fix that.

Closes #423

## Reviewer Focus

This PR needs a quick review. Very small change!

### DIFF
--- a/fly.toml
+++ b/fly.toml
@@ -5,6 +5,10 @@
 
 app = "seedcase-sprout"
 primary_region = "ams"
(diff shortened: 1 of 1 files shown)


martonvago (approved):
Nice!

Issue #523, opened 2024-07-17
Tests are failing because file upload deletion doesn't really work


lwjohnst86:
See #524

PR #524, opened 2024-07-17
build: :bricks: add pytest and rearrange tests to make use of pytest

### LINKED ISSUE #505: Begin migrating over to pytest

### LINKED ISSUE #506: Setup infrastructure to use pytest (on top of or along with unittest?)

lwjohnst86:
## Description

These changes are adding pytest infrastructure, converting some tests over into the correct format, as well as some small updates to make working with pytest easier in VSCode.

A problem this has identified is that the file upload isn't properly deleting the file if the metadata extraction isn't finished correctly. So I made an issue here: #523

Closes #506, closes #505, and related to seedcase-project/.github#165

## Reviewer Focus

This PR needs a bit of an in-depth review. Mostly adding infrastructure to allow us to use pytest.

Unfortunately, the build fails because of an issue with deleting a file during 'upload', which I already created an issue about early. Not sure why unittest didn't pick up this issue earlier...

## Checklist

- [x] Added or updated tests
- [ ] Tests passed locally
- [ ] Build passed locally
- [ ] Updated documentation


justfile
```diff
@@ -41,12 +41,11 @@ start-app: install-deps update-migrations
 
 # Install Python package dependencies
 install-deps:
-  # no-root to not install the parent folder as a package
-  poetry install --no-root
+  poetry install
```
martonvago, line 44: Is it okay that it gives this warning when installing?
```
Warning: The current project could not be installed: No file/folder found for package seedcase-sprout
If you do not want to install the current project use --no-root.
If you want to use Poetry only for dependency management but not for packaging, you can disable package mode by setting package-mode = false in your pyproject.toml file.
In a future version of Poetry this warning will become an error!
```
I've seen this before, but I can't say I fully understand it or how to fix it (if we even need to fix it). Something to do with folder names and which folders count as packages?

martonvago (approved):
Just a question, otherwise looking great and very nice test output!

PR #529, opened 2024-07-24
fix: 🐛 debug code depended on old unittest, updated for pytest

### LINKED ISSUE #523: Tests are failing because file upload deletion doesn't really work

martonvago:
## Description

These changes fix failing tests by updating the way we check whether tests are running. This became out of date after switching to pytest. As a result, sample data was loaded into the test database, so tests checking if the `Files` table was empty were failing.

I've also added an extra test debug config allowing us to set breakpoints etc. in external files (e.g. library files or Django core code) -- feel free to ignore/change this.

Closes #523 

## Reviewer Focus

This PR needs a quick review.

## Checklist

- [x] Added or updated tests
- [x] Tests passed locally
- [x] Build passed locally
- [x] Updated documentation

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -14,6 +14,17 @@
       "args": [
         "runserver"
       ]
+    },
+    {
+      "name": "test-debugger",
+      "type": "debugpy",
+      "request": "launch",
+      "program": "${file}",
+      "purpose": [
+        "debug-test"
+      ],
+      "console": "integratedTerminal",
+      "justMyCode": false
     }
   ]
 }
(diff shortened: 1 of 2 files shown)


K-Beicher (approved):
I'm not sure if you want to push this today, or wait until Luke and Signe are back to review as well, but I'll do an approve anyway.

martonvago:
Thanks, it's not particularly urgent for me, so it can wait :)

lwjohnst86 (approved):
Neat! I'm not a big fan of having code that only works with debugging/for demo purposes within the code itself. But that's something we can come back to later.

Issue #512, opened 2024-07-13
Update models to match new project structure


martonvago:
Update models to match the new setup where metadata and data files are stored on the disk as part of `core`.

- Establish the correspondence between [frictionless metadata schemas](https://framework.frictionlessdata.io/docs/guides/describing-data.html) (data package, data resource, table) and our current Django models.
- Create custom properties in frictionless metadata schemas matching current model properties.
- Have some logic that loads properties from frictionless metadata schemas into Django models (so we can access them as if they were coming from the database).

martonvago:
Notes:
- Frictionless allows us to load the metadata for each project very easily. We just load the `datapackage.json` file in the project folder using the path to the folder, for example.
- `datapackage.json` is parsed into a `Package` object containing a list of `Resource`s. Each `Resource` has a schema with a list of `Field`s. Each `Field` defines the (data)`type` of the `Field` and any `constraints` (`required`, `unique` etc).
- This suggests the following correspondence between our Django models and classes in frictionless:
    * `Tables` = `Resource` + custom properties `is_draft`, `created_at`, `created_by`, `modified_at`, `modified_by`, `last_data_upload`, `data_rows`, `files`
    * `Files` = the objects listed under the `files` property above
    * `Columns` = `Field`+ custom property `machine_readable_name`
    * `DataType` = `type` property on `Field`
- We could have a folder structure for data and metadata where, within a `secure-storage` folder (or something equivalent), we have separate folders for each data project. Inside each data project folder we would have a single `datapackage.json` file and all the data files (e.g. `my-data1.csv`).
- Then, in the Django `app`, we could get away with storing only the data projects in a database table with a path to the project folder. All other metadata would live in `datapackage.json`. If the folder containing all data projects is known / configurable in advance, then maybe we don't even have to have a list of projects in the database.

Questions:
- Should we use `json` or `yaml` metadata files?
- Can a `Table` be composed of multiple `File`s? Most of the code suggests yes, but sometimes only one `File` is expected for a `Table`.
- Why does `Table` have an `original_file_name` property when `File` also has this property and, pending the question above, one `Table` can be composed of multiple `File`s?
- To what extent do we want to handle `Resource`s and data `File`s separately? Can a `Package` have a `Resource` without a corresponding data `File`? And vice versa, can a data `File` exist in a project folder without it being listed as a `Resource` in the `Package`?
- Is putting data file information into a `files` array on `Resource` acceptable?
[...]

PR #530, opened 2024-07-24
refactor: :sparkles: add frictionless-style sample data and metadata

### LINKED ISSUE #511: Update sample data
Update sample data (currently in `fixtures`) to meet frictionless standards and fit with the new no-database setup in `core`.

martonvago:
## Description

This adds metadata for `patients.csv` using the frictionless standard. See the Notes section of [this comment](https://github.com/seedcase-project/seedcase-sprout/issues/512#issuecomment-2245427197) for some more context.

I put this in `persistent_storage`, imagining that we would eventually want a folder structure where data projects live in separate folders in some sort of secure storage space, with each project folder containing one `datapackage.json` metadata file and any number of data files. But I'm not super sure where to actually put the project folders.

We also have a very sensible rule to gitignore data in `persistent_storage`, which I overruled in this case to add the sample data. I'm also very open to any other ideas about how we should do this.

I haven't yet removed the old sample data, as it's still used by the database in `app`. I thought we could remove it when we switch over to the new sample data.

Closes #511 

## Reviewer Focus

This PR needs more of an in-depth review. Focus on the structure of the metadata and the folder structure.

## Checklist

- [x] Added or updated tests
- [x] Tests passed locally
- [x] Build passed locally
- [x] Updated documentation


persistent_storage/projects/sample-project/datapackage.json
martonvago, line 1: This corresponds to a frictionless `Package` object. I generated it using the frictionless cli and then added custom properties.

Correspondence between our Django models and classes in frictionless:
  - `Tables` = `Resource` + custom properties `is_draft`, `created_at`, `created_by`, `modified_at`, `modified_by`, `last_data_upload`, `data_rows`, `files`
  - `Files` = the objects listed under the `files` property above
  - `Columns` = `Field`+ custom property `machine_readable_name`
  - `DataType` = `type` property on `Field`

lwjohnst86, line 1: Very nice. As per the naming doc on project files, this would go under `projects/0/` rather than `sample-project`. Or more likely, if we align with frictionless data, it would be `package/0/`.

Since we will be starting all regular projects at and above 1, we can use 0 as our example/sample.

persistent_storage/projects/sample-project/patients.csv
lwjohnst86, line 1: Following the naming guide, this would go under `projects/0/data/1/raw/<timestamp>-<uuid>.<extension>.gzip`. Or if we align with frictionless, it would be `packages/0/resources/1/`.

lwjohnst86 (changes requested):
Very nice!! This is a good starting point, and we can move our models to align with frictionless

martonvago:
I also removed custom properties, as discussed.

lwjohnst86 (approved):

PR #547, opened 2024-08-05
chore: :wrench: keep only relevant design docs on website

lwjohnst86:
## Description

These changes are done to keep out design docs that don't make sense (right now), since a lot of the docs are out of date with what we are doing.

## Reviewer Focus

This PR needs a quick review.


signekb (approved):
Makes sense!

PR #549, opened 2024-08-05
docs: :memo: diagram of internal flow of creating a resource structure

lwjohnst86:
## Description

This is a sketch design of the internals of the `create_resource_structure()` function. I like to sketch out what the inside is doing with a function, at least at a high level, with many smaller functions doing the work. I've tried to designed the functions to be relatively independent of each other and easy to test.

## Reviewer Focus

This PR needs a in-depth review.


lwjohnst86:
@seedcase-project/developers this is still a rough draft, especially the diagram image itself as it is super ugly, but I think it's a good state to get feedback. Could you all review it?

docs/design/images/function-flow.puml
```diff
@@ -1,126 +1,143 @@
...
-  projects_id_metadata_id_data_update : - check for problems in the data
-  projects_id_metadata_id_data_update : - inform user of problems in data
-  projects_id_metadata_id_data_update : - provide a file with a list of problems
-  projects_id_metadata_id_data_update : - merge new data with existing data
-
-  state "<&trash> projects <id> metadata <id> data delete" as projects_id_metadata_id_data_delete
-  projects_id_metadata_id_data_delete : - could be a button only
-  projects_id_metadata_id_data_delete : - delete all contents of the data
+state "Input" as input {
+  input : Args package_id and resource_id are not included,\nas they are only integers
```
signekb, line 8: when you create a new package, you wouldn't have to input `package_id` and `resource_id`, right?

docs/design/images/function-flow.puml
```diff
@@ -1,126 +1,143 @@
...
+  pfrr : Create the file path for the raw file
+
+  state "path_file_resource_data()" as pfrd
+  pfrd : Create the file path for the data file
+
+  state "path_file_package_database()" as pfpd
+  pfpd : Create the file path for the database file
+
+  state "path_file_package_properties()" as pfpp
+  pfpp : Create the file path for the properties file
```
signekb, line 44: is the "properties file" = the `datapackage.json` file?

docs/design/images/function-flow.puml
```diff
@@ -1,126 +1,143 @@
...
+
+  pfrr -[hidden]> pfrd
+  pfrd -[hidden]> pfpd
+  pfpd -[hidden]> pfpp
+  pfpd -[hidden]> gpd
+
+  ' Read functions -----
+  state "read_resource_properties()" as rrp
```
signekb, line 52: So this function will read the properties for a specific resource within a package? Or for all resources within a package?

docs/design/images/function-flow.puml
```diff
@@ -1,126 +1,143 @@
...
+  rrp : - Required arguments: path
+
+  state "read_package_database()" as rpd
+  rpd : - Required arguments: path
+
+  state "read_resource()" as rr
```
signekb, line 58: Will this read the descriptor, i.e., the metadata, of a specific resource?

docs/design/images/function-flow.puml
```diff
@@ -1,126 +1,143 @@
...
 
-projects_id_metadata_id -> projects_id_metadata_id_update
-projects_id_metadata_id -> projects_id_metadata_id_delete
+input --[hidden]> prc
+prc --[hidden]> output
 
-projects_id_metadata_id --> projects_id_metadata_id_data_update
-projects_id_metadata_id --> projects_id_metadata_id_data_delete
 @enduml
+
```
signekb, line 143: Do we also need some edit functions? Like when we use a data file to infer something about the resource, but the user needs to edit it (either bc validation fails or bc they don't agree with e.g., the inferred data type)?

docs/design/images/function-flow.puml
```diff
@@ -1,126 +1,143 @@
...
+  state "write_resource_data()" as wrd
+  wrd : - Required arguments: data, path
+
+  state "write_package_database()" as wpd
+  wpd : - Required arguments: json_object, path
+
+  ' Verify functions -----
+  ' Checks that it fits with data package spec
+  state "verify_resource_properties()" as vrp
```
signekb, line 84: So this will verify that the `properties` fit the data resource descriptor, right? What will the `verify_resource_data()` function at L92 do?

docs/design/images/function-flow.puml
```diff
@@ -1,126 +1,143 @@
...
+  state "verify_package_properties()" as vpp
+  vpp : - Required arguments: json_object
+  vpp : - Checks against full spec of data package
+
+  state "verify_resource_data()" as vrd
+  vrd : - Required arguments: json_object
+  vrd : - Checks against full spec of data package
+
+  ' Validate functions -----
+  ' Checks against what is expected of the data,
```
signekb, line 97: Will this not be validating the metadata of a resource against the data, like the frictionless-py [`validate` functions](https://framework.frictionlessdata.io/docs/guides/validating-data.html#validate-functions) do?

docs/design/images/page-flow.puml
signekb, line 1: Has this been edited or just renamed? I feel like I have seen this figure being renamed in another PR? :)

signekb (commented):
So nice to get this started, @lwjohnst86! Nice work! 🔥 🤩 
I have some initial thoughts/confusion, but have tried not to comment on too much so it doesn't become overwhelming. I will be back with more reviews when we have resolved these comments :)

docs/design/images/function-flow.puml
```diff
@@ -1,126 +1,143 @@
...
+  file_rid : Parquet file for tabular\ndata to act as a backup
 
-projects_id_metadata -> projects_id_metadata_create
-projects_id_metadata --> projects_id_metadata_id
+  state "/resources/<id>/raw/*.gzip" as file_rir
+  file_rir : Untouched raw file with data
+}
 
-projects_id_metadata_create --> projects_id_metadata_id
+state "package_resource_create()" as prc {
```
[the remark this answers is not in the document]
lwjohnst86: There will be a separate `package_create()` function that I will make soon. This one is to add a new resource file to an existing package.

docs/design/images/function-flow.puml
lwjohnst86: Good catch!

docs/design/images/function-flow.puml
```diff
@@ -1,126 +1,143 @@
...
+  prc : - Required arguments: package_id, resource_id, and path
+  prc : - Optional argument is properties object or manual user input
```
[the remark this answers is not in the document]
lwjohnst86: The properties object would be optional only for the CLI and if not provided would need user input to go through and confirm that the extracted metadata is correct or not. But in the app, the user input would happen on the UI, and that metadata would go into the properties object.... which also means that the `get_properties_data()` will have to be external facing since we would use that in the app to help with showing everything there... Something to think about...

docs/design/images/function-flow.puml
lwjohnst86: Yea, this is something I wasn't sure what to call it. We could stick with metadata? Properties the term is used in Data Package, so that's why I used it here.

docs/design/images/function-flow.puml
lwjohnst86: Hmmm, good question. I think all resources, so we can know which new id to assign to the new resource..?

docs/design/images/function-flow.puml
lwjohnst86: This will be the actual data. I'll append `data()` to the end.

docs/design/images/function-flow.puml
lwjohnst86: That will do very basic checks of the data, like do the data types match, etc. This is a bit confusing, as I get validation and verification mixed up... (e.g. [here](https://understandingdata.com/posts/what-is-data-validation-and-where-do-you-validate-data/#Data_Validation_vs_Verification)). Verify is compared against something that exists already and validate is comparing against standards or expectations. So in this case, checking against data types would be validation. But when someone adds new data, we would also do a verification against the existing data. So maybe here we would have to do both anyway... I'll think about that a bit more.

docs/design/images/function-flow.puml
lwjohnst86: Yea, you're right, I got the two mixed up 😛

docs/design/images/function-flow.puml
lwjohnst86: There will be a `package_resource_edit()` function, I will make that later.

docs/design/images/page-flow.puml
lwjohnst86: OOoo, I opened a PR against main instead of as a stacked PR, oops

docs/design/images/function-flow.puml
martonvago: And, on a related note, the `resource_id` is always autogenerated, right? Because we’re creating a new resource here

docs/design/images/function-flow.puml
```diff
@@ -1,126 +1,143 @@
...
-projects_id -> projects_id_delete
-projects_id --> projects_id_metadata
+  state "/resources/<id>/data.parquet" as file_rid
+  file_rid : Parquet file for tabular\ndata to act as a backup
 
-projects_id_metadata -> projects_id_metadata_create
-projects_id_metadata --> projects_id_metadata_id
+  state "/resources/<id>/raw/*.gzip" as file_rir
+  file_rir : Untouched raw file with data
+}
```
martonvago, line 27: Can one resource have multiple data files associated with it? The frictionless standard allows this, but do we want to allow it as well, or do we instead want to merge additional data for a resource into preexisting data (as implied by `docs/design/images/page-flow.svg`)? Or is the idea that whenever an additional data file is uploaded, we generate a new `gzip` in `raw`, but update the one `parquet` file?

docs/design/images/function-flow.puml
martonvago: Plus, if we have a single `datapackage.json` file then we would always read that file in its entirety, right? So kind of automatically including all resources and their properties.

docs/design/images/function-flow.puml
```diff
@@ -1,126 +1,143 @@
...
+  properties : (Optional) JSON object with\nkey-value pairs that match\ndata package spec
 }
 
-' To get proper alignment
-group_projects --[hidden]> group_data
-group_metadata --[hidden]> group_data
+state "Output" as output {
+  output : The package parent folder is shown as /
+  state "/database.sqlite" as file_sql
+  file_sql : The SQLite database with\nthe resources as tables
```
martonvago, line 17: This means that there will be a `Tables` table in the database listing all resources, right (like there is now)? Not that each resource will have its own database table?

docs/design/images/function-flow.puml
```diff
@@ -1,126 +1,143 @@
...
+  path : Path to the file with data in it
+  properties : (Optional) JSON object with\nkey-value pairs that match\ndata package spec
 }
 
-' To get proper alignment
-group_projects --[hidden]> group_data
-group_metadata --[hidden]> group_data
+state "Output" as output {
+  output : The package parent folder is shown as /
+  state "/database.sqlite" as file_sql
```
martonvago, line 16: So do `database.sqlite` and `datapackage.json` contain the same information (i.e. the metadata of the package)?

docs/design/images/function-flow.puml
signekb: Amazing 🎉

docs/design/images/function-flow.puml
martonvago: Just bumping this, because the verification/validation process is a bit confusing to me as well. E.g. whether we check the data against the supplied metadata or, vice versa, the metadata against the submitted data.

martonvago (commented):
I think this is super useful and I (hopefully) mostly understand what’s going on (i.e. when a new resource is added, we want to generate new resource-specific `gzip` and `parquet` files based on the data, and update the package-specific `sqlite` and `datapackage.json` files based on the metadata).

Maybe it would be nice to show a sequential view of the sub-functions as well, to see in what order they will be called? I’m thinking here mainly about the 4 main flows generating the 4 types of output. This would help answer questions like “where will the `id` of the new resource be generated, when updating the database or when updating `datapackage.json`?"

docs/design/images/function-flow.puml
lwjohnst86: No, the SQLite database only contains the data itself.

docs/design/images/function-flow.puml
lwjohnst86: Hmm, no, each resource's data will be a table in the database. So the data itself will be in the SQLite database.

docs/design/images/function-flow.puml
lwjohnst86: Hmm, the frictionless standard defines a resource as a single data file/table (https://datapackage.org/standard/data-resource/). Updates to the data would be merged in with the parquet and database file, rather than create a new resource. And when additional data is added, a new gzip file will be added to the raw folder to keep track of additions to the data.

docs/design/images/function-flow.puml
lwjohnst86: Yea, we could totally do that 😁 I think it would still be good to have a clear function to access explicitly the resources, rather than every time selecting the relevant field from the JSON when we want the resources only.

docs/design/images/function-flow.puml
lwjohnst86: I think I will try to find other words to use, since I also get very confused about the terms!! (went down a rabbit hole a bit with that hahah)

docs/design/images/function-flow.puml
signekb: Frictionless does allow multiple files, as described [here](https://datapackage.org/standard/data-resource/#multiple-files). I thought that we would allow data having the same properties (like e.g., two files with the same kind of data for two different individuals) to be a part of the same resource? 

I'm a bit confused about the storage of data:  For one data file, will we have one`gzip` in raw, a `parquet` file *and* the same data in a database?

docs/design/images/function-flow.puml
lwjohnst86, line 58:
```suggestion
  state "read_resource_data()" as rr
```

docs/design/images/function-flow.puml
lwjohnst86: Hmm, it allows it as a "may", but doesn't seem to recommend it. And I would agree, it gets tricky to have multiple files in that field and making it easy to later on import those data files if following the datapackage spec. But I will write more about the reasons for each of the output files in the naming scheme PR.

docs/design/images/function-flow.puml
martonvago: For sure!

docs/design/images/function-flow.puml
martonvago: Ah, okay!

docs/design/images/function-flow.puml
signekb: I think, at least for now, it might make sense to keep the naming as close to frictionless as possible to avoid confusion? :)

signekb: To be explicit ,it might be nice to add that it's the `datapackage.json` file?
```suggestion
  pfpp : Create the file path for the properties file (i.e., `datapackage.json`)
```

docs/design/images/function-flow.puml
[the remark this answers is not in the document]
signekb: Maybe it should be called `resource_create()` then? Since a resource is always a part of a package, having package in the name is kind of unnecessary?

docs/design/images/function-flow.puml
```diff
@@ -1,126 +1,143 @@
...
+
+  ' Path functions -----
+  state "path_file_resource_raw()" as pfrr
+  pfrr : Create the file path for the raw file
+
+  state "path_file_resource_data()" as pfrd
+  pfrd : Create the file path for the data file
+
+  state "path_file_package_database()" as pfpd
+  pfpd : Create the file path for the database file
```
martonvago, line 41: This question is about the entire `path_file_package_database` flow:

When we update the sqlite db, do we (1) interact with it using some sort of ORM library (i.e. issue commands through an interface for adding a table and populating it), (2) issue SQL commands to the database connection directly, or (3) use polars and dataframes somehow?

I’m asking this because I don't think I fully understand what the intermediary dataframe objects represent on the arrows `read_package_database` > `create_database_table` > `write_package_database`. I could see the data file loaded into a dataframe and that loaded into the new table, but that doesn’t quite agree with the diagram.

docs/design/images/function-flow.puml
```diff
@@ -1,126 +1,143 @@
...
+  wrr : - Required arguments: input_path, output_path
+
+  state "write_resource_properties()" as wrp
+  wrp : - Required arguments: json_object, path
+
+  state "write_resource_data()" as wrd
+  wrd : - Required arguments: data, path
+
+  state "write_package_database()" as wpd
+  wpd : - Required arguments: json_object, path
```
martonvago, line 80: What exactly is in this `json_object`? Related to my broader question about how we interact with the sqlite db, I think.

docs/design/images/function-flow.puml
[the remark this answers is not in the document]
signekb: Ah, ofc, makes sense. 

Why`get_properties_data()` and not `read_[...]`?

lwjohnst86:
@seedcase-project/developers I've updated this diagram and its ready for review now!

docs/design/images/core-python-functions/create-resource-structure.puml
martonvago, line 1: And, just to make sure I understand, the next step would be calling `create_resource_properties` with the newly created resource’s path as argument, so that’s how subsequent functions would know the `id` of the new resource. 

I guess this answers my question of "based on what do we determine new resource `id`s: `datapackage.json` entries, sqlite database tables or folders within `packages/<id>/resources`?" — it is the latter.

martonvago (commented):
☄️

docs/design/images/core-python-functions/create-resource-structure.puml
```diff
@@ -0,0 +1,62 @@
+@startuml create-resource-structure
+!theme seedcase from https://raw.githubusercontent.com/seedcase-project/seedcase-theme/main
+
+skinparam StateFontColor black
+skinparam StateFontName Fira Code
+
+state "Input" as input {
+  input : 'path_*()' functions assist\nwith giving correct paths
+' Arguments -----
+  path : Path to resource\nfolder in a package
```
martonvago, line 10: Should this be the path to the resource**s** folder?

docs/design/images/core-python-functions/create-resource-structure.puml
lwjohnst86: Just a small clarification of your question. The sqlite database won't have a table of folders at this level. In the app, there will be a Postgres table with that information, but not here.

To answer the question, it could be either datapackage.json, or the resources folder that would give the information on the ID. To stick with the other functions, giving a path works just as well.

docs/design/images/core-python-functions/create-resource-structure.puml
lwjohnst86, line 10:
```suggestion
  path : Path to resources\nfolder in a package
```

docs/design/images/core-python-functions/create-resource-structure.puml
```diff
@@ -0,0 +1,62 @@
...
+state "Output" as output {
+  output : The package parent folder is shown as /
+  output : Output: list of paths
+  state "/resources/<id>/" as folder_id
+  folder_id : Folder with assigned ID (e.g. '1')
+
+  state "/resources/<id>/raw/" as folder_raw
+  folder_raw : Raw folder with assigned ID (e.g. '1')
+  folder_id --[hidden]> folder_raw
+}
```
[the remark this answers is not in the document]
lwjohnst86: In general, for things that write to the OS, from a functional perspective, it is good to return the paths back that were created, so that you can test whether those returned paths exist or not. So yes, we include both.

docs/design/images/core-python-functions/create-resource-structure.puml
```diff
@@ -0,0 +1,62 @@
...
+  cri : Required arg: list of existing IDs
+  cri : Output: ID
+  gri --> cri
+
+  state "create_resource_dir()" as crd
+  crd : Required arg: ID to assign for directory
+  crd : Output: path of created directory
+  cri --> crd
+
+  state "create_resource_raw_dir()" as crrd
```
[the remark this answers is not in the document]
lwjohnst86: Nothing decides that. Both get created. Unless explicitly said, there is no early return for any of the functions.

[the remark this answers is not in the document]
martonvago: Okay, I think I got confused because in an activity diagram that would be an early return.

docs/design/images/core-python-functions/create-resource-structure.puml
```diff
@@ -0,0 +1,62 @@
...
+  output : Output: list of paths
+  state "/resources/<id>/" as folder_id
+  folder_id : Folder with assigned ID (e.g. '1')
+
+  state "/resources/<id>/raw/" as folder_raw
+  folder_raw : Raw folder with assigned ID (e.g. '1')
+  folder_id --[hidden]> folder_raw
+}
+
+state "create_resource_structure()" as crs {
```
signekb, line 24: Should this also create the sqlite and parquet files or will they only be created when you add data to it?

docs/design/images/core-python-functions/create-resource-structure.puml
```diff
@@ -0,0 +1,62 @@
...
+
+  state arg <<entryPoint>>
+  input --> arg
+
+  state "verify_is_directory()" as vid
+  vid : Required arg: path
+  vid : Output: path or Error
+  arg --> vid
+
+  state "get_resource_ids()" as gri
```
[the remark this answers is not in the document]
signekb: That makes sense to me at least?

docs/design/images/core-python-functions/create-resource-structure.puml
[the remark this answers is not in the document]
signekb: I understand your confusion, @martonvago, about how it's shown in the diagram with the arrows. But I also understand what you mean, @lwjohnst86, and I think it's clear (enough) to me when the output is specified as the two paths 🤷‍♀️

docs/design/images/core-python-functions/create-resource-structure.puml
```diff
@@ -0,0 +1,62 @@
...
+
+state "Input" as input {
+  input : 'path_*()' functions assist\nwith giving correct paths
+' Arguments -----
+  path : Path to resources\nfolder in a package
+}
+
+state "Output" as output {
+  output : The package parent folder is shown as /
+  output : Output: list of paths
```
signekb, line 15:
```suggestion
  output : Output: list two paths, the resource path and the resource <raw> path
```

signekb (commented):
Very nice and clear 👍 A couple of minor suggestions:

docs/design/images/core-python-functions/create-resource-structure.puml
lwjohnst86: At least in core, I would say the source of truth is the ids that make up the folder names in `resources/`. Since the `datapackage.json` is actually referring to them, not the other way around.

docs/design/images/core-python-functions/create-resource-structure.puml
lwjohnst86: Hmm, only when the data is added. Doesn't make sense to create something that has nothing.

docs/design/images/core-python-functions/create-resource-structure.puml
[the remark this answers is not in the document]
lwjohnst86: @martonvago thankfully this isn't an activity diagram 😉 😛 But I understand the confusion ☺️

martonvago (dismissed):

docs/design/images/core-python-functions/create-resource-structure.puml
```diff
@@ -0,0 +1,62 @@
+@startuml create-resource-structure
+!theme seedcase from https://raw.githubusercontent.com/seedcase-project/seedcase-theme/main
+
+skinparam StateFontColor black
+skinparam StateFontName Fira Code
+
+state "Input" as input {
+  input : 'path_*()' functions assist\nwith giving correct paths
+' Arguments -----
+  path : Path to resources\nfolder in a package
```
signekb, line 10: The svg says "resource" not "resources". Maybe it needs to be updated?

lwjohnst86: Yes, once you've all approved, I'll update the svg before merging

signekb: Perfect, just wanted to make sure you wouldn't forget :)

signekb (dismissed):

PR #556, opened 2024-08-12
docs: :memo: diagram for flow of create resource properties

lwjohnst86:
## Description

These changes add a function flow for the create resource properties function.

## Reviewer Focus

This PR needs a in-depth review.


docs/design/images/core-python-functions/create-resource-properties.puml
```diff
@@ -0,0 +1,92 @@
...
+  erp : - Output: JSON
+  vsf --> erp
+
+  state "prompt_confirm_properties()" as pcp
+  pcp : - Required arg: json
+  pcp : - Output: JSON
+  pcp : - Prompt user to fill in\n  missing details
+  pcp : - Likely has internal\n  verification
+  erp --> pcp
+  pcp --> vrp
```
lwjohnst86, line 84: Thinking about this some more, this doesn't actually work in a Python code setting, but would work in the CLI/App setting.

docs/design/images/core-python-functions/create-resource-properties.puml
```diff
@@ -0,0 +1,92 @@
...
+  state condition <<choice>>
+  arg_properties --> condition
+  
+  state "verify_resource_properties()" as vrp
+  vrp : - Required arg: properties
+  vrp : - Output: JSON or Error
+  vrp : - Checks against spec
+  condition --> vrp : If properties given
+
+  state "add_resource_properties()" as arp
```
martonvago, line 55: What exactly does this function do? I don't think it adds the resource properties to the package because we have `write_resource_properties(json_object, path)` for that. Does it just add the path to the parquet file to `properties`? In that case I would move "Add `path`\nproperty" inside the box from the arrow.

docs/design/images/core-python-functions/create-resource-properties.puml
```diff
@@ -0,0 +1,92 @@
...
+  state "prompt_confirm_properties()" as pcp
+  pcp : - Required arg: json
+  pcp : - Output: JSON
+  pcp : - Prompt user to fill in\n  missing details
+  pcp : - Likely has internal\n  verification
+  erp --> pcp
+  pcp --> vrp
+
+  state "return" <<exitPoint>>
+  vrp --> return
```
martonvago, line 87: What does this arrow mean? Isn't "return `properties`" indicated by the other arrow (`arp --> return`)?

docs/design/images/core-python-functions/create-resource-properties.puml
```diff
@@ -0,0 +1,92 @@
...
+  condition --> vif : If data_path given
+
+  state "verify_supported_formats()" as vsf
+  vsf : - Required arg: path
+  vsf : - Output: path or Error
+  vif --> vsf
+
+  state "extract_resource_properties()" as erp
+  erp : - Required arg: path
+  erp : - Output: JSON
```
martonvago, line 75: Here and elsewhere JSON stands for `properties`, right? If so, writing that would make it easier to follow maybe.

docs/design/images/core-python-functions/create-resource-properties.puml
```diff
@@ -0,0 +1,92 @@
...
+  path --> arg_path
+  data_path --> arg_data_path
+  properties --> arg_properties
+
+  state "verify_is_directory()" as vid
+  vid : - Required arg: path
+  vid : - Output: path or Error
+  arg_path --> vid
+
+  state "path_root_is_package()" as papr
```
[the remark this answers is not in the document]
martonvago: And at what point is the parquet file actually created, before this step?

docs/design/images/core-python-functions/create-resource-properties.puml
```diff
@@ -0,0 +1,92 @@
...
+  vid --> papr
+
+  state condition <<choice>>
+  arg_properties --> condition
+  
+  state "verify_resource_properties()" as vrp
+  vrp : - Required arg: properties
+  vrp : - Output: JSON or Error
+  vrp : - Checks against spec
+  condition --> vrp : If properties given
```
martonvago, line 53: What if both `properties` and `data_path` are given? And if neither?

martonvago (changes requested):
Looks great, thanks!!
Sorry for going a bit overboard with the comments 😅

docs/design/images/core-python-functions/create-resource-properties.puml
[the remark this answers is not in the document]
lwjohnst86: We've already had experience with these types of diagrams (well, @K-Beicher and @signekb have mostly) and they are really shitty to create, update, and look so damn ugly. State is the best that I found that provides the right balance of flexibility and features. Also, I think at this level it is overkill for making a simple sketch of what the internals will be.

docs/design/images/core-python-functions/create-resource-properties.puml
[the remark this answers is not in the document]
lwjohnst86: Yea, I struggled with naming this function. The path relative one works fine, though.

I'm not super happy with this solution that I made, since it makes a dependency that the datapackage file is two directories up (or down?).

The Parquet file would be created later.

lwjohnst86, line 41:
```suggestion
  state "path_relative_resource_data()" as prrd
```

docs/design/images/core-python-functions/create-resource-properties.puml
```diff
@@ -0,0 +1,92 @@
...
+  data_path --> arg_data_path
+  properties --> arg_properties
+
+  state "verify_is_directory()" as vid
+  vid : - Required arg: path
+  vid : - Output: path or Error
+  arg_path --> vid
+
+  state "path_root_is_package()" as papr
+  papr : - Required arg: absolute path
```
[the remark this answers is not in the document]
lwjohnst86: Yea, exactly.

lwjohnst86, line 42:
```suggestion
  prrd : - Required arg: absolute path from input
```

docs/design/images/core-python-functions/create-resource-properties.puml
```diff
@@ -0,0 +1,92 @@
...
+  properties --> arg_properties
+
+  state "verify_is_directory()" as vid
+  vid : - Required arg: path
+  vid : - Output: path or Error
+  arg_path --> vid
+
+  state "path_root_is_package()" as papr
+  papr : - Required arg: absolute path 
+  papr : - Output: relative path\n  (e.g. `resource/1/data.parquet`)
```
lwjohnst86, line 43:
```suggestion
  prrd : - Output: relative path\n  (e.g. `resource/1/data.parquet`)
```

docs/design/images/core-python-functions/create-resource-properties.puml
```diff
@@ -0,0 +1,92 @@
...
+  vid --> papr
```
lwjohnst86, line 44:
```suggestion
  vid --> prrd
```

docs/design/images/core-python-functions/create-resource-properties.puml
```diff
@@ -0,0 +1,92 @@
...
+  state "verify_resource_properties()" as vrp
+  vrp : - Required arg: properties
+  vrp : - Output: JSON or Error
+  vrp : - Checks against spec
+  condition --> vrp : If properties given
+
+  state "add_resource_properties()" as arp
+  arp : - Required arg: json 
+  arp : - Output: JSON
+  papr --> arp : Add `path`\nproperty
```
lwjohnst86, line 58:
```suggestion
  prrd --> arp : Add `path`\nproperty
```

docs/design/images/core-python-functions/create-resource-properties.puml
```diff
@@ -0,0 +1,92 @@
...
+  papr : - Output: relative path\n  (e.g. `resource/1/data.parquet`)
+  vid --> papr
+
+  state condition <<choice>>
+  arg_properties --> condition
+  
+  state "verify_resource_properties()" as vrp
+  vrp : - Required arg: properties
+  vrp : - Output: JSON or Error
+  vrp : - Checks against spec
```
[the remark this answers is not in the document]
lwjohnst86: Yup! More likely ☺️

docs/design/images/core-python-functions/create-resource-properties.puml
lwjohnst86: An error then 😛

docs/design/images/core-python-functions/create-resource-properties.puml
lwjohnst86, line 55:
```suggestion
  state "edit_resource_properties_field()" as arp
```

docs/design/images/core-python-functions/create-resource-properties.puml
```diff
@@ -0,0 +1,92 @@
...
+  state "add_resource_properties()" as arp
+  arp : - Required arg: json 
+  arp : - Output: JSON
+  papr --> arp : Add `path`\nproperty
+  vrp --> arp
+
+  arg_data_path --> condition
+
+  state "verify_is_file()" as vif
+  vif : - Required arg: path
```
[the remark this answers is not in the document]
lwjohnst86: Hmm, I see what you mean, but this function could be used in other places, so doesn't make sense to call it `data_path`.

docs/design/images/core-python-functions/create-resource-properties.puml
lwjohnst86: Hmm, I understand the confusion, but properties isn't a data type that you can output in Python. Both path and JSON are.

docs/design/images/core-python-functions/create-resource-properties.puml
```diff
@@ -0,0 +1,92 @@
...
+  pcp : - Output: JSON
+  pcp : - Prompt user to fill in\n  missing details
+  pcp : - Likely has internal\n  verification
+  erp --> pcp
+  pcp --> vrp
+
+  state "return" <<exitPoint>>
+  vrp --> return
+  arp --> return
+  return --> output
```
lwjohnst86, line 89:
```suggestion
  state "join" <<join>>
  state "return" <<exitPoint>>
  vrp --> join
  arp --> join
  join --> return
  return --> output
```

@martonvago how is this?

docs/design/images/core-python-functions/create-resource-properties.puml
```diff
@@ -0,0 +1,92 @@
...
+state "Output" as output {
+  state "JSON" as json_output
+  json_output : - JSON object of the resource following\n  the Data Package spec
+  json_output : - Use `write_resource_properties()` to\n  save back to `datapackage.json`
+}
+
+state "create_resource_properties()" as crp {
+  crp : - Required arg: path
+  crp : - Conditional args: data_path or properties object
+  crp : - Needs user input if data_path given
```
lwjohnst86, line 26:
```suggestion
  crp : - Conditional args: data_path or properties object
  crp : - Either data_path or properties must be given, not both
```

docs/design/images/core-python-functions/create-resource-properties.puml
lwjohnst86: It would be adding a field to the properties. Maybe edit would be a better word.

docs/design/images/core-python-functions/create-resource-properties.puml
lwjohnst86: @martonvago love it! Yea, I think that makes more sense. Also means we will need to expose a few more functions, like extractors and validators, in the core functions.

lwjohnst86, line 84:
```suggestion
```

docs/design/images/core-python-functions/create-resource-properties.puml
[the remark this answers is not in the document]
martonvago: I see what you mean about the dependency, but I guess that’s just how we designed the folder structure, so it’s kinda okay? (And I guess technically we could locate the `datapackage.json` of a package without knowing where it is nested under the package’s root directory.)

Ah okay, I see, true, we can add the path to the parquet file to `properties` without that file actually existing.

docs/design/images/core-python-functions/create-resource-properties.puml
martonvago: Ah, what field? So not the path to the parquet file? (Sorry 😅 )

docs/design/images/core-python-functions/create-resource-properties.puml
martonvago: Yeah, okay, I think I initially read this more in a conceptual way, if that makes sense

martonvago: Also, sorry for unresolving, but another related JSON question that has been at the back of my mind: 
So a piece of information in JSON is technically a string, right? So in Python this would actually be deserialised into a dictionary or a list, right?

So when we work with these JSON inputs/outputs in Python, would we pass them around as a string and deserialise and serialise them every time we process them, or would we actually pass them around as a dictionary/list?

docs/design/images/core-python-functions/create-resource-properties.puml
martonvago: And I guess this would not arise anyway if we followed something like the design sketched out in the comment above about confirming properties with the user.

docs/design/images/core-python-functions/create-resource-properties.puml
martonvago: Wait, so maybe I misunderstood what the output of `edit_resource_properties_field()` is. Is it the final `properties` of the resource? Because, in that case, I would have a single arrow pointing to `return`: the one coming from `edit_resource_properties_field()`. Because the `properties` output by `verify_resource_properties()` has yet to go through `edit_resource_properties_field()`. But I may well be misunderstanding this.

Based on my understanding, I would also expect an arrow from `extract_resource_properties()` to `edit_resource_properties_field()`.

docs/design/images/core-python-functions/create-resource-properties.svg
martonvago, line 1: Does this need to be updated?

docs/design/images/core-python-functions/create-resource-properties.puml
martonvago: Yes, my feeling would also be to pass them around as dicts because they're obviously easier to read, manipulate etc. And yes we could serialise them back to a JSON string whenever we needed to using that library.

(I also have some vague memories that `frictionless-py` actually handles lots of this (de)serialisation anyway.)

lwjohnst86: Yes, we would use the json package to handle these things so that they are always JSON objects ☺️

docs/design/images/core-python-functions/create-resource-properties.puml
lwjohnst86: Yes, that is the field. The function would make it easier to edit a single field within the spec without needing to provide the full spec each time we want to make an edit.

lwjohnst86:
@martonvago I've updated the diagram to make the arrows clearer about where they are going. Does that help?

martonvago:
> @martonvago I've updated the diagram to make the arrows clearer about where they are going. Does that help?

Yes!!! To me that makes sense 100% now, thanks :D

martonvago (dismissed):

docs/design/images/core-python-functions/create-resource-properties.puml
lwjohnst86: Technically it would be a Python dictionary I guess. But always being parsed/validated as JSON

K-Beicher (commented):
Looks good!

docs/design/images/core-python-functions/create-resource-properties.puml
```diff
@@ -0,0 +1,87 @@
...
+  state "extract_resource_properties()" as erp
+  erp : - Required arg: path
+  erp : - Output: JSON
+  vsf --> erp
+
+  state condition_output <<choice>>
+  erp --> condition_output
+  vrp --> condition_output
+
+  state "edit_resource_properties_field()" as erpf
```
signekb, line 76: Should this not be `create_`?

signekb (changes requested):

docs/design/images/core-python-functions/create-resource-properties.puml
lwjohnst86: No, because the field already exists. So we're editing something that is already there, though probably blank.

signekb (dismissed):

PR #578, opened 2024-08-19
feat: ✨ add `verify_is_dir()`

### LINKED ISSUE #562: Create and test `verify_is_directory` function
- signature: `verify_is_directory(path: Path) -> Path`
- arguments:
  - `path`: the path to verify
- returns `path` (same as in the input) if `path` points to a directory
- raises `NotADirectoryError` if `path` does not point to a directory

signekb:
## Description

These changes are done to start the work of creating the `create_resource_structure()`.

Closes #562 

## Reviewer Focus

This PR needs an in-depth review. Focus on:

1. Do we want to use `pathlib` and not `os`?  --> Yes.
2. I have added a new file `verify_is_directory.py` since I wasn't sure how we want to structure helper functions like this one. Where should it be located? --> This is fine for now.
3. Tests: 
    1. How do we want to name them? --> We don't need the name of the function in the name of the test as long as we keep one test file per function. Then the function name will be in the name of the test file. 
    2. Am I missing a test case?

## Checklist

- [X] Added or updated tests
- [X] Tests passed locally
- [X] Updated documentation (yes: docstrings)


tests/core/test_verify_is_directory.py
```diff
@@ -0,0 +1,26 @@
+import pytest
```
signekb, line 1:
```suggestion
from pytest import raises
```

tests/core/test_verify_is_directory.py
```diff
@@ -0,0 +1,26 @@
...
+
+
+def test_verify_is_directory_raises_not_a_directory_error(tmp_path):
+    """Test that the verify_is_directory function raises a NotADirectoryError
+    if the directory does not exist.
+
+    Args:
+        tmp_path: Temporary directory created by pytest
+    """
+    with pytest.raises(
```
signekb, line 22:
```suggestion
    with raises(
```

sprout/core/verify_is_directory.py
```diff
@@ -0,0 +1,19 @@
...
+
+
+def verify_is_directory(path: Path) -> Path:
+    """Verifies if the path is an existing directory.
+
+    Args:
+        path: The path to verify
+
+    Raises:
+        NotADirectoryError: If the path is not an existing directory
```
lwjohnst86, line 11:
```suggestion
    Raises:
        NotADirectoryError: When the directory in the path doesn't exist.
```

Just some small grammar/language change.

sprout/core/verify_is_directory.py
```diff
@@ -0,0 +1,19 @@
+from pathlib import Path
+
+
+def verify_is_directory(path: Path) -> Path:
+    """Verifies if the path is an existing directory.
```
lwjohnst86, line 5:
```suggestion
    """Verifies that directory given by the path exists or not.
```

sprout/core/verify_is_directory.py
```diff
@@ -0,0 +1,19 @@
...
+
+    Args:
+        path: The path to verify
+
+    Raises:
+        NotADirectoryError: If the path is not an existing directory
+
+    Returns:
+        path if it's a directory
```
lwjohnst86, line 14:
```suggestion
        A path object if it is a directory.
```

tests/core/test_verify_is_directory.py
```diff
@@ -0,0 +1,26 @@
...
+    assert verify_is_directory(tmp_path) == tmp_path
+
+
+def test_verify_is_directory_raises_not_a_directory_error(tmp_path):
+    """Test that the verify_is_directory function raises a NotADirectoryError
+    if the directory does not exist.
+
+    Args:
+        tmp_path: Temporary directory created by pytest
+    """
```
lwjohnst86, line 21:
```suggestion
    """Test that the verify_is_directory function raises a NotADirectoryError
    if the directory does not exist."""
```

tests/core/test_verify_is_directory.py
martonvago, line 1: Maybe include a test for when the path points to a file?

lwjohnst86 (changes requested):
Great start! Just some small comments.

martonvago (changes requested):
Nice, I didn't know about `tmp_path` 😁

sprout/core/verify_is_directory.py
martonvago: But "Verifies if/whether...", right? If it includes an "or"

tests/core/test_verify_is_directory.py
signekb: Ah, good idea. Will do 👍

tests/core/test_verify_is_directory.py
```diff
@@ -0,0 +1,19 @@
...
+    assert verify_is_directory(tmp_path) == tmp_path
+
+
+def test_verify_is_directory_raises_not_a_directory_error(tmp_path):
+    """Test that the verify_is_directory function raises a NotADirectoryError
+    if the directory does not exist."""
+    with raises(
+        NotADirectoryError,
+        match=r"/non_existent_directory is not an existing directory",
```
signekb, line 17:
```suggestion
        match=r"/non_existent_directory doesn't exist.",
```

tests/core/test_verify_is_directory.py
```diff
@@ -0,0 +1,19 @@
...
+    """
+    assert verify_is_directory(tmp_path) == tmp_path
+
+
+def test_verify_is_directory_raises_not_a_directory_error(tmp_path):
+    """Test that the verify_is_directory function raises a NotADirectoryError
+    if the directory does not exist."""
+    with raises(
+        NotADirectoryError,
+        match=r"/non_existent_directory doesn't exist.",
```
signekb, line 17:
```suggestion
        match=r"/non_existent_directory directory doesn't exist.",
```

signekb:
@lwjohnst86 Ah, I see what you mean. Thanks for the tip!

tests/core/test_verify_is_dir.py
```diff
@@ -0,0 +1,34 @@
...
+    """Test that the verify_is_dir function returns the directory if it exists."""
+    assert verify_is_dir(tmp_path) == tmp_path
+
+
+def test_raises_error_with_non_existent_directory(tmp_path):
+    """Test that the verify_is_dir function raises a NotADirectoryError
+    if the directory does not exist."""
+    with raises(
+        NotADirectoryError,
+        match=r"/non_existent_directory directory doesn't exist.",
```
lwjohnst86, line 16: Is this message required? It feels like a brittle way of running the tests since if we make a change to the function, we also need to make a change to the test here.

lwjohnst86 (changes requested):

tests/core/test_verify_is_dir.py
signekb: You're right, it's not actually needed. We could either remove the `match` line altogether or we could only look for the "/non_existent_directory" match, like so: 
```suggestion
        match=r"/non_existent_directory",
```

tests/core/test_verify_is_dir.py
```diff
@@ -0,0 +1,34 @@
+from pytest import raises
+
+from sprout.core.verify_is_dir import verify_is_dir
+
+
+def test_returns_existing_directory(tmp_path):
+    """Test that the verify_is_dir function returns the directory if it exists."""
```
signekb, line 7:
```suggestion
    """Test the path is returned if the directory exists."""
```

tests/core/test_verify_is_dir.py
```diff
@@ -0,0 +1,34 @@
...
+    assert verify_is_dir(tmp_path) == tmp_path
+
+
+def test_raises_error_with_non_existent_directory(tmp_path):
+    """Test that the verify_is_dir function raises a NotADirectoryError
+    if the directory does not exist."""
```
signekb, line 13:
```suggestion
    """Test that NotADirectoryError is raised if the directory doesn't exist."""
```

tests/core/test_verify_is_dir.py
```diff
@@ -0,0 +1,34 @@
...
+    with raises(
+        NotADirectoryError,
+        match=r"/non_existent_directory directory doesn't exist.",
+    ):
+        verify_is_dir(tmp_path / "non_existent_directory")
+
+
+def test_returns_error_with_file(tmp_path):
+    """Test that the verify_is_dir function raises a NotADirectoryError
+    if the directory does not exist."""
```
signekb, line 23:
```suggestion
    """Test that NotADirectoryError is raised when a path to a file is given as input."""
```

tests/core/test_verify_is_dir.py
signekb: I have implemented the suggestion above and kept only the last part of the input path in `match` 👍

lwjohnst86 (approved):
Lovely! So I guess the match is partially needed, at least to pick up the folder name.

PR #582, opened 2024-08-20
feat: ✨ add `get_ids()` function

### LINKED ISSUE #563: Create and test `get_ids` function
- signature: `get_ids(path_to_resources_or_package_folder: Path) -> list[int]`
- arguments:
  - `path`: path to the resources folder within a package or path to packages folder
- returns a list of existing IDs or the empty list if there is no folders with IDs

signekb:
## Description

These changes add the internal `get_ids()` function.

Closes #563 

## Reviewer Focus

## Checklist

- [X] Added or updated tests
- [X] Tests passed locally


tests/core/test_get_ids.py
```diff
@@ -0,0 +1,42 @@
...
+        tmp_path: Temporary directory created by pytest
+    """
+    (tmp_path / "1").mkdir()  # single digit
+    (tmp_path / "20").mkdir()  # double digit
+    (tmp_path / "resources").mkdir()  # directory without digits
+    (tmp_path / "test1").mkdir()  # directory with digits
+    (tmp_path / "datapackage.json").mkdir()  # file without digits
+    (tmp_path / "test1.json").mkdir()  # file with digits
+
+    assert sorted(get_ids(tmp_path)) == [1, 20]
```
signekb, line 19: I sort here to ensure the output is in the expected order

sprout/core/get_ids.py
```diff
@@ -0,0 +1,27 @@
...
+
+    Raises:
+        NotADirectoryError: If the path is not a directory
+
+    Returns:
+        A list of integers representing the resource ids.
+        If no IDs are found, an empty list is returned.
+    """
+    if not path.is_dir():
+        raise NotADirectoryError(f"{path} is not an existing directory")
```
signekb, line 19: I have added this like in the `verify_is_directory()` function to ensure `path` is a directory. But maybe that's actually not necessary here?

sprout/core/get_ids.py
```diff
@@ -0,0 +1,32 @@
...
+
+    Args:
+        path: Directory to search for IDs.
+
+    Returns:
+        A list of integers representing the ids of the subdirectories.
+        If no IDs are found, an empty list is returned.
+    """
+    # Keep only directories
+    directories = list(path.glob("*/"))
```
lwjohnst86, line 16: The previous code didn't actually get only folders with numbers in them. So I revised it so the first step is keeping only directories.

sprout/core/get_ids.py
```diff
@@ -0,0 +1,32 @@
...
+
+def get_number(path: Path) -> int:
+    """Get only the number from directory.
+
+    Args:
+      path: Directory to extract the number from.
+    """
+    directory_name = path.name
+    if re.match(r"^\d+$", directory_name):
+        return int(directory_name)
```
lwjohnst86, line 32: I created this function because the previous code didn't catch some test cases I added (see below). I also didn't understand some of the code, like why `.group(0)` was needed.

lwjohnst86 (changes requested):
Very nice! I made some changes directly in the code because there was a test case that the code you wrote didn't catch. So I refactored it to fix it. Let me know what you think!

sprout/core/get_ids.py
```diff
@@ -0,0 +1,32 @@
...
+    # Keep only directories
+    directories = list(path.glob("*/"))
+    ids = list(map(get_number, directories))
+    # Drop any empty items
+    ids = list(filter(None, ids))
+
+    return ids
+
+
+def get_number(path: Path) -> int:
```
martonvago, line 24:
```suggestion
def get_number(path: Path) -> int | None:
```
Do we need this?

sprout/core/get_ids.py
```diff
@@ -0,0 +1,32 @@
...
+    """Get only the number from directory.
+
+    Args:
+      path: Directory to extract the number from.
+    """
```
martonvago, line 29: Do we need `Returns:...`?

tests/core/test_get_ids.py
```diff
@@ -0,0 +1,54 @@
...
+def test_only_dirs_gotten(tmp_path):
+    """Test that the get_ids function returns a list of resources."""
+    (tmp_path / "1").mkdir()  # single digit
+    (tmp_path / "datapackage.json").touch()  # file without digits
+
+    assert get_ids(tmp_path) == [1]
+
+
+def test_empty_list_when_no_ids(tmp_path):
+    """Test that the get_ids function returns the directory if it exists."""
```
martonvago, line 19: Is this description about this function?

tests/core/test_get_ids.py
```diff
@@ -0,0 +1,54 @@
...
+def test_different_numbers_output(tmp_path):
+    """Test that the get_ids function returns numbers of any size."""
+    (tmp_path / "1").mkdir()
+    (tmp_path / "20").mkdir()
+    (tmp_path / "999").mkdir()
+
+    assert sorted(get_ids(tmp_path)) == [1, 20, 999]
+
+
+def test_get_only_dirs_with_numbers(tmp_path):
```
martonvago, line 48: Could be merged with `test_only_numbers_output`?

tests/core/test_get_ids.py
martonvago, line 1: We could have a test for when there's a file with a numeric name? And for extensionless files?

martonvago (changes requested):

tests/core/test_get_ids.py
lwjohnst86: Oops, it was copy pasted from an old test.

lwjohnst86, line 19:
```suggestion
    """Return an empty list when there are no IDs."""
```

sprout/core/get_ids.py
lwjohnst86, line 29:
```suggestion
      path: Directory to extract the number from.
    
    Returns:
    	An single integer for the ID.
    """
```

lwjohnst86 (changes requested):

lwjohnst86 (approved):
Very nice :grin:

PR #584, opened 2024-08-20
docs: get_resource_ids -> get_ids

signekb:
## Description

These changes are done so the diagram is updated with the more generalised function `get_ids()` instead of `get_resource_ids()`

## Reviewer Focus

This PR needs a quick review.


martonvago (commented):
Ah so I've also included this change with the rest of the updates in https://github.com/seedcase-project/seedcase-sprout/pull/589

lwjohnst86 (approved):

PR #585, opened 2024-08-20
feat: :sparkles: add `create_next_id()`

### LINKED ISSUE #564: Create and test `create_next_id` function
- signature: `create_next_id(existing_ids: list[int]) -> int`
- arguments:
  - `existing_ids`: a list of IDs of all existing entities within a collection of entities
- returns the next ID in the sequence based on the existing IDs

martonvago:
## Description

This adds the `create_next_id()` function.

Closes #564 

## Reviewer Focus

This PR needs a quick review.

## Checklist

- [x] Added or updated tests
- [x] Tests passed locally
- [x] Build passed locally
- [x] Updated documentation


sprout/core/generate_next_id.py
martonvago, line 1: I'm not doing any checks on the argument here (because we're using type hinting and `mypy`), but let me know if I should.

tests/core/test_generate_next_id.py
martonvago, line 1: Because of not explicitly checking for input type in the implementation, I'm also not testing for errors in the tests. I think it would be inconsistent to check e.g. that `generate_next_id([1, 2, "three"])` throws a `TypeError` (which it does because of `max(...)`), but not to check that it rejects `False` as an input (which it doesn't because it is falsey). But again, let me know if I should do it differently.

sprout/core/generate_next_id.py
```diff
@@ -0,0 +1,10 @@
+def generate_next_id(existing_ids: list[int]) -> int:
```
signekb, line 1: To simplify - i.e., not use a new action - should we call this `create_next_id()` instead?

sprout/core/generate_next_id.py
signekb: I think it makes sense to omit checks like that. 
These functions would not be called on their own, but always be called by the outer function, e.g., `create_resource_structure()`, right? 
But I'm not sure where we should put limit as to how much we should test for functions like this...

signekb (commented):
Very nice!

sprout/core/generate_next_id.py
martonvago: Sure, done!

tests/core/test_create_next_id.py
```diff
@@ -0,0 +1,18 @@
+from sprout.core.create_next_id import create_next_id
+
+
+def test_create_next_id_creates_first_id_correctly():
+    """Given no existing IDs, it should return 1.
+    """
```
lwjohnst86, line 6:
```suggestion
    """Given no existing IDs, it should return 1."""
```

tests/core/test_create_next_id.py
```diff
@@ -0,0 +1,18 @@
...
+    assert create_next_id([]) == 1
+
+def test_create_next_id_creates_second_id_correctly():
+    """Given one existing ID, it should return that ID + 1.
+    """
+    assert create_next_id([1]) == 2
```
lwjohnst86, line 12:
```suggestion
def test_create_next_id_creates_second_id_correctly():
    """Given one existing ID, it should return that ID + 1."""
    assert create_next_id([1]) == 2
```

lwjohnst86 (changes requested):
🎉 Very nice, only some small edits!

lwjohnst86 (approved):

signekb (approved):

PR #586, opened 2024-08-20
feat: ✨ add `create_dir()` and `create_dirs()` with tests

### LINKED ISSUE #566: Create and test `create_dirs` function
- signature: `create_dirs(path: list[Path]): -> list[Path]`
- arguments:
  - `path`: A list of paths pointing to the new directory to create
- returns the path to the newly created directory (same as in the input)

K-Beicher:
## Description

These changes are done to make the internal function that creates directories.

Closes #566 

## Reviewer Focus

This PR needs an in-depth review. The first function I've made, I'm fairly sure I've done it correctly.

## Checklist

- [x] Added or updated tests
- [x] Tests passed locally
- [ ] Build passed locally
- [x] Updated documentation (docstrings)


tests/core/test_create_dir.py
```diff
@@ -0,0 +1,39 @@
+from sprout.core.create_dir import create_dir
+
+
+def test_create_dir_existing_folder(tmp_path):
+    """Test that the create_dir function checks for existing folder and returns the name
+    if the folder already exists.
+
+    Args:
+        tmp_path: Temporary directory created by pytest
+    """
```
lwjohnst86, line 10:
```suggestion
    """Checks for an existing folder and returns the name
    if the folder already exists.
    """
```

Don't need the documentation for arguments in tests

lwjohnst86 (changes requested):
Very nice! Only very small edits for the docstrings 🎉

sprout/core/create_dirs.py
```diff
@@ -0,0 +1,31 @@
+from pathlib import Path
+
+
+def create_dirs(path: list[Path]) -> list[Path]:
```
lwjohnst86, line 4: I converted this function to create multiple directories when given a list of paths. I pushed these changes rather than through the PR suggestions since I needed to test that it worked as I expected, plus because it involved several changes.

lwjohnst86 (changes requested):
Very nice! I made some small changes, mainly in allowing the function to create several directories at once rather than one at a time.

K-Beicher:
@seedcase-project/developers Just thinking about this function, and what it should do, if it encounters a directory that already exists.

1/ does it go back and informs the user that it is already there (and may already contain data)

2/ does it just confirm that the directory is created, and ignores the fact that it wasn't created now.

This becomes a bit more tricky if we are talking about more than one directory (as Luke's mod allows), what happens if one of several directories exists.

1/ goes back and informs about the one that is already there and then:
1a/ doesn't create the rest of them
1b/ creates the rest of the directories

2/ creates the new directories and confirms that all directories are now in place.

PS We need more tests if it is to run more than one directory at a time.

sprout/core/create_dirs.py
```diff
@@ -0,0 +1,35 @@
...
+        path: The path pointing to the new directory to create
+
+    Raises:
+        NoDirectoryError: If the directory wasn't created
+
+    Returns:
+        path to the newly created directory.
+    """
+    if path.is_dir():
+        return path
```
martonvago, line 32: Is this different from doing `path.mkdir(parents=True, exist_ok=True)`? Because it will just skip existing directories (and not recreate them), right?

K-Beicher: I wasn't actually sure about it skipping them, and in the meantime I've realised that we may want to talk about what we want to do if the directory is actually already there... Do we want to let the user know that the name is already potentially in use?

lwjohnst86:
To answer the first question, this is an internal function so no user interaction is expected so we don't need to inform on anything.

My feeling is, if any directory already exists, it throws an error about creating it so no directories are created in the end. Which would mean we would need to run `verify_is_dir()` before actually creating the dirs.

sprout/core/create_dirs.py
```diff
@@ -0,0 +1,35 @@
+from pathlib import Path
+
+
+def create_dirs(path: list[Path]) -> list[Path]:
+    """Creates directories from a list of directories.
```
signekb, line 5:
```suggestion
    """Creates directories from a list of paths.
```

sprout/core/create_dirs.py
```diff
@@ -0,0 +1,35 @@
...
+
+    Args:
+        path: A list of paths pointing to where directories should be created.
+
+    Raises:
+        NoDirectoryError: If the directory wasn't created
```
signekb, line 11: Should this be NotADirectoryError?
```suggestion
        NotADirectoryError: If the directory wasn't created
```

sprout/core/create_dirs.py
```diff
@@ -0,0 +1,35 @@
...
+
+
+def create_dir(path: Path) -> Path:
+    """Creates a directory from a path.
+
+    Args:
+        path: The path pointing to the new directory to create
+
+    Raises:
+        NoDirectoryError: If the directory wasn't created
```
signekb, line 26:
```suggestion
        NotADirectoryError: If the directory wasn't created
```
?

tests/core/test_create_dirs.py
signekb, line 1: Do we need tests for if the input is a not a directory? Like a file path, `tmp_path / "file.py" ?

signekb (changes requested):

sprout/core/create_dirs.py
K-Beicher: I think it is short for 'no directory was created', so No is correct...?

tests/core/test_create_dirs.py
K-Beicher: We definitely need more tests!

sprout/core/create_dirs.py
signekb: I was thinking more in the sense that if we think about when this function would raise an error, it would be if the input `path` is not a directory, right? 

`NotADirectoryError` is also a built-in exception, see the docs [here](https://docs.python.org/3/library/exceptions.html#NotADirectoryError), whereas the `NoDirectoryError` is a custom exception. And it might be nicer to use the built-in exceptions when it makes sense for us to do so :)

tests/core/test_create_dirs.py
lwjohnst86: @signekb when making directories, the computer doesn't care about the name. So the "extension" would be added to the created directory: `package/file.py/`. Extensions are just arbitrary names, it just so happens that computers often use the ending of a file name to decide which program to use to open that particular file. But if you remove the, e.g., `.zip` file extension from a file, the file itself remains a zip file. With directories, at the computer level they are classified as directories, not as files, so the computer will always interpret them as directories regardless of the file name. This type of test would still pass and wouldn't add any useful information for us.

signekb: Ah, yes! I actually discovered that yesterday as well 😅 But thanks for elaborating!

K-Beicher:
@signekb I think I've got the error handling to work now.

tests/core/test_create_dirs.py
```diff
@@ -0,0 +1,61 @@
+from sprout.core.create_dirs import create_dirs
+
+
+def test_creating_existing_directory(tmp_path):
+    """Checks for an existing directory and returns the name
```
K-Beicher, line 5: @lwjohnst86 This fails when testing locally, do we need it?

martonvago (changes requested):
Nice!! 😁 Just some comments

sprout/core/create_dirs.py
```diff
@@ -0,0 +1,47 @@
...
+    Raises:
+        NotADirectoryError: If the directory wasn't created
+
+    Returns:
+        path to the newly created directory.
+    """
+    try:
+        path.mkdir(parents=True, exist_ok=False)
+    except FileExistsError:
+        raise NotADirectoryError(f"Failed to create directory: {path}")
```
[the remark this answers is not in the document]
K-Beicher: You are right, I'm trying to catch the situation where the directory wasn't created even though the mkdir was executed. I'll see if I can rewrite to make that more clear.

sprout/core/create_dirs.py
```diff
@@ -0,0 +1,47 @@
...
+    return path
+
+
+def create_dirs(path: list[Path]) -> list[Path]:
```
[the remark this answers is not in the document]
K-Beicher: This wasn't written by me, but by @lwjohnst86. I'm going to assume that the plural won't mess up the code and correct it.

sprout/core/create_dirs.py
```diff
@@ -0,0 +1,47 @@
...
+    """Creates directories from a list of paths using the function create_dir.
+
+    Args:
+        path: A list of paths pointing to where directories should be created.
+
+    Raises:
+        NotADirectoryError: If any of the directories fail to be created, and
+        removes any directories already created.
```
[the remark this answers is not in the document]
K-Beicher: I'll get that re-worded.

tests/core/test_create_dirs.py
```diff
@@ -0,0 +1,61 @@
...
+    result = create_dirs(new_dir)
+
+    # Check if the dir is created
+    assert new_dir[0].exists()
+
+    # Check if the result is equal to the new directory
+    assert result[0] == new_dir[0]
+
+    # Check that only one directory is created
+    assert len(result) == len(new_dir)
```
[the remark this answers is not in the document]
K-Beicher: Again, this bit was made by @lwjohnst86, I'll leave this one for him to resolve.

tests/core/test_create_dirs.py
```diff
@@ -0,0 +1,61 @@
...
+
+
+def test_create_two_dirs(tmp_path):
+    """Create a new directory and returns the path correctly."""
```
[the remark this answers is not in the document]
K-Beicher: It is, I'll get it sorted.

tests/core/test_create_dirs.py
```diff
@@ -0,0 +1,61 @@
...
+
+    # Check if the dirs are created
+    assert new_dirs[0].exists()
+    assert new_dirs[1].exists()
+
+    # Check if the result is equal to the new directory
+    assert result == new_dirs
+
+    # Check that two directories were created
+    assert len(result) == len(new_dirs)
```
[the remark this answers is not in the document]
K-Beicher: again, I'll leave this one to @lwjohnst86.

sprout/core/create_dirs.py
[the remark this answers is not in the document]
martonvago: And do we want to catch the error here because we want to change what kind of error it is? Because `FileExistsError` already includes the path in the error message, so one option is to just let that bubble up to the caller without catching it here.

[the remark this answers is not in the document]
K-Beicher: Not entirely sure, sorry - over to @lwjohnst86 and @signekb

sprout/core/create_dirs.py
```diff
@@ -0,0 +1,47 @@
...
+    """
+    created_dirs = []
+    for path in paths:
+        try:
+            created_dirs.append(create_dir(path))
+        except NotADirectoryError:
+            # Clean up already created directories
+            for dir_path in created_dirs:
+                dir_path.rmdir()
+            raise
```
lwjohnst86, line 46:
```suggestion
		created_dirs = [create_dir(path) for path in paths]
```

Just to keep it cleaner 😁  and since we don't need to do the error anymore.

tests/core/test_create_dirs.py
lwjohnst86: True. So we could rewrite this test then.

tests/core/test_create_dirs.py
```diff
@@ -0,0 +1,61 @@
...
+    for dir in temp_dirs:
+        dir.mkdir()
+
+    result = create_dirs(temp_dirs)
+
+    # Check if the result is equal to the existing paths
+    assert result == temp_dirs
+
+    # Check the number of directories created
+    assert len(result) == len(temp_dirs)
```
lwjohnst86, line 23:
```suggestion
		with pytest.raises(FileExistsError):
				create_dirs(temp_dirs)
```

tests/core/test_create_dirs.py
```diff
@@ -0,0 +1,61 @@
+from sprout.core.create_dirs import create_dirs
```
lwjohnst86, line 1:
```suggestion
import pytest
from sprout.core.create_dirs import create_dirs
```

lwjohnst86 (changes requested):
Nice discussions here! I've made some suggested changes that I think will clean up things and answer/address many of @martonvago comments.

tests/core/test_create_dirs.py
```diff
@@ -0,0 +1,33 @@
...
+    ]
+
+    result = create_dirs(new_dirs)
+
+    # Check if the dirs are created
+    assert new_dirs[0].exists()
+    assert new_dirs[1].exists()
+
+    # Check if the result is equal to the new directory
+    assert result == new_dirs
```
signekb, line 33: I just tested a bit locally what happens if this function gets an existing directory as input path: It raises `FileExistsError`. @lwjohnst86 is that the wanted behaviour? 

I have added a passing test here. 

```suggestion
    assert result == new_dirs

def test_raise_error_with_existing_dir(tmp_path):
    """Raise FileExistsError when the input path is an existent directory."""
    # Create a temporary directory
    new_dir = [tmp_path / "new_dir"]
    new_dir[0].mkdir()

    with raises(
        FileExistsError,
    ):
        create_dirs(new_dir)

```

sprout/core/create_dirs.py
```diff
@@ -0,0 +1,29 @@
+from pathlib import Path
+
+
+def create_dir(path: Path) -> Path:
+    """Creates a directory from a path.
+
+    The path has already been verified as not existing.
```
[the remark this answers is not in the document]
K-Beicher: That may be me not being entirely clear. As I understand conversations with @lwjohnst86 the idea is that a 'verify' function has been run and confirmed that there is no directory of that name (ie path), so it will need to be created. Thus there is no reason for throwing an error if the directory can't be created, which is what I was trying to say with this sentence.

tests/core/test_create_dirs.py
lwjohnst86: I think that's fine. Since it is an internal function, we can assume we won't create things that already exist. But yea, this test is fine 👍🏻

sprout/core/create_dirs.py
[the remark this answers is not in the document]
lwjohnst86: @K-Beicher has it right 🦾 We will rarely use this function without running verify first. It doesn't need to be within this function since we'll often be using verify for other purposes. So we don't need to run it twice.

lwjohnst86 (dismissed):

tests/core/test_create_dirs.py
```diff
@@ -0,0 +1,61 @@
...
+    assert result == temp_dirs
+
+    # Check the number of directories created
+    assert len(result) == len(temp_dirs)
+
+
+def test_create_single_dir(tmp_path):
+    """Create a new directory and returns the path correctly."""
+    # Create a temporary directory
+    new_dir = [tmp_path / "new_dir"]
```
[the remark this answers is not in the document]
martonvago: For me I think it comes down to there being something semantically incorrect about calling a list object a `dir` because it creates false expectations about what methods I can call on it or what I can assert about it. The same for indexing into that list: `dir[0]` just doesn’t read as an object referring to a directory, it reads as some constituent part of a directory. But all this is very very parenthetical of course 😁

martonvago (changes requested):
Just a missing import

martonvago (approved):

signekb (approved):

PR #592, opened 2024-08-22
feat: :sparkles: add `verify_format_supported()`

### LINKED ISSUE #572: Create and test `verify_format_supported` function
- signature: `verify_format_supported(path: Path) -> Path`
- arguments:
  - `path`: path pointing to the (data) file whose format we want to check
- returns `path` (same as in the input) if `path` points to a data file whose format is among the formats supported by Frictionless (and, therefore, by Sprout)
- Raises an error if the file format is not supported. We should decide what error to raise, some options: `FrictionlessException` or a custom error.
- For the function logic, we'll also need to decide whether to defer the format check to Frictionless or keep a list of supported formats ourselves, against which we can check the file’s format.

martonvago:
## Description

This adds the `verify_format_supported` function.

Closes #572 

## Reviewer Focus

This PR needs a quick review.

## Checklist

- [x] Added or updated tests
- [x] Tests passed locally
- [x] Build passed locally
- [x] Updated documentation


sprout/core/verify_format_supported.py
```diff
@@ -0,0 +1,34 @@
+from pathlib import Path
+
+SUPPORTED_FORMATS = ["csv", "tsv", "xlsx", "xls", "json", "jsonl", "ndjson", 
+                     "geojson", "topojson", "ods", "parq", "parquet"]
```
martonvago, line 4: I’ve included the ones where the Frictionless [page](https://v4.framework.frictionlessdata.io/docs/references/plugins-reference) explicitly says that they have stable support for the format, plus parquet (which wasn’t mentioned on that page), but there are many others with varying levels of support ([code](https://github.com/frictionlessdata/frictionless-py/tree/main/frictionless/formats), [docs](https://github.com/frictionlessdata/frictionless-py/tree/main/docs/formats)).

martonvago: For now, I put them here as a list, but we may well move this somewhere else later or change it into an enum etc.

sprout/core/verify_format_supported.py
```diff
@@ -0,0 +1,34 @@
...
+
+class UnsupportedFormatError(Exception):
```
martonvago, line 6: Not sure if we want to have an errors/exceptions folder, so just leaving it here for now.
Also not sure if we have any preference about calling it an error vs an exception -- they seem to be used somewhat interchangeably in Python?

sprout/core/verify_format_supported.py
```diff
@@ -0,0 +1,34 @@
...
+    Raises:
+        UnsupportedFormatError: raised if the file format is not supported
+
+    Returns:
+        the path pointing to the file, if the file format is supported
+    """
+    format = path.suffix[1:]
+    if format not in SUPPORTED_FORMATS:
+        raise UnsupportedFormatError(format)
+    return path
```
martonvago, line 34: I couldn't find a Frictionless function that checked the format, so I'm doing it "by hand". In fact, it seems like Frictionless will process any file regardless of the extension, but obviously it can't extract the (table) schema from all files.

sprout/core/verify_format_supported.py
```diff
@@ -0,0 +1,34 @@
...
+    """Raised if file format is not supported by Sprout."""
+    def __init__(self, format, *args, **kwargs):
+        """Initialises UnsupportedFormatError.
+
+        Args:
+            format: the extension of the unsupported format
+            *args: non-keyword arguments
+            **kwargs: keyword arguments
+        """
+        message = f"File format '{format}' is not supported."
```
signekb, line 16: Would it be helpful to add the formats we actually support in the message?

sprout/core/verify_format_supported.py
signekb: I'm not sure how many custom exceptions we'll need eventually, but maybe it's fine to keep it here for now? 

I think when we have created the helper functions for `create_package_structure()`, `create_package_properties()` as well as the one's were working on now (`create_resource_structure()` and `create_resource_properties()`), we'll have a better overview of what kind of folder structure we need. 
At that time, we could spend some time to refactor accordingly? :)

signekb (approved):
Very nice work 🎉

sprout/core/verify_format_supported.py
signekb, line 1: Following the name of `verify_is_file()` and `verify_is_dir()`, could/should this function actually be named `verify_is_supported_format()`?

signekb, line 1: Following the name of `verify_is_file()` and `verify_is_dir()`, could/should this function actually be named `verify_is_supported_format()`?

signekb (changes requested):

sprout/core/verify_format_supported.py
martonvago: Sure!

sprout/core/verify_format_supported.py
```diff
@@ -0,0 +1,34 @@
...
+        super().__init__(message, *args, **kwargs)
+
+def verify_format_supported(path: Path) -> Path:
```
lwjohnst86:
```suggestion
def verify_is_format_supported(path: Path) -> Path:
```

tests/core/test_verify_format_supported.py
```diff
@@ -0,0 +1,21 @@
+from pytest import mark, raises
+
+from sprout.core.verify_format_supported import (
+    SUPPORTED_FORMATS,
+    UnsupportedFormatError,
+    verify_format_supported,
```
lwjohnst86:
```suggestion
    verify_is_supported_format,
```

tests/core/test_verify_format_supported.py
```diff
@@ -0,0 +1,21 @@
+from pytest import mark, raises
+
+from sprout.core.verify_format_supported import (
```
lwjohnst86:
```suggestion
from sprout.core.verify_is_supported_format import (
```

You'll need to rename the files after though.

lwjohnst86 (approved):
Very nice!!!

PR #593, opened 2024-08-22
fix: :bug: rename own csv folder to avoid clash with built-in csv

### LINKED ISSUE #590: Rename `sprout/core/csv` folder
We have to rename the `sprout/core/csv` folder (if we are keeping it) because it's clashing with the Python `csv` module when libraries do imports.

martonvago:
## Description

This renames the `sprout/core/csv` folder because it was clashing with the Python `csv` module when libraries (well, Frictionless) did imports.

I guess eventually we might just delete our `csv` things, but in the meantime.

Closes #590 

## Reviewer Focus

This PR needs a quick review.

## Checklist

- [x] Added or updated tests
- [x] Tests passed locally
- [x] Build passed locally
- [x] Updated documentation


lwjohnst86 (approved):
nice! Yea, I was worried about that name clash. There have been other cases like this, which means we should probably be a bit aware of any similar issues.

signekb (approved):

PR #594, opened 2024-08-22
feat: :sparkles: add `extract_properties_from_file()`

### LINKED ISSUE #573: Create and test `extract_resource_properties` function
- signature: `extract_resource_properties(path_to_data_file: Path) -> dict`
- arguments:
  - `path_to_data_file`: path pointing to the data file
- Returns a dictionary object containing the resource properties / metadata. These are conform to the Fricitonless standard and determined based on the data file.
- use `frictionless-py` to extract these properties and convert them to a dictionary

martonvago:
## Description

This adds the `extract_properties_from_file()` function.

Closes #573 

## Reviewer Focus

This PR needs a quick review.

## Checklist

- [x] Added or updated tests
- [x] Tests passed locally
- [x] Build passed locally
- [x] Updated documentation

### DIFF
--- a/sprout/core/extract_resource_properties.py
+++ b/sprout/core/extract_resource_properties.py
@@ -1,0 +1,15 @@
+from pathlib import Path
+
+from frictionless import describe
+
+
+def extract_resource_properties(path: Path) -> dict:
+    """Extracts resource properties from the provided data file.
+
+    Args:
+        path: the path pointing to the data file
+
+    Returns:
+        a dictionary of resource properties
+    """
+    return describe(path).to_dict()
--- a/tests/core/test_extract_properties_from_file.py
+++ b/tests/core/test_extract_properties_from_file.py
@@ -1,7 +1,7 @@
 import csv
 import json
 
-from sprout.core.extract_resource_properties import extract_resource_properties
+from sprout.core.extract_properties_from_file import extract_properties_from_file
 
 
 def test_extracts_properties_from_empty_csv(tmp_path):
@@ -22,7 +22,7 @@ def test_extracts_properties_from_empty_csv(tmp_path):
     }
 
     # when + then
-    assert extract_resource_properties(file_path) == expected_properties
+    assert extract_properties_from_file(file_path) == expected_properties
 
 
 def test_extracts_properties_from_complete_csv(tmp_path):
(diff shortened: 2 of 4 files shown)


sprout/core/extract_resource_properties.py
martonvago, line 1: This will extract resource properties, including, if possible, a table schema with basic information on the fields/columns (name, data type). It looks like more detailed information (e.g. constraints, whether a string is an enum) have to be added by hand.

tests/core/test_extract_resource_properties.py
martonvago, line 1: - Given that we're passing a file of the correct format to this function (and, in any case, all kinds of file formats are processed by `describe`), I couldn't make it throw an error in an obvious way, so I'm not testing for that.
- I'm also not sure if testing for all our supported formats here makes sense (for some we would need libraries to construct the data files), so for now I just added csv and json.
- Once we have more tests like these, it might make sense to commonise the data file creation logic.

lwjohnst86 (changes requested):
This is a fantastic! Given that we'll need to expose this function (see #605), this function could be renamed to `extract_properties_from_file()` as per the internal diagram given in the linked PR.

tests/core/test_extract_properties_from_file.py
```diff
@@ -0,0 +1,78 @@
...
+    with open(file_path, mode="w", newline="") as file:
+        writer = csv.writer(file)
+        writer.writerows(data)
+
+    # when + then
+    assert extract_properties_from_file(file_path) == expected_properties
+
+
+def test_extracts_properties_from_json(tmp_path):
+    """Given a JSON file, it should extract resource properties without a schema."""
```
signekb, line 60: Oh, so frictionless-py can't  extract the schema from`json`?

signekb (approved):
Looks great 🎉 🎉

tests/core/test_extract_properties_from_file.py
martonvago: I think the schema might be for tabular data specifically and I guess you wouldn't give that kind of data as json? But I still have some reading to do about what frictionless-py can do with what formats 😅

lwjohnst86 (approved):
🎉

tests/core/test_extract_properties_from_file.py
signekb: Same! I think you're right :)

PR #596, opened 2024-08-23
feat: :sparkles: add `create_relative_resource_data_path()`

### LINKED ISSUE #575: Create and test `path_relative_resource_data` function
- signature: `path_relative_resource_data(path_to_resource: Path) -> Path`
- arguments:
  - `path_to_resource`: path pointing to a specific resource folder
- returns a relative path pointing to the parquet file of the given resource, relative to the package folder, i.e. `resources/<resource_id>/data.parquet`
- note that the parquet file doesn’t actually exist yet at this point

signekb:
## Description

Closes #575 

## Reviewer Focus

This PR needs an in-depth review. Focus on:

- I have renamed the function from `path_relative_resource_data()` to `create_relative_resource_data_path()` to fit with the naming scheme. Thoughts? 
- This function actually does two things: 1) create a relative path to the resource and 2) add `data.parquet` to the path. Is this alright?

## Checklist

- [X] Added or updated tests
- [X] Tests passed locally

### DIFF
--- a/sprout/core/create_relative_resource_data_path.py
+++ b/sprout/core/create_relative_resource_data_path.py
@@ -1,5 +1,4 @@
 from pathlib import Path
-from re import split
 
 
 def create_relative_resource_data_path(path: Path) -> Path:
@@ -12,6 +11,4 @@ def create_relative_resource_data_path(path: Path) -> Path:
         Relative path from the package root to the resource data file
         E.g., "resources/1/data.parquet"
     """
-    relative_resource_path = split(r"packages/[0-9]*/", str(path))[1]
-
-    return Path(relative_resource_path) / "data.parquet"
+    return Path(*path.parts[-2:]) / "data.parquet"
--- a/sprout/core/create_relative_resource_data_path.py
+++ b/sprout/core/create_relative_resource_data_path.py
@@ -1,0 +1,17 @@
+from pathlib import Path
+from re import split
+
+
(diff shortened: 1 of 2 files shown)


sprout/core/create_relative_resource_data_path.py
```diff
@@ -0,0 +1,17 @@
...
+    Args:
+        path: Absolute path to the folder of a specific resource
+
+    Returns:
+        Relative path from the package root to the resource data file
+        E.g., "resources/1/data.parquet"
+    """
+    relative_resource_path = split(r"packages/[0-9]*/", str(path))[1]
+
+    return Path(relative_resource_path) / "data.parquet"
```
martonvago, line 17: Maybe something like `Path("resources") / path.name / "data.parquet"` would be simpler? Or there's `Path(*path.parts[-2:]) / "data.parquet"` if we don't want to hardcode folder names.
And then you can't break it with a silly path like `packages/22/packages/100/resources/100`

martonvago (changes requested):
Nice ⭐ !!

sprout/core/create_relative_resource_data_path.py
signekb: Ah, good idea! Thanks! 🎈

martonvago (approved):

lwjohnst86 (approved):
Amazing! And yes, I think it's fine that it adds the data to the end.

PR #599, opened 2024-08-23
docs: :memo: description of the function to create the properties template

lwjohnst86:
## Description

These changes add a description for actually getting the Data Package Spec from the source. The aim of this function is that we use it to download a copy on the sprout core and use that for the other functions, like when making a package or adding resources.

## Reviewer Focus

This PR needs a quick review.

### DIFF
--- a/docs/design/python-functions.qmd
+++ b/docs/design/python-functions.qmd
@@ -303,13 +303,14 @@ These functions are still being developed.
 
 :   Create a prettier, human readable version of a JSON object.
 
-`create_package_properties_template()`
+`create_properties_template(path)`
 
-:   TODO
-
-`create_resource_properties_template()`
-
-:   TODO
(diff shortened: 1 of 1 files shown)


signekb (approved):
👍 👍 
Just one question: Will this function create the properties with an empty `resources` field? And how would that look since `name` and `path` or `data` are required? Would it be e.g. a few lines of inline example data?

martonvago (approved):
👍 👍 
My only question is the one I asked on #600

lwjohnst86:
@martonvago See my comment at https://github.com/seedcase-project/seedcase-sprout/pull/600/files#r1731141697

@signekb this file will only be a template, so doesn't need all fields filled in. It would only be important if we used the frictionless package (or something that reads in the correct spec). We aren't doing that, so "required" is only when used as an actual data package (rather than a template).

PR #600, opened 2024-08-23
docs: :building_construction: function diagram for `create_properties_template()`

lwjohnst86:
## Description

These changes add the diagram for creating the properties template of the official Data Package Spec. I think it's possible to download the full spec from one URL. But maybe this will need two functions, one for the package and one for the resources. Thoughts? Alternatively, I could add a URL or options arg to select which spec to download, the package or the resource.

## Reviewer Focus


docs/design/images/core-python-functions/create-properties-template.puml
```diff
@@ -0,0 +1,41 @@
+@startuml create-properties-template
+!theme seedcase from https://raw.githubusercontent.com/seedcase-project/seedcase-theme/main
+
+state "create_properties_template()" as cpt {
+  cpt : - Required arg: path to where to save file
+
+  state arg <<entryPoint>>
+
+  state "DATA_PACKAGE_SPEC_URL" as url 
+  url : - Variable of the URL to the\n  Data Package Spec
```
martonvago, line 10: By Data Package Spec we mean a sample `datapackage.json` (and not the [schema](https://specs.frictionlessdata.io/schemas/data-package.json) that specifies the structure of that file), right? Where would we download this from? I can see various examples on the Frictionless website, but I couldn't find a big "master" example.

martonvago (commented):
Nice!! Just a question

docs/design/images/core-python-functions/create-properties-template.puml
lwjohnst86: That would be the standard here: https://datapackage.org/profiles/2.0/datapackage.json from this website: https://datapackage.org/standard/data-package/

docs/design/images/core-python-functions/create-properties-template.puml
```diff
@@ -0,0 +1,41 @@
...
+
+  state "scrape_url()" as su
+  su : - Required arg: url
+  su : - Output: JSON
+  su : - Scrapes the JSON using web API
+  url --> su : string
+
+  state "drop_property_field()" as dpf
```
signekb, line 18:
```suggestion
  state "drop_properties_field()" as dpf
```
I think you mean the `properties` field on ≈L9 [here](https://datapackage.org/profiles/2.0/datapackage.json), right?

docs/design/images/core-python-functions/create-properties-template.puml
```diff
@@ -0,0 +1,41 @@
...
+  dpf : - Required arg: properties, field
+  dpf : - Output: JSON
+  dpf : - Drop any fields we don't use
+  su --> dpf : JSON object
+
+  state "edit_property_field()" as epf
```
signekb, line 24: ?? Or is this also expected to be run multiple times, one time for each property field to edit?
```suggestion
  state "edit_properties_field()" as epf
```

docs/design/images/core-python-functions/create-properties-template.puml
```diff
@@ -0,0 +1,41 @@
...
+  url : - Variable of the URL to the\n  Data Package Spec
+
+  state "scrape_url()" as su
+  su : - Required arg: url
+  su : - Output: JSON
+  su : - Scrapes the JSON using web API
+  url --> su : string
+
+  state "drop_property_field()" as dpf
+  dpf : - Required arg: properties, field
```
signekb, line 19: So, the `properties` arg is the JSON `datapackage` object and the `field` is `properties`?

docs/design/images/core-python-functions/create-properties-template.puml
```diff
@@ -0,0 +1,41 @@
...
+  dpf : - Output: JSON
+  dpf : - Drop any fields we don't use
```
signekb, line 21: "Any fields we don't use" - does that include other fields than the `properties` field? If yes; maybe this function needs a new name?
Or is it bc we expect to run this function multiple times to get rid of each field we don't use individually?

docs/design/images/core-python-functions/create-properties-template.puml
```diff
@@ -0,0 +1,41 @@
...
+  su --> dpf : JSON object
+
+  state "edit_property_field()" as epf
+  epf : - Required arg: properties,\n  field, value
+  epf : - Output: JSON
+  epf : - Edit field values to\n  defaults we use
+  dpf --> epf : JSON object
+
+  state "write_file()" as wf
```
signekb, line 30: Oh, I expected the template to not be saved as a file right away, but only saved to the `datapackage.json` after it had been edited?

signekb (commented):
Very nice ⭐ 
I might be a bit confused about this function - hopefully the answers to my questions can help in that regard 😅 

For me, intuitively, it makes sense to keep it as one function and then add an arg for whether it's the resource or package props that should be kept. But I guess what's the best solution depends on the context of when it's used (which I don't have the full overview of).

docs/design/images/core-python-functions/create-properties-template.puml
martonvago: Ohhhh I see!

martonvago:
If this function will be used to update a locally saved / cached sample `datapackage.json` file (as opposed to being run every time that file viewed, for example), then I also think that it works as a single function. Because it prepares a properties file that includes package and resource properties, so any `view...` function can expose the part of the object it needs (without redownloading anything). But yeah, I'm also not sure if I understand the context 100%.

docs/design/images/core-python-functions/create-properties-template.puml
lwjohnst86: It's the second one.

docs/design/images/core-python-functions/create-properties-template.puml
lwjohnst86: Ha, yea, I can see how that might be confusing. No, not that property field. Property as in what we've been talking about the whole time, as in the data package JSON file with it's metadata and key-value (or field-value) pairing. It would be singular, not plural because it only drops one field. But I guess we could give it a list...

Also, maybe it would be clearer to call it: `drop_json_fields()`?

docs/design/images/core-python-functions/create-properties-template.puml
lwjohnst86: Maybe rename this to `edit_json_field()`?

docs/design/images/core-python-functions/create-properties-template.puml
lwjohnst86: Hm, not sure I understand. What do you mean?

signekb: Maybe I have mis-/not-fully-understood the use of the templates. I thought that the initial template (with no input, just the standard template) wouldn't be saved but rather returned to the user, so they could edit it. And then, after e.g., name and description was filled out, it would be written/saved. But I guess not?

docs/design/images/core-python-functions/create-properties-template.puml
signekb: Yeahh, maybe that's clearer? But I also understand what you meant now, so maybe `property` is fine too?
I noticed that you switched from singular to plural of `field(s)`. How come?

docs/design/images/core-python-functions/create-properties-template.puml
signekb, line 1: You have skipped the input and output states in this diagram. Shouldn't they be included to fit with the other diagrams?

docs/design/images/core-python-functions/create-properties-template.puml
```diff
@@ -0,0 +1,41 @@
...
+  wf : - Required arg: string and path
```
signekb, line 31: Maybe?
```suggestion 
  wf : - Required arg: JSON and path
```

lwjohnst86:
@martonvago related to the conversation during the meeting about frictionless and getting the official schema, I can't find any function within the package that does that:
 https://framework.frictionlessdata.io/docs/framework/actions.html 

The closest seems to be the `Package` class: https://framework.frictionlessdata.io/docs/framework/package.html but even then, it doesn't look like it provides the full template, only that it takes a data file and makes a package from it. Which isn't exactly what we want.

docs/design/images/core-python-functions/create-properties-template.puml
lwjohnst86: I decided that since the input and output are very simple (a path to save the file), it didn't make sense to add more to the diagram than was necessary. It made sense in the other diagrams because there were more inputs and the outputs needed a bit more explanation. But in this case, not really.

docs/design/images/core-python-functions/create-properties-template.puml
lwjohnst86: Ok, I'll switch to property fields. I changed to fields because if we drop one, we can easily make the function drop more.

docs/design/images/core-python-functions/create-properties-template.puml
lwjohnst86: I think this was cleared up during the meeting.

martonvago:
> @martonvago related to the conversation during the meeting about frictionless and getting the official schema, I can't find any function within the package that does that: https://framework.frictionlessdata.io/docs/framework/actions.html
> 
> The closest seems to be the `Package` class: https://framework.frictionlessdata.io/docs/framework/package.html but even then, it doesn't look like it provides the full template, only that it takes a data file and makes a package from it. Which isn't exactly what we want.

Ah okay! Yeah, I was thinking of `Package` maybe being able to do it somehow, but then not really

PR #602, opened 2024-08-23
docs: :memo: add extract action to naming scheme

lwjohnst86:
## Description

These changes are done to the naming scheme by adding an action: extract. This is because this is needed to be exposed by the Python package so that we can use it in the web app. More detail on how it will connect with the web app when I get to the design of the web app.

## Reviewer Focus

This PR needs a quick review.


K-Beicher (approved):
Looks good.

martonvago (approved):

PR #605, opened 2024-08-23
docs: :building_construction: diagram of internal code for `extract_resource_properties()`

lwjohnst86:
## Description

I split these out from the diagram @signekb started on, after moving out the extraction of properties function (see #602).

## Reviewer Focus


docs/design/images/core-python-functions/extract-resource-properties.puml
```diff
@@ -0,0 +1,69 @@
...
+
+  state "verify_is_supported_format()" as vsf
+  vsf : - Required arg: path
+  vsf : - Output: path or Error
+  vif --> vsf : path object
+
+  state "extract_properties_from_file()" as epff
+  epff : - Required arg: path
+  epff : - Output: JSON
+  vsf --> epff : JSON object
```
martonvago, line 50:
```suggestion
  vsf --> epff : path object
```
Right?

martonvago (changes requested):
Amazing, makes sense to me 😁

docs/design/images/core-python-functions/extract-resource-properties.puml
signekb: Would make sense :)

signekb (commented):
It also makes sense to me! I agree with @martonvago 's comment :)

PR #606, opened 2024-08-23
docs: :memo: include descriptions and details for the `path` functions

lwjohnst86:
## Description

These changes add the descriptions and explanations for all the path functions

## Reviewer Focus

### DIFF
--- a/docs/design/python-functions.qmd
+++ b/docs/design/python-functions.qmd
@@ -258,11 +258,11 @@ follows the Frictionless Data specification.
     `resource_id` to complete the correct path for the specific
     package/resource.
 -   All of these functions output a path object.
--   The path's returned by all of these functions are paths that exist,
+-   The paths returned by all of these functions are paths that exist,
     so all include a `verify_is_dir()` or `verify_is_file()` check.
 -   If the wrong `package_id` or `resource_id` is given, an error
     message will include a list of all the actual `package_id` at the
-    `path_sprout_root()` location or `resource_id` for a specific
+    `path_sprout_root()` location or all actual `resource_id` for a specific
     package.
 
 ### `path_resources(package_id)`
--- a/docs/design/python-functions.qmd
+++ b/docs/design/python-functions.qmd
@@ -253,45 +253,61 @@ follows the Frictionless Data specification.
 
 ## Path functions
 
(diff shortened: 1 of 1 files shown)


docs/design/python-functions.qmd
```diff
@@ -253,45 +253,61 @@ follows the Frictionless Data specification.
...
-::: callout-warning
-These functions are still being developed.
-:::
+-   Nearly all the below `path_*()` functions use `path_sprout_root()`
+    internally to get the root path as well as `package_id` and/or
+    `resource_id` to complete the correct path for the specific
+    package/resource.
+-   All of these functions output a path object.
+-   The path's returned by all of these functions are paths that exist,
+    so all include a `verify_is_dir()` or `verify_is_file()` check.
```
signekb, line 262: Do we need a `verify_exists()` function? Currently, the `verify_is_dir()` only checks whether the path is a directory or not - not whether it exists?

docs/design/python-functions.qmd
```diff
@@ -253,45 +253,61 @@ follows the Frictionless Data specification.
...
+-   If the wrong `package_id` or `resource_id` is given, an error
+    message will include a list of all the actual `package_id` at the
+    `path_sprout_root()` location or `resource_id` for a specific
```
signekb, line 265:
```suggestion
    `path_sprout_root()` location or all actual `resource_id` for a specific
```

signekb (commented):
Nice! 🎉  Two comments:

docs/design/python-functions.qmd
martonvago: Doesn't it throw if the directory doesn't exist, as well?

martonvago (changes requested):
Very nice 👍 👍

docs/design/python-functions.qmd
martonvago: > Do we need a `verify_exists()` function? Currently, the `verify_is_dir()` only checks whether the path is a directory or not - not whether it exists?

Doesn't it throw if the directory doesn't exist, as well?

signekb: omg, you're right ... and I have even made the test for it 😝

Issue #533, opened 2024-07-29
Put all user facing functions into `core/__init__.py`


lwjohnst86:
So that it is easier to keep track of what is internal and external files as well as making it easier to import, so we can just do `import spourt.core` (with a `as` to abbreviate the namespace call when using functions inside the script).

PR #610, opened 2024-08-27
feat: ✨ add `create_resource_structure()`

### LINKED ISSUE #568: Create and test `create_resource_structure` function
- signature: `create_resource_structure(path: Path) -> list[Path]`
- arguments:
  - `path`: path to the resources folder within a package
- returns a list of `Path`s with two elements:
  - path to the newly created resource, i.e. `<SPROUT_ROOT>/packages/<id>/resources/<new_resource_id>`
  - path to the newly created raw folder for the resource, i.e. `<SPROUT_ROOT>/packages/<id>/resources/<new_resource_id>/raw`
- use the inner functions sketched out in the [diagram](https://github.com/seedcase-project/seedcase-sprout/blob/main/docs/design/images/core-python-functions/create-resource-structure.svg) to implement the function logic
- think about which inner functions can throw what errors and whether we can handle them within `create_resource_structure` or want to let them propagate further

signekb:
## Description

Bringing together the smaller functions to make an external facing function.

Closes #568 

Relates to #533 (bc I have added a `__init__.py` in `core` with `create_resource_structure()`)

## Reviewer Focus

Focus on: 

1. How do we want to import internal functions to user-facing functions like `create_resource_structure()`? This also relates to general refactoring; right now all core functions are in the same folder which will become messy as we add new functions. Thoughts at this point? Do we want to keep as is for now and then wait for a refactoring session later on?

## Checklist

- [X] Added or updated tests
- [X] Tests passed locally


signekb:
Hmmmm..... The tests don't fail locally

martonvago (dismissed):
Looking very nice ⭐ ⭐ 

Only thing I can think of is maybe adding a test for the case when there is already a resource and so the ID has to be > 1.

Personally, I’m happy to move the functions into folders later when we have a bit more of them. But yeah, maybe group them by functions related to packages, resources, verification? And then commonly used functions of course.

I’ve never really managed to develop an amazing understanding of how Python imports should be organised (consistently, I guess 😄), so maybe Luke?

K-Beicher:
Looks good, I'm a big fan of putting things away in folders, but I'm also of the opinion that sometimes it is best to wait a little, until you can see a pattern emerging and make folders based on that. So again - let's see what @lwjohnst86 has to say :-)

tests/core/test_create_resource_structure.py
```diff
@@ -3,7 +3,7 @@
 import sprout.core as sp
 
 
-def test_returns_resource_structure(tmp_path):
+def test_returns_resource_structure_when_no_resources_exists(tmp_path):
```
martonvago, line 6:
```suggestion
def test_returns_resource_structure_when_no_resources_exist(tmp_path):
```
Sorry 😅

tests/core/test_create_resource_structure.py
```diff
@@ -12,6 +12,19 @@ def test_returns_resource_structure(tmp_path):
     assert lst_created_paths == [tmp_path / "1", tmp_path / "1" / "raw"]
 
 
+def test_returns_resource_structure_when_resources_exists(tmp_path):
```
martonvago, line 15:
```suggestion
def test_returns_resource_structure_when_resources_exist(tmp_path):
```

martonvago (dismissed):

sprout/core/create_resource_structure.py
```diff
@@ -0,0 +1,33 @@
...
+from sprout.core.create_id_path import create_id_path
+from sprout.core.create_next_id import create_next_id
+from sprout.core.create_resource_raw_path import create_resource_raw_path
+from sprout.core.get_ids import get_ids
+from sprout.core.verify_is_dir import verify_is_dir
+
+
+def create_resource_structure(path: Path) -> list[Path]:
+    """Create the directory structure of a new resource.
+
```
lwjohnst86, line 13: Only thing to add would be the documentation from the python functions doc. Eventually, all the docs there should be over with the actual functions. That way we can automatically build the website docs from these docs.

lwjohnst86 (dismissed):
Fuck yeaaasss! Only have one comment 😁

lwjohnst86:
@signekb and I am also of the similar opinion as the others. I think keeping them all in the same folder works for now and explicit importing of needed functions 👍🏻

tests/core/test_create_resource_structure.py
signekb: Haha, no reason to say sorry, @martonvago ! Thanks for the catch 👍

sprout/core/create_resource_structure.py
signekb: Do you mean this section, @lwjohnst86 ? How should we keep links in these?  

> This is the first function to use to set up the structure for a data resource. It creates the paths for a new data resource in a specific (existing) package by creating the folder setup described in the [Outputs](https://sprout.seedcase-project.org/docs/design/outputs) section. Use the path_resources() function to provide the correct path location. It creates two paths: the resources/<id>/ path and the resources/<id>/raw/ path. The output is a list of these two path objects.

lwjohnst86: Yea, exactly that one ☺️ And links... Hmm, I think a full URL will have to do for now. I think there's a way to do it internally but I'll need to read more about how that's done in the doc generator.

martonvago (approved):

lwjohnst86 (approved):

PR #611, opened 2024-08-27
build: :hammer: major restructure of deps plus upgrade to python3.12

lwjohnst86:
## Description

Some tests were failing because they used python 3.10. But the version we need is 3.12. So I updated the pyproject and poetry dependencies to use Python 3.12. If any of you are not using that, you can switch by using:

```
poetry env use python3.12
```

You will need 3.12 installed for that to work I think. I believe most of you have the latest version, but at least now it is formalized here.

I also moved the Django specific packages into their own dependency group, since we (and users) don't need to install them to just use Sprout core.

## Reviewer Focus

This PR needs a quick review. This PR is mostly to get everyone aware of the change.

## Checklist

- [x] Added or updated tests
- [x] Tests passed locally
- [x] Build passed locally

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,7 +13,7 @@ RUN pip install --no-cache-dir poetry==1.7.1
 # executed when dependency files change)
 COPY pyproject.toml poetry.lock /code/
 RUN poetry config virtualenvs.create false
-RUN poetry install --only main --no-root --no-interaction
+RUN poetry install --with django --no-root --no-interaction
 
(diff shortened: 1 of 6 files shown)


K-Beicher:
Is it meant to fail on the Django-test stage?

K-Beicher (approved):
It seems to work here.

signekb (approved):
Well, the checks pass now 🥳

PR #613, opened 2024-08-27
feat: :sparkles: add `verify_resource_properties()`

### LINKED ISSUE #574: Create and test `verify_resource_properties` function
- signature: `verify_resource_properties(properties: dict) -> dict`
- arguments:
  - `properties`: a dictionary object containing resource properties
- returns `properties` (same as in the input) if `properties` is a valid set of resource properties according to the Frictionless standard
- Raises an error if `properties` is not a valid set of resource properties according to the Frictionless standard. We should decide what error to raise, some options: `FrictionlessException` or a custom error.
- use `frictionless-py` to do the verification

martonvago:
## Description

This adds `verify_resource_properties()`.

Closes #574 

## Reviewer Focus

## Checklist

- [x] Added or updated tests
- [x] Tests passed locally
- [x] Build passed locally
- [x] Updated documentation

### DIFF
--- a/sprout/core/verify_resource_properties.py
+++ b/sprout/core/verify_resource_properties.py
@@ -13,6 +13,8 @@ class InvalidResourcePropertiesError(Exception):
             *args: non-keyword arguments
             **kwargs: keyword arguments
         """
+        # TODO: Consider if it's a problem for us that report.errors is not guaranteed
+        # to include all errors.
         errors = [
             f"{error.title}: {error.description} {error.message}"
             for error in report.errors
--- a/sprout/core/verify_resource_properties.py
+++ b/sprout/core/verify_resource_properties.py
@@ -1,0 +1,42 @@
+from frictionless import Report, Resource
+
+
+class InvalidResourcePropertiesError(Exception):
+    """Raised if the resource properties are invalid."""
+
+    def __init__(self, report: Report, properties: dict, *args, **kwargs):
+        """Initialises InvalidResourcePropertiesError.
+
+        Args:
+            report: validation report provided by Frictionless
+            properties: invalid resource properties
+            *args: non-keyword arguments
+            **kwargs: keyword arguments
+        """
+        errors = [
+            f"{error.title}: {error.description} {error.message}"
+            for error in report.errors
+        ]
+        message = (
+            f"Invalid resource properties provided:\n{properties}"
+            f"\nThe following errors were found:\n{'\n'.join(errors)}"
+        )
+        super().__init__(message, *args, **kwargs)
+
+
+def verify_resource_properties(properties: dict) -> dict:
+    """Checks if the resource properties provided are valid.
+
+    Args:
+        properties: the resource properties to check
+
+    Raises:
+        InvalidResourcePropertiesError: if Frictionless finds an error in the properties
+
+    Returns:
+        the properties, if valid
+    """
+    report = Resource.validate_descriptor(properties)
+    if not report.valid:
+        raise InvalidResourcePropertiesError(report, properties)
+    return properties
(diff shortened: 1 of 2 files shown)


sprout/core/verify_resource_properties.py
```diff
@@ -0,0 +1,42 @@
...
+            **kwargs: keyword arguments
+        """
+        errors = [
+            f"{error.title}: {error.description} {error.message}"
+            for error in report.errors
+        ]
+        message = (
+            f"Invalid resource properties provided: \n{properties}"
+            f"\nThe following errors were found:\n{'\n'.join(errors)}"
+        )
```
martonvago, line 23: This is a bit verbose, but maybe that's helpful

sprout/core/verify_resource_properties.py
```diff
@@ -0,0 +1,42 @@
...
+    def __init__(self, report: Report, properties: dict, *args, **kwargs):
+        """Initialises InvalidResourcePropertiesError.
+
+        Args:
+            report: validation report provided by Frictionless
+            properties: invalid resource properties
+            *args: non-keyword arguments
+            **kwargs: keyword arguments
+        """
+        errors = [
```
martonvago, line 16: When there are multiple errors they are not actually guaranteed all to be listed in here unfortunately. This is because some errors throw an exception internally when Frictionless does the validation and that prevents other errors from being listed.

tests/core/test_verify_resource_properties.py
```diff
@@ -0,0 +1,113 @@
...
+            ]
+        },
+    }
+
+    with raises(InvalidResourcePropertiesError) as error:
+        verify_resource_properties(properties)
+
+    message = str(error.value)
+    assert "'name' is a required property" in message
+    assert "123 is not of type 'string' at property 'format'" in message
```
martonvago, line 113: I saw this pattern on the `pytest` github and thought it was nice

sprout/core/verify_resource_properties.py
```diff
@@ -0,0 +1,42 @@
...
+            f"{error.title}: {error.description} {error.message}"
+            for error in report.errors
```
martonvago, line 18: When there are multiple errors they are not actually guaranteed all to be listed in here unfortunately. This is because some errors throw an exception internally when Frictionless does the validation and that prevents other errors from being listed.

sprout/core/verify_resource_properties.py
```diff
@@ -0,0 +1,42 @@
...
+        ]
+        message = (
+            f"Invalid resource properties provided:\n{properties}"
+            f"\nThe following errors were found:\n{'\n'.join(errors)}"
+        )
```
martonvago, line 23: This is a bit verbose, but maybe that's helpful?

K-Beicher (commented):
I'm going to leave it to others to read through and comment, but it looks convincing to me, and it works on my local machine

sprout/core/verify_resource_properties.py
lwjohnst86: Could you add a TODO note in the code about this? So we know to return to it.

sprout/core/verify_resource_properties.py
lwjohnst86: Nothing wrong with verbose and descriptive error messages! 😁

tests/core/test_verify_resource_properties.py
lwjohnst86: Really nice! 🤩

lwjohnst86 (changes requested):
Very very nice! ⭐ 🎉 Only one small comment 😁

signekb (dismissed):
Very nice 🌟

lwjohnst86 (approved):

PR #626, opened 2024-08-28
chore: :children_crossing: limit functions accessible when importing all

### LINKED ISSUE #533: Put all user facing functions into `core/__init__.py`

lwjohnst86:
## Description

These changes sets up the structure so that whenever `from sprout.core import *` is used, it only gives access to the functions we want. So far, I've only added commented code, but we can use this also as a nice visual to see our progress :tada:

Closes #533

## Reviewer Focus

This PR needs a quick review.

### DIFF
--- a/sprout/core/__init__.py
+++ b/sprout/core/__init__.py
@@ -14,7 +14,7 @@
 # Resources -----
 # from .list_resources import *
 # from .view_resource import *
-from .create_resource_structure import *
+from .create_resource_structure import create_resource_structure
 
 # from .create_resource_properties import *
 # from .extract_resource_properties import *
--- a/sprout/core/__init__.py
+++ b/sprout/core/__init__.py
@@ -1,0 +1,87 @@
+"""Core, external-facing functions of Sprout."""
+# This exposes only the functions we want exposed when
+# the package is imported via `from sprout.core import *`.
+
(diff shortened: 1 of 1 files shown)


sprout/core/__init__.py
```diff
@@ -1,3 +1,88 @@
...
+# from .create_package_properties import *
+# from .edit_package_properties import *
+# from .write_package_properties import *
+# from .delete_package import *
+
+# Resources -----
+# from .list_resources import *
+# from .view_resource import *
 from .create_resource_structure import create_resource_structure
+
```
signekb, line 18: Very minor  thing: I'm guessing we don't actually want a blank line here? 
```suggestion
```

signekb (commented):
Nice 🌟

sprout/core/__init__.py
lwjohnst86: yea, I know, it is annoying but it is ruff formatted. "It will go away" as you say 😛

signekb (approved):

martonvago (approved):
Ohhhh nice ⭐

PR #628, opened 2024-08-29
feat: ✨ add `scrape_json_from_url()`

### LINKED ISSUE #620: Create and test `scrape_url()` - a part of `create_properties_template()`
Signature: `scrape_url(url: str) -> json:`

Arguments:
- `url`: URL to the Data Package specification, found in the variable `DATA_PACKAGE_SPEC_URL`

Returns: JSON object with the contents scraped in the URL

Context: see the `create_properties_template` [diagram](https://github.com/seedcase-project/seedcase-sprout/blob/main/docs/design/images/core-python-functions/create-properties-template.svg)

signekb:
## Description

These changes are done so we can scrape the Data Package specs for the `properties` template. To be able to scrape URL content, I have added `requests` to our dependencies (To do this, I ran `poetry add requests` in the Terminal).

Closes #620 

## Reviewer Focus

This PR needs in-depth review.

## Checklist

- [X] Added or updated tests
- [X] Tests passed locally
- [X] Build passed locally


tests/core/test_scrape_url.py
```diff
@@ -0,0 +1,37 @@
...
+from requests import JSONDecodeError
+
+from sprout.core.scrape_url import scrape_url
+
+
+def test_returns_dict_with_expected_content_when_url_contains_json():
+    url = "https://datapackage.org/profiles/2.0/datapackage.json"
+    url_content = scrape_url(url)
+
+    expected_keys = [
```
signekb, line 11: Since the specs are a very large/long JSON object, I only check the keys and an except of the content. Am I missing a check of something I haven't thought of, when I do it like this?

lwjohnst86 (dismissed):
Super nice, didn't realize how easy it was to scrape a URL! 🦾 well, I guess the URL is also just plain JSON 😝

tests/core/test_scrape_url.py
```diff
@@ -0,0 +1,37 @@
+from pytest import raises
+from requests import JSONDecodeError
+
+from sprout.core.scrape_url import scrape_url
+
+
+def test_returns_dict_with_expected_content_when_url_contains_json():
```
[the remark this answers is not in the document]
lwjohnst86: I think this is fine ☺️

tests/core/test_scrape_url.py
martonvago, line 1: This is more of a general comment, but in my experience, we usually don't want to be making real network calls in tests. Instead, we would have a way of mocking the `requests` module and making it return a fake response of our choice to us, and test against that. This leads to more involved code, but it makes tests less brittle, because they won't start failing if the machine running the tests can't connect to the internet or if the server we're testing against is temporarily down. Here, writing our own fake response would also solve the issue of the real response being too long to test.

That said, in this case, because `scrape_url` is just calling 2 library functions, all we would be testing is "if I tell you to return `{"fake": "response"}`, will you return `{"fake": "response"}`?". Provided that library code works as expected, this test will always pass, so arguably there's not much point in including it.

martonvago: So just to be explicit, I would either add some custom logic to this function (e.g. more involved error handling) or just not test it at all (because we'd be testing library code, not our code).

lwjohnst86: Nice catch! Completely agree, tests should preferably not need to rely on an internet connection/have an external dependency we don't control.

signekb: That's some very good points! I was also very much in doubt as to how this function should be tested. I'm learning towards not testing it then and "trust" that the `requests` code work as expected. What do you think, @lwjohnst86 and @martonvago?

sprout/core/scrape_url.py
```diff
@@ -0,0 +1,18 @@
...
+    """Scrapes a URL with a JSON object.
+
+    Args:
+        url: URL with JSON object to scrape.
+
+    Returns:
+        A dictionary with the JSON object from the URL.
+
+    Raises:
+        JSONDecodeError: If the URL does not contain a JSON object.
```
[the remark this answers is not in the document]
signekb: I see! I have added "ConnectionError" which is what requests throws when an invalid url is given. Thoughts?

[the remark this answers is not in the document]
signekb: But again; This is just what the `requests` library code throws. I'm not sure whether we want to include those in the docstring?

tests/core/test_scrape_url.py
martonvago: To me it would make sense not to test this because there isn't really any custom logic to test?

sprout/core/scrape_url.py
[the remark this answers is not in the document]
martonvago: To be honest, I'm also not too sure. I tried looking this up once, but there didn't seem to be an obvious answer, instead advice like "include whatever is helpful to users of the function" or "include errors that can happen under normal operation". Plus, always listing every single type of error library code can throw seems like it can turn into a bit of an investigation... (Just to contradict my first comment 😅 )

tests/core/test_scrape_url.py
lwjohnst86: @martonvago agreed!

sprout/core/scrape_url.py
signekb, line 1: @lwjohnst86 should we rename this to `scrape_json_from_url()` since this is more specific to what it actually does and will be used to accomplish?

lwjohnst86 (approved):
🎉

sprout/core/scrape_url.py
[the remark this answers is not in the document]
signekb: Yeah, I can also only find descriptions like "include what's relevant" and "helpful" 🤷

PR #631, opened 2024-08-29
feat: ✨ add `write_file()`

### LINKED ISSUE #618: Create and test `write_file()`
Signature: `write_file(path: Path, string: str) -> Path:`

Arguments:
- `string`: Content of file to be created (either readme or properties file)
- `path`: path with the location of the file to be created

Returns: A path to the created file

Context: 
- see the `create_package_structure` [diagram](https://github.com/seedcase-project/seedcase-sprout/blob/main/docs/design/images/core-python-functions/create-package-structure.svg) 
- see the `create_properties_template` [diagram](https://github.com/seedcase-project/seedcase-sprout/blob/main/docs/design/images/core-python-functions/create-properties-template.svg) 

Note: 
- This is a utility function that will be used in multiple contexts. E.g., within `create_package_structure()` and `create_properties_template()`
- Only one string and path as input at a time.

signekb:
## Description

Closes #618 

## Reviewer Focus

## Checklist

- [X] Added or updated tests
- [X] Tests passed locally
- [X] Build passed locally

### DIFF
--- a/sprout/core/write_file.py
+++ b/sprout/core/write_file.py
@@ -5,6 +5,7 @@ def write_file(string: str, path: Path) -> Path:
     """Writes a file to the given path with the given content.
 
     The parent folder of the file given in path must exist.
+    If the file already exists, it will be overwritten.
 
     Args:
         string: Content to be written to the file.
--- a/sprout/core/write_file.py
+++ b/sprout/core/write_file.py
@@ -1,12 +1,12 @@
 from pathlib import Path
 
 
-def write_file(path: Path, string: str) -> Path:
+def write_file(string: str, path: Path) -> Path:
     """Writes a file to the given path with the given content.
 
     Args:
-        path: Path of the file to be created, including file name and extension.
         string: Content to be written to the file.
+        path: Path of the file to be created, including file name and extension.
 
     Returns:
         A path to the file that was created.
--- a/sprout/core/write_file.py
+++ b/sprout/core/write_file.py
@@ -4,12 +4,17 @@ from pathlib import Path
 def write_file(string: str, path: Path) -> Path:
     """Writes a file to the given path with the given content.
 
+    The parent folder of the file given in path must exist.
+
     Args:
         string: Content to be written to the file.
         path: Path of the file to be created, including file name and extension.
 
     Returns:
         A path to the file that was created.
+
+    Raises:
+        FileNotFoundError: If the parent folder of the file doesn't exist.
     """
     path.write_text(string)
 
--- a/sprout/core/write_file.py
+++ b/sprout/core/write_file.py
@@ -1,0 +1,16 @@
+from pathlib import Path
+
+
+def write_file(path: Path, string: str) -> Path:
+    """Writes a file to the given path with the given content.
+
+    Args:
+        path: Path of the file to be created, including file name and extension.
+        string: Content to be written to the file.
+
+    Returns:
+        A path to the file that was created.
+    """
+    path.write_text(string)
+
+    return path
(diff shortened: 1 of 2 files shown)


sprout/core/write_file.py
martonvago, line 1: And then would we dump the `dict`s into the argument like `write_file(path, json.dumps(properties))`?

tests/core/test_write_file.py
```diff
@@ -0,0 +1,24 @@
...
+
+    assert write_file(file_path, file_content) == file_path
+
+
+def test_creates_a_file_with_the_given_content(tmp_path, file_content):
+    """Tests that a file with the given content is created."""
+    file_path = tmp_path / "test.md"
+
+    assert write_file(file_path, file_content) == file_path
+    assert file_path.is_file() is True
```
martonvago, line 23:
```suggestion
    assert file_path.is_file()
```

sprout/core/write_file.py
```diff
@@ -0,0 +1,16 @@
+from pathlib import Path
+
+
+def write_file(path: Path, string: str) -> Path:
+    """Writes a file to the given path with the given content.
```
martonvago, line 5: Maybe include here that the directory structure specified by the path has to exist (i.e. that this will fail if the parent dir doesn't exist)? Or otherwise add a test for that failure

martonvago (changes requested):
Niiiiice 😁

sprout/core/write_file.py
signekb, line 1: @lwjohnst86 would it be incorrect to call this function `create_file()` to avoid having one function with this action?

signekb: That would be my guess, yes!

lwjohnst86: Yup, that makes sense 🦾

sprout/core/write_file.py
lwjohnst86: Both! 😁

lwjohnst86, line 5:
```suggestion
def write_file(string: str, path: Path) -> Path:
    """Writes a file to the given path with the given content.
```

When writing, the path should always be after the string, so as to pipe the output into the function's first position (as is done in R's `|>` and in bash's `|`). I know Python doesn't really have a pipe (😢) but it's good to think in that framing.

sprout/core/write_file.py
```diff
@@ -0,0 +1,16 @@
...
+
+    Args:
+        path: Path of the file to be created, including file name and extension.
+        string: Content to be written to the file.
```
lwjohnst86, line 9:
```suggestion
        string: Content to be written to the file.
        path: Path of the file to be created, including file name and extension.
```

Just to match the order.

tests/core/test_write_file.py
```diff
@@ -0,0 +1,24 @@
...
+@fixture
+def file_content():
+    return "This is test content."
+
+
+def test_returns_path(tmp_path, file_content):
+    """Tests that the path is returned."""
+    file_path = tmp_path / "test.txt"
+
+    assert write_file(file_path, file_content) == file_path
```
lwjohnst86, line 15:
```suggestion
    assert write_file(file_content, file_path) == file_path
```

tests/core/test_write_file.py
```diff
@@ -0,0 +1,24 @@
...
+
+
+def test_creates_a_file_with_the_given_content(tmp_path, file_content):
+    """Tests that a file with the given content is created."""
+    file_path = tmp_path / "test.md"
+
+    assert write_file(file_path, file_content) == file_path
```
lwjohnst86, line 22:
```suggestion
    assert write_file(file_content, file_path) == file_path
```

lwjohnst86 (changes requested):

sprout/core/write_file.py
signekb: Added 👍

lwjohnst86 (approved):
🎉

PR #633, opened 2024-08-30
feat: :sparkles: add `create_resource_properties()`

### LINKED ISSUE #577: Create and test `create_resource_properties()`
- signature: `create_resource_properties(path: Path, data_path: Path | None, properties: dict | None) -> dict`
- arguments:
  - `path`: path pointing to a specific resource folder
  - `data_path`: path pointing to a data file (optional)
  - `properties`: a dictionary object containing resource properties (optional)
  - out of `data_path` and `properties` strictly one must be given
- returns a dictionary object of resource properties following the Data Package specification
- use the inner functions sketched out in the function [diagram](https://github.com/seedcase-project/seedcase-sprout/blob/main/docs/design/images/core-python-functions/create-resource-properties.svg) to implement the function logic
- when assembling the inner functions, have a look at whether we convert `properties` from a `dict` to a `frictionless.Resource` object or vice versa multiple times, and, if so, if there’s a way to avoid this
- think about which inner functions can throw what errors and whether we can handle them within `create_resource_properties` or want to let them propagate further

martonvago:
## Description

This adds `create_resource_properties()`.

Closes #577 

## Reviewer Focus

## Checklist

- [x] Added or updated tests
- [x] Tests passed locally
- [x] Build passed locally
- [x] Updated documentation

### DIFF
--- a/sprout/core/__init__.py
+++ b/sprout/core/__init__.py
@@ -14,9 +14,9 @@
 # Resources -----
 # from .list_resources import *
 # from .view_resource import *
+from .create_resource_properties import create_resource_properties
 from .create_resource_structure import create_resource_structure
 
-# from .create_resource_properties import *
 # from .extract_resource_properties import *
 # from .edit_resource_properties import *
 # from .write_resource_data_to_raw import *
--- a/sprout/core/create_resource_properties.py
+++ b/sprout/core/create_resource_properties.py
@@ -13,17 +13,15 @@ def create_resource_properties(path: Path, properties: dict) -> dict:
 
     This function sets up and structures a new resource property by taking
     the fields given in the `properties` argument to fill them and prepare
-    them to be added to the `datapackage.json` file. It must be given as a
-    JSON object following the Data Package specification (use
-    `view_resource_properties_template()` to get a JSON object that follows
-    the Frictionless Data Package standard). The `path` argument provides
-    the path to the resource `id` that the properties are for; use
-    `path_resource()` to provide the correct path or use the output of
-    `create_resource_structure()`.
+    them to be added to the `datapackage.json` file.
 
     Args:
-        path: the path to the resource folder
-        properties: the properties of the resource
+        path: the path to the resource `id` folder; use `path_resource()`
+        to provide the correct path or use the output of `create_resource_structure()`
+        properties: the properties of the resource; must be given as a
+        JSON object following the Data Package specification; use
+        `view_resource_properties_template()` to get a JSON object
+        that follows the Frictionless Data Package standard
 
     Raises:
         NotADirectoryError: if path does not point to a directory
--- a/sprout/core/create_resource_properties.py
+++ b/sprout/core/create_resource_properties.py
@@ -17,11 +17,11 @@ def create_resource_properties(path: Path, properties: dict) -> dict:
 
     Args:
         path: the path to the resource `id` folder; use `path_resource()`
-        to provide the correct path or use the output of `create_resource_structure()`
+        	to provide the correct path or use the output of `create_resource_structure()`
         properties: the properties of the resource; must be given as a
-        JSON object following the Data Package specification; use
-        `view_resource_properties_template()` to get a JSON object
-        that follows the Frictionless Data Package standard
+        	JSON object following the Data Package specification; use
+        	`view_resource_properties_template()` to get a JSON object
+        	that follows the Frictionless Data Package standard
 
     Raises:
         NotADirectoryError: if path does not point to a directory
--- a/sprout/core/create_resource_properties.py
+++ b/sprout/core/create_resource_properties.py
@@ -11,6 +11,16 @@ from sprout.core.verify_resource_properties import verify_resource_properties
 def create_resource_properties(path: Path, properties: dict) -> dict:
     """Creates a valid properties object for the specified resource.
 
+    This function sets up and structures a new resource property by taking
+    the fields given in the `properties` argument to fill them and prepare
+    them to be added to the `datapackage.json` file. It must be given as a
+    JSON object following the Data Package specification (use
+    `view_resource_properties_template()` to get a JSON object that follows
+    the Frictionless Data Package standard). The `path` argument provides
+    the path to the resource `id` that the properties are for; use
+    `path_resource()` to provide the correct path or use the output of
+    `create_resource_structure()`.
+
     Args:
         path: the path to the resource folder
         properties: the properties of the resource
--- a/sprout/core/create_resource_properties.py
+++ b/sprout/core/create_resource_properties.py
@@ -1,0 +1,29 @@
+from pathlib import Path
+
+from sprout.core.create_relative_resource_data_path import (
+    create_relative_resource_data_path,
+)
+from sprout.core.edit_resource_properties_field import edit_resource_properties_field
+from sprout.core.verify_is_dir import verify_is_dir
+from sprout.core.verify_resource_properties import verify_resource_properties
+
+
+def create_resource_properties(path: Path, properties: dict) -> dict:
+    """Creates a valid properties object for the specified resource.
+
+    Args:
+        path: the path to the resource folder
+        properties: the properties of the resource
+
+    Raises:
+        NotADirectoryError: if path does not point to a directory
+        InvalidResourcePropertiesError: if properties are not valid Frictionless
+        resource properties
+
+    Returns:
+        the properties object, verified and updated
+    """
+    verify_is_dir(path)
+    verify_resource_properties(properties)
+    data_path = create_relative_resource_data_path(path)
+    return edit_resource_properties_field(properties, "path", str(data_path))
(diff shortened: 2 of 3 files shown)


sprout/core/create_resource_properties.py
```diff
@@ -0,0 +1,29 @@
...
+    """Creates a valid properties object for the specified resource.
+
+    Args:
+        path: the path to the resource folder
+        properties: the properties of the resource
+
+    Raises:
+        NotADirectoryError: if path does not point to a directory
+        InvalidResourcePropertiesError: if properties are not valid Frictionless
+        resource properties
```
martonvago, line 21: Alternatively, we could remap all errors coming from here into a `FailedToCreatePropertiesError`, but we may actually want to keep them separate to be able to handle them separately later.

sprout/core/create_resource_properties.py
```diff
@@ -0,0 +1,29 @@
...
+
+    Returns:
+        the properties object, verified and updated
+    """
+    verify_is_dir(path)
+    verify_resource_properties(properties)
+    data_path = create_relative_resource_data_path(path)
+    return edit_resource_properties_field(properties, "path", str(data_path))
```
martonvago, line 29: Converting `data_path` to string because a `Path` can't be serialised to json automatically. If we keep doing similar conversions when using `edit_resource_properties_field`, we could consider moving the conversion in there (but we would need to ask what types we want to allow for values in `properties`).

sprout/core/create_resource_properties.py
```diff
@@ -0,0 +1,39 @@
...
+
+    This function sets up and structures a new resource property by taking
+    the fields given in the `properties` argument to fill them and prepare
+    them to be added to the `datapackage.json` file. It must be given as a
+    JSON object following the Data Package specification (use
+    `view_resource_properties_template()` to get a JSON object that follows
+    the Frictionless Data Package standard). The `path` argument provides
+    the path to the resource `id` that the properties are for; use
+    `path_resource()` to provide the correct path or use the output of
+    `create_resource_structure()`.
```
martonvago, line 22: Was I supposed to copy this in from the function description?

lwjohnst86: You got it!!! 🌠

sprout/core/create_resource_properties.py
```diff
@@ -0,0 +1,39 @@
...
+from sprout.core.edit_resource_properties_field import edit_resource_properties_field
+from sprout.core.verify_is_dir import verify_is_dir
+from sprout.core.verify_resource_properties import verify_resource_properties
+
+
+def create_resource_properties(path: Path, properties: dict) -> dict:
+    """Creates a valid properties object for the specified resource.
+
+    This function sets up and structures a new resource property by taking
+    the fields given in the `properties` argument to fill them and prepare
```
lwjohnst86, line 15: The description for the args here should be moved to the arg description section.

sprout/core/create_resource_properties.py
lwjohnst86: hmmm, interesting. Let's see how other conversions happen and modify as needed.

sprout/core/create_resource_properties.py
lwjohnst86, line 1: Now that #626 has been merged, can you uncomment this function from that list?

lwjohnst86 (approved):
Super nice! Just some small comments

lwjohnst86 (changes requested):
Ops, clicked approve by mistake.

sprout/core/__init__.py
```diff
@@ -14,9 +14,9 @@
 # Resources -----
 # from .list_resources import *
 # from .view_resource import *
+from .create_resource_properties import create_resource_properties
```
martonvago, line 17: Like this? With just `import *` the formatter was unhappy

lwjohnst86: You got it! Yea, the formatter doesn't like those `*`...

sprout/core/create_resource_properties.py
```diff
@@ -0,0 +1,37 @@
...
+    them to be added to the `datapackage.json` file.
+
+    Args:
+        path: the path to the resource `id` folder; use `path_resource()`
+        to provide the correct path or use the output of `create_resource_structure()`
+        properties: the properties of the resource; must be given as a
+        JSON object following the Data Package specification; use
+        `view_resource_properties_template()` to get a JSON object
+        that follows the Frictionless Data Package standard
```
lwjohnst86, line 24:
```suggestion
        path: the path to the resource `id` folder; use `path_resource()`
        	to provide the correct path or use the output of `create_resource_structure()`
        properties: the properties of the resource; must be given as a
        	JSON object following the Data Package specification; use
        	`view_resource_properties_template()` to get a JSON object
        	that follows the Frictionless Data Package standard
```

As per the Google style guide.

sprout/core/create_resource_properties.py
lwjohnst86: I think keeping them separate for now makes sense.

lwjohnst86 (changes requested):

sprout/core/create_resource_properties.py
```diff
@@ -0,0 +1,38 @@
...
+    """Creates a valid properties object for the specified resource.
+
+    This function sets up and structures a new resource property by taking
+    the fields given in the `properties` argument to fill them and prepare
+    them to be added to the `datapackage.json` file.
+
+    Args:
+        path: the path to the resource `id` folder; use `path_resource()`
+            to provide the correct path or use the output of
+            `create_resource_structure()`
```
signekb, line 21: Just looked it up in Google's style guide and it seems that we should remember `.` in docstrings - see e.g., the examples under [Functions and Methods](https://google.github.io/styleguide/pyguide.html#383-functions-and-methods) in their docs.
```suggestion
            `create_resource_structure()`.
```

sprout/core/create_resource_properties.py
```diff
@@ -0,0 +1,38 @@
...
+        properties: the properties of the resource; must be given as a
+            JSON object following the Data Package specification; use
+            `view_resource_properties_template()` to get a JSON object
+            that follows the Frictionless Data Package standard
```
signekb, line 25:
```suggestion
            that follows the Frictionless Data Package standard.
```

sprout/core/create_resource_properties.py
```diff
@@ -0,0 +1,38 @@
...
+
+    Raises:
+        NotADirectoryError: if path does not point to a directory
```
signekb, line 28:
```suggestion
        NotADirectoryError: if path does not point to a directory.
```

sprout/core/create_resource_properties.py
```diff
@@ -0,0 +1,38 @@
...
+        InvalidResourcePropertiesError: if properties are not valid Frictionless
+        resource properties
```
signekb, line 30:
```suggestion
        resource properties.
```

sprout/core/create_resource_properties.py
```diff
@@ -0,0 +1,38 @@
...
+
+    Returns:
+        the properties object, verified and updated
```
signekb, line 33: It also seems from [these examples](https://google.github.io/styleguide/pyguide.html#383-functions-and-methods) that the Raises section should be after Returns.

sprout/core/create_resource_properties.py
signekb, line 1: Ah, so this function doesn't write the resource properties to `datapackage.json`? Are we missing a `write` function for this in the Core functions, @lwjohnst86 ?

signekb (changes requested):
Nice 🎈 Only docstring things and a question for Luke.

sprout/core/create_resource_properties.py
martonvago: Hmmm, this was generated by the `njpwerner.autodocstring` extension, but that could be wrong

sprout/core/create_resource_properties.py
lwjohnst86: We aren't missing it, it's called [`write_resource_properties()`](https://sprout.seedcase-project.org/docs/design/python-functions#write_resource_propertiesproperties-path) ☺️

sprout/core/create_resource_properties.py
```diff
@@ -0,0 +1,38 @@
...
+            `create_resource_structure()`.
+        properties: the properties of the resource; must be given as a
+            JSON object following the Data Package specification; use
+            `view_resource_properties_template()` to get a JSON object
+            that follows the Frictionless Data Package standard.
+
+    Raises:
+        NotADirectoryError: if path does not point to a directory.
+        InvalidResourcePropertiesError: if properties are not valid Frictionless
+        resource properties.
```
signekb, line 30: Sorry for being hyper-focused on the docstring styling 😁 
```suggestion
            resource properties.
```

signekb (changes requested):

lwjohnst86 (approved):

PR #664, opened 2024-09-03
docs: :building_construction: internal flow for `create_package_properties()`

### LINKED ISSUE #627: Create internal flow diagram for `create_package_properties()`
From the [core functions post](https://sprout.seedcase-project.org/docs/design/python-functions#create_package_propertiespath-properties): 

> This is the second function to use to create a new data package. It is
used to fill in the details in the `datapackage.json` file by using the
`properties` argument which must be a JSON object. The `properties`
argument must follow the Frictionless Data specification. To get a JSON
object that follows the correct specification use
`view_package_properties_template()`. Use `path_properties()` to provide
the correct path location. Outputs a JSON object only. Use
`write_package_properties()` to save back to the `datapackage.json`
file.

lwjohnst86:
## Description

These changes add the internal flow diagram.

Closes #627

## Reviewer Focus


docs/design/images/core-python-functions/create-package-properties.puml
```diff
@@ -0,0 +1,53 @@
...
+  vpp : - Output: JSON or Error
+  vpp : - Checks against spec
+  arg_properties --> vpp : JSON object
+
+  state "read_properties()" as rp
+  rp : - Required arg: path
+  rp : - Output: JSON
+  vif --> rp : path object
+
+  state "join_properties()" as jp
```
signekb, line 42: Coudl/should this be `edit_*` or `update_*` instead?

docs/design/images/core-python-functions/create-package-properties.puml
```diff
@@ -0,0 +1,53 @@
...
+  jp : - Required args: current and new
```
signekb, line 43:
```suggestion
  jp : - Required args: current_properties and new_properties
```

signekb (changes requested):
Quite simple and straightforward 👌 
I do have some trouble grasping what will be in the properties file already (from `create_package_structure()` that uses `create_properties_template()`) and what will be added here? Do you mind elaborating a bit on that?

martonvago (commented):
Thanks, makes sense!

My questions are pretty similar to Signe's: are the current contents of `datapackage.json` the contents of the template (with maybe some values filled in)? Or I guess it's possible that the user manually edited the file in the meantime. When combining the properties, will the new ones take precedence if there's a clash?

lwjohnst86:
@signekb I'm not completely sure what might already be in the JSON file, that's something we'll have to discuss once we have something tangible to discuss around and modify. But I could think of some things that might already be in the file, like `id`, `$schema`, and `created`. And yes @martonvago right to both first questions. We can't necessarily assume the user has directly edited the file or not. And yes, we always assume the new is intended to overwrite the old/current.

docs/design/images/core-python-functions/create-package-properties.puml
lwjohnst86: For sure! Update would work.

martonvago:
> @signekb I'm not completely sure what might already be in the JSON file, that's something we'll have to discuss once we have something tangible to discuss around and modify. But I could think of some things that might already be in the file, like `id`, `$schema`, and `created`. And yes @martonvago right to both first questions. We can't necessarily assume the user has directly edited the file or not. And yes, we always assume the new is intended to overwrite the old/current.

Okay, makes sense, thanks!

docs/design/implementation/images/core/create-package-properties.puml
```diff
@@ -0,0 +1,62 @@
...
+  vrp : - Output: JSON or Error
+  vrp : - Checks that required fields\n  are added
+  vpp --> vrp : JSON object
+
+  state "read_properties()" as rp
+  rp : - Required arg: path
+  rp : - Output: JSON
+  vif --> rp : path object
+
+  state "update_properties()" as up
```
martonvago, line 50: This is somewhat related to `create_properties_template`, but what do we want to do with unfilled properties? I.e. optional properties left empty both in `current_properties` and `new_properties`. I guess we don't want to keep them around, so drop them?

docs/design/implementation/images/core/create-package-properties.puml
```diff
@@ -0,0 +1,62 @@
...
+  up : - Required args: current_properties\n  and new_properties
```
martonvago, line 51: I just thought of this, but what if the user really edited `current_properties` and now it is not valid -- should we validate the final `properties` to make sure?

docs/design/implementation/images/core/create-package-properties.puml
```diff
@@ -0,0 +1,62 @@
...
+
+  path --> arg_path : path object
+  properties --> arg_properties : JSON object
+
+  state "verify_is_file()" as vif
+  vif : - Required arg: path
+  vif : - Output: path or Error
+  arg_path --> vif
+
+  state "verify_package_properties()" as vpp
```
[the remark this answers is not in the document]
lwjohnst86: Yea, I'm thinking we will want to require more from the user than Frictionless does.

docs/design/implementation/images/core/create-package-properties.puml
lwjohnst86: Hmm, good question. My instinct is to leave them in but have a default value of `""` or something similar. I personally like knowing what else is missing in a formal spec like this.

docs/design/implementation/images/core/create-package-properties.puml
lwjohnst86: Within `read_properties()` there should ideally be something to verify that it is a valid property file.

docs/design/implementation/images/core/create-package-properties.puml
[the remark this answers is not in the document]
martonvago: Okay, makes sense! (Then I'd be inclined to add that logic in our `verify...` functions)

docs/design/implementation/images/core/create-package-properties.puml
martonvago: Okay, if we're talking about a manageable set of properties (as we discussed today), then that works I think

docs/design/implementation/images/core/create-package-properties.puml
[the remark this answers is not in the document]
lwjohnst86: Do you mean include combine this verify as well as the required properties one below into the one function? Or what do you mean by adding that logic?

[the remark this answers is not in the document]
martonvago: I mean add an extra check within `verify_package_properties` that checks if all required properties are present. I think that would make sense because Frictionless already checks for its required properties there, so we would just be checking our own as well.

docs/design/implementation/images/core/create-package-properties.puml
signekb: @martonvago Hello, it's me again, trying to catch up ☺️  What do you mean by a "manageable set of properties"?

signekb (dismissed):
Makes sense to me 👍

docs/design/implementation/images/core/create-package-properties.puml
martonvago: Ah yeah, sorry, not obvious without context: looking at the [package schema](https://datapackage.org/profiles/2.0/datapackage.json), if we wanted to include absolutely all properties all the way down the hierarchy (so everything under package > resource > schema), that would be very many properties, most of which would go unused. So we talked about only including the ones we need from the top 2 levels (package, resource) -- which is more manageable.

signekb: Ah, thanks! That makes a lot more sense!

PR #665, opened 2024-09-03
docs: :memo: functions to help create correct properties object

lwjohnst86:
## Description

These changes add the start of some helper functions that help create the correct Data Package properties structure. Could be a dataclass or something, since it would contain data and default values, etc etc. I haven't developed them too much, but the idea came up as I was writing the usage documentation.

## Reviewer Focus

This PR needs a quick review.


martonvago (dismissed):
So would these take as argument a set of properties (individually or in some collective form) and combine/arrange them into a valid properties object?

lwjohnst86:
@martonvago yea, we could even convert the schema downloaded from `create_properties_template()` into a Python dataclass so that we can set data types and default values more easily and have it clearer for the user what they need to update and provide. The usage guide using this function would be a good example (though that still needs to be updated a bit): 

```
sp.package_properties(
    title="Diabetes and Hypertension Study", 
    description="Data from the 2021 study on diabetes and hypertension", 
    contributors=[{
        "title": "Jamie Jones",
        "email": "jamie_jones@example.com",
        "roles": ["creator"]
      }, 
      {
        "title": "Jimmy James",
        "email": "jimmy_james@example.com",
        "roles": ["contact"]
      }],
    licenses=["ODC-BY-1.0"]
)
```

If it was a dataclass, it might be:

```
sp.properties(
    title="Diabetes and Hypertension Study", 
    description="Data from the 2021 study on diabetes and hypertension", 
    contributors=[properties.contributors(
		title="Jamie Jones",
        email="jamie_jones@example.com",
        roles="creator"
      )],
    licenses=["ODC-BY-1.0"],
	resources=[properties.resource(...)]
)

**Comment by martonvago** [comment]
And we want these objects simpler than a full-fledged `frictionless.Package` / `frictionless.Resource`? I'm just thinking that Frictionless can already do (just copying from the website):
```
package = Package(
    name='package',
    title='My Package',
    description='My Package for the Guide',
    resources=[Resource(path='table.csv')],
    # it's possible to provide all the official properties like homepage, version, etc
)
```

lwjohnst86:
@martonvago ah nice, I had missed that! Ok we just use those then 😛 still will be good to wrap them up in our own function just so it gets listed as an external and usable function.

martonvago:
> @martonvago ah nice, I had missed that! Ok we just use those then 😛 still will be good to wrap them up in our own function just so it gets listed as an external and usable function.

And actually we could include the `id` on both the package and the resource there, so it would make double sense wrapping them

signekb:
@lwjohnst86 @martonvago 
Just trying to reiterate what I understand from your conversation - and please correct me if I'm wrong: 
- We'll use the `Package()` class from `frictionless` to - potentially - create a dataclass template with the default values we want. This template be created in `create_properties_template()`? 
- I'm not sure what the properties functions in this PR will do? Should they also create properties objects or does that overlap with the `create_properties_template()` function now? If they are meant to create properties objects, should we add the prefix `create_` to them?

martonvago:
This is a partial answer, but as far as I understand, the templates created by `create_properties_template` are not necessarily valid package properties: they have all the required fields but some are not filled and some have default values. They are not necessarily valid (as judged by the `verify…` functions), because many fields have format requirements that are not met by a placeholder value (e.g. the empty string is not a valid name).

And then the functions in this PR are for creating fully fledged, valid properties objects?

lwjohnst86:
@martonvago and @signekb almost. Both `create_properties_template()` and `verify_*_properties()` need a reference to compare to, so they can either create or verify against. Think of these as data that is input into the other functions. So they expose to the user what a full spec looks like from Sprout, without having to dig into the documentation in another package (frictionless or the Data Package website).

signekb:
Aha, so the functions in this PR are the "input" functions you refer to here that contain the full spec?

signekb:
@lwjohnst86 Should these functions have an action suffix? `view_`, `create_`, something else?

lwjohnst86:
@signekb no, because they are objects, not actions. So no actions apply to them.

signekb:
Ah, that makes a lot of sense, actually. Thanks!

signekb (approved):

martonvago (approved):

Issue #619, opened 2024-08-28
Create and test `create_package_structure()`


signekb:
Signature: `create_package_structure(path: Path) -> list[Path]:`

Arguments:
- `path`: path to where packages are stored

Returns: A list of the two paths that have been created: 
- one to the `datapackage.json` (`/packages/<id>/datapackage.json`) and 
- one to the `README.md` (`/packages/<id>/README.md`). 

Context: see the `create_package_structure` [diagram](https://github.com/seedcase-project/seedcase-sprout/blob/main/docs/design/images/core-python-functions/create-package-structure.svg) 

Is dependent on: 

- [x] #578 
- [x] #582 
- [x] #585
- [x] #591 
- [x] #586 
- [x] #615  
- [x] #617 
- [x] #630 
- [x] #632 
- [x] #631

PR #669, opened 2024-09-03
feat: ✨ add `create_readme_text()`

### LINKED ISSUE #617: Create and test `create_readme_text()`  - a part of `create_package_structure()`
Signature: `create_readme_text(json_object: json) -> str:`

Arguments:
- `json_object`: A json object containing the package properties template.

Returns: A string with the README text.

Context: see the `create_package_structure` [diagram](https://github.com/seedcase-project/seedcase-sprout/blob/main/docs/design/images/core-python-functions/create-package-structure.svg) 

NB: I have just named the arg `json_object` here, but maybe there's something else that would be more fitting and less generic.

K-Beicher:
## Description

Closes #617 

## Reviewer Focus

I'm not sure I've picked the correct argument to pass in, or that we want the README file to be structured this way. Also, there is only a test to make sure that it works.

## Checklist

- [x] Added or updated tests
- [x] Tests passed locally
- [x] Build passed locally
- [x] Updated documentation (docstring)


sprout/core/create_readme_text.py
```diff
@@ -0,0 +1,17 @@
...
+
+    Args:
+      json_object: A JSON object containing the package properties template
+
+    Returns:
+      A string with the README text
+    """
+    readme_text = (
+        f"This is the README file for the package.\n\n"
+        f"{json.dumps(json_object, indent=4)}"
```
lwjohnst86, line 15: This is a good start, but will need a bit more customization since we don't want a direct copy paste of the machine-readable JSON in the human-readable README. So for instance, the README title could be something like:

```
properties = json.loads(json_object)
resources = # Python code to convert the resource details in the dict as 
	# a Markdown list.
...
	f"# {properties.name}: {properties.title}\n\n"
	f"{properties.description}\n\n"
	f"Within this data package contains:\n\n"
	f"{properties.resources}
...
```

sprout/core/create_readme_text.py
```diff
@@ -0,0 +1,17 @@
+import json
+
+
+def create_readme_text(json_object: json) -> str:
```
lwjohnst86, line 4: I'm wondering if it is better to call this `properties` instead of `json_object` 🤔 Others thoughts?

lwjohnst86 (changes requested):

sprout/core/create_readme_text.py
signekb: I agree! `properties` is aligned with what we call it elsewhere

martonvago: Agreed, and it should be type `dict`. (The `json` there is the JSON library)

tests/core/test_create_readme_text.py
```diff
@@ -0,0 +1,24 @@
+import json
+
+from sprout.core.create_readme_text import create_readme_text
+
+
+def test_create_readme_text_with_valid_json():
+    # Arrange
+    json_string = '{"name": "My Package", "version": "1.0.0", "description": "Sample"}'
+    json_object = json.loads(json_string)
```
martonvago, line 9: So I think the input to this function is a regular Python dictionary object, so we could just pass `{"name": "My Package", "version": "1.0.0", "description": "Sample"}` directly.

martonvago (changes requested):

sprout/core/create_readme_text.py
K-Beicher: @lwjohnst86 Do we have an example of what the json is likely to look like, I was struggling with visualising that before, and still is, although that could be the remnants of covid as well. I agree that properties is a good idea, and that it could be a dict type.

lwjohnst86: @K-Beicher yea, since we don't have that object yet, it's hard for all of us to visualize it. I think for now we can keep it pretty simple and not assume too much, since we'll probably have to make a lot of changes, especially to this function, along the way.

lwjohnst86, line 4:
```suggestion
def create_readme_text(properties: dict) -> str:
```

sprout/core/create_readme_text.py
```diff
@@ -0,0 +1,17 @@
...
+    """Create a json object containing the readme text.
+
+    Args:
+      json_object: A JSON object containing the package properties template
```
lwjohnst86, line 8:
```suggestion
    """Create a string to use for the README file.

    Args:
      properties: An object containing the package and resource properties.
```

sprout/core/create_readme_text.py
```diff
@@ -0,0 +1,17 @@
...
+
+    Returns:
+      A string with the README text
```
lwjohnst86, line 11:
```suggestion
      A string for the README text.
```

sprout/core/create_readme_text.py
lwjohnst86: @K-Beicher if you add something like the above for now, we can merge this in and update it as we go along ☺️

lwjohnst86 (changes requested):

tests/core/test_create_readme_text.py
K-Beicher: Hi @martonvago How will this work with the new layout of the function?

sprout/core/create_readme_text.py
```diff
@@ -0,0 +1,22 @@
...
+
+    Args:
+      properties: An object containing the package and resource properties.
+
+    Returns:
+      A string for the README text
+    """
+    properties = json.loads(properties)
+    resources = # Python code to convert the resource details in the dict as
+    # a Markdown list.
```
K-Beicher, line 14: @lwjohnst86 it is of course now complaining about the resources being not defined here...

sprout/core/create_readme_text.py
```diff
@@ -0,0 +1,22 @@
...
+def create_readme_text(properties: dict) -> str:
+    """Create a json object containing the readme text.
+
+    Args:
+      properties: An object containing the package and resource properties.
+
+    Returns:
+      A string for the README text
+    """
+    properties = json.loads(properties)
```
K-Beicher, line 12: @lwjohnst86 in your first review you'd left the json_object in the (), I may have misunderstood, but we've since agreed that it should be renamed to properties...

tests/core/test_create_readme_text.py
lwjohnst86: @K-Beicher we haven't changed anything related to the input, we still expect a `dict` object, so that doesn't change anything.

tests/core/test_create_readme_text.py
```diff
@@ -0,0 +1,24 @@
...
+    result = create_readme_text(json_object)
+
+    expected_output = (
+        "This is the README file for the package.\n\n"
+        "{\n"
+        '    "name": "My Package",\n'
+        '    "version": "1.0.0",\n'
+        '    "description": "Sample"\n'
+        "}"
+    )
```
lwjohnst86, line 21: This will need to be updated :relaxed:

lwjohnst86 (changes requested):

K-Beicher:
@lwjohnst86 I've removed the test file so we can get the main script merged in.

sprout/core/create_readme_text.py
```diff
@@ -0,0 +1,24 @@
...
+    # TODO: Finish this once Properties have been finished.
+    # resources = # Python code to convert the resource details in the dict as
+    # a Markdown list.
+
+    readme_text = (
+        f"This is the README file for the package.\n\n"
+        f"{properties.name}: {properties.title}\n\n"
+        f"Within this data package contains:\n\n"
+        f"{properties.resources}"
```
lwjohnst86, line 22: I wonder if it will be easier to create this text in a Markdown file with templating, like done with Django and the `jinja` used to create the template Property dataclasses... :thinking:

K-Beicher: I'll have a look at that.

sprout/core/create_readme_text.py
```diff
@@ -0,0 +1,24 @@
...
+def create_readme_text(properties: dict) -> str:
+    """Create a json object containing the readme text.
+
+    Args:
+      properties: An object containing the package and resource properties.
+
+    Returns:
+      A string for the README text
+    """
+    properties = json.loads(properties)
```
martonvago, line 13:
```suggestion
```

Here we don't need to load `properties` from a string because it's already a `dict`.

sprout/core/create_readme_text.py
martonvago, line 22:
```suggestion
        f"{properties["name"]}: {properties["title"]}\n\n"
        f"Within this data package contains:\n\n"
        f"{properties["resources"]}"
```
Here we cannot access fields using a dot because `properties` is just a dictionary. If we passed `properties` as a `PackageProperties` we could use the dot notation. (Maybe we'll do that later, but I think for now everything else expects a `dict`.)

martonvago (changes requested):
Sorry to start on this again, I was just having a look at https://github.com/seedcase-project/seedcase-sprout/issues/619 and realised this needed some more changes.

sprout/core/create_readme_text.py
```diff
@@ -0,0 +1,23 @@
...
+    """
+    # TODO: Finish this once Properties have been finished.
+    # resources = # Python code to convert the resource details in the dict as
+    # a Markdown list.
+
+    readme_text = (
+        f"This is the README file for the package.\n\n"
+        f"{properties["name"]}: {properties["title"]}\n\n"
+        f"Within this data package contains:\n\n"
+        f"{properties["resources"]}"
```
lwjohnst86, line 21:
```suggestion
        f"# {properties["name"]}: {properties["title"]}\n\n"
        f"properties["description"]\n\n"
        f"There are {len(properties["resources"])} resources in this package."
```

PR #676, opened 2024-09-06
feat: ✨ add dataclasses for package and constituent structures

### LINKED ISSUE #625: Create dataclasses representing the top levels of the data package spec
Create dataclasses to represent the top levels of the [data package spec](https://datapackage.org/profiles/2.0/datapackage.json), including most* package and resource properties, but excluding `resource > schema` and everything under there. 

\* which ones we want to include is a work in progress

Dataclasses:
- PackageProperties
- ResourceProperties
- SourceProperties
- LicenseProperties
- ContributorProperties

We want to be able to instantiate the classes with no arguments, so provide sensible default values.

Ideally, the classes should be generated automatically based on the data package spec.

martonvago:
## Description

This adds dataclasses for `Package`, `Resource`, `Source`, `Contributor` and `License`. These make up the top 2 levels of the data package spec. I've included some sensible default values for some of the properties.

I expect this to change a lot, so feel free to make many comments!

Closes #625 

## Reviewer Focus

## Checklist

- [x] Added or updated tests
- [x] Tests passed locally
- [x] Build passed locally
- [x] Updated documentation


sprout/core/create_properties_template.py
martonvago, line 1: Here, I'm writing out the object by hand.
The other 2 options were:
- instantiate a `Package` using the constructor, i.e. `package = Package(...keyword properties...)`. The problem with this is that `id` is not a property on `Resource` and while it is a property on `Package` according to the DP specification, it is not a property in the `frictionless-py` implementation.
- instantiate a `Package` from a descriptor, i.e. `package = Package.from_descriptor(package_properties)`. This allows custom properties, so `id` is processed fine.

However, both options want to create a valid `Package`, so not all fields can be left blank (e.g. it wants a `name` matching a certain pattern). We can of course put in some default values there, but then our code becomes:
```
package_properties = {...}
json_object = Package.from_descriptor(package_properties).to_dict()
```
But this `json_object` is the same as `package_properties`, so we might as well just write our `package_properties` to a file and then we can be more flexible with leaving values blank or putting whatever defaults we want. This means that the template is not necessarily a valid `Package`, but, given that it is a template, I think that's actually fine?

But let me know what you think.

martonvago: And let me know if there are any defaults we want to provide. E.g. `"version": "0.1.0"` or `"created": <now>`.

sprout/core/verify_package_properties.py
martonvago, line 1: Just a set of our own required properties. Putting it in the `verify...` files because presumably we will want to check that they are present when we do the verification. This is just my first guess, so feel free to change.

tests/core/test_create_properties_template.py
```diff
@@ -0,0 +1,34 @@
...
+
+    assert package["resources"]
+    resource = package["resources"][0]
+    assert set(resource.keys()) == RESOURCE_REQUIRED_PROPERTIES
+
+
+def test_rejects_unresolvable_path(tmp_path):
+    """Given a path with a nonexistent parent, should throw FileNotFoundError."""
+    with raises(FileNotFoundError):
+        create_properties_template(tmp_path / "nonexistent" / "test.test")
```
martonvago, line 34: I'm testing for this for completeness, but this is basically just repeating a test in `test_write_file.py`.

sprout/core/create_properties_template.py
martonvago, line 1: Here, I'm writing out the object by hand.
The other 2 options were:
- instantiate a `Package` using the constructor, i.e. `package = Package(...keyword properties...)`. The problem with this is that `id` is not a property on `Resource` and while it is a property on `Package` according to the DP specification, it is not a property in the `frictionless-py` implementation.
- instantiate a `Package` from a descriptor, i.e. `package = Package.from_descriptor(package_properties)`. This allows custom properties, so `id` is processed fine.

However, both options want to create a valid `Package`, so not all fields can be left blank (e.g. it wants a `name` matching a certain pattern). We can of course put in some default values there, but then our code becomes:
```
package_properties = {...}
json_object = Package.from_descriptor(package_properties).to_dict()
```
But this `json_object` is the same as `package_properties`, so we might as well just write our `package_properties` to a file and then we can be more flexible with leaving values blank or putting whatever defaults we want. This means that the template is not necessarily a valid `Package`, but, given that it is a template, I think that's actually fine?

But let me know what you think.

martonvago: And let me know if there are any defaults we want to provide. E.g. `"version": "0.1.0"` or `"created": <now>`.

sprout/core/verify_package_properties.py
martonvago, line 1: Just a set of our own required properties. Putting it in the `verify...` files because presumably we will want to check that they are present when we do the verification. This is just my first guess, so feel free to change.

sprout/core/create_properties_template.py
lwjohnst86: Hmmm, the whole point of scraping the URL and storing the full spec somewhere was so we don't have this situation of having to write them all out. But maybe that's fine. I think then, these explicitly written out properties need to be moved to another file/function/class/dataclass/constant rather than be kept internal to this function here.

And yea, as a template, it doesn't need to be valid just yet. But putting in default values makes the most sense for now.

lwjohnst86 (changes requested):

martonvago:
For reference, here’s a naive implementation of extracting the same template from the schema:
```python
def create_properties_template(path: Path) -> Path:
    schema = scrape_json_from_url("https://datapackage.org/profiles/2.0/datapackage.json")

    # TODO: handle error case when these keys are missing
    package_tmpl = pick_required_props(schema, PACKAGE_REQUIRED_PROPERTIES)
    resource_tmpl = pick_required_props(schema["properties"]["resources"]["items"], RESOURCE_REQUIRED_PROPERTIES)
    resource_tmpl["id"] = ""
    package_tmpl["resources"].append(resource_tmpl)

    return write_json(package_tmpl, path)

def pick_required_props(schema: dict, required_fields: list) -> dict:
    return {key: get_default_value(value) 
            for key, value in schema["properties"].items() 
            if key in required_fields}

def get_default_value(value: dict) -> str | list | None:
    # TODO: handle case where there are multiple options for 'type'
    match value.get("type"):
        case "string":
            return ""
        case "array":
            return []
        case _:
            return None
```
In this form, the fields we want are still hardcoded, so there is no gain there. We could extract all the fields from the top 2 levels (so package and resource properties) and then drop the ones we don’t want, but that’s pretty much the same thing. If we wanted to provide defaults, we would need to specify those property by property, which would mean hardcoding those properties once again, in a way.

So, looking at this again, writing out the template by hand seems simpler and less error prone to me. Or can anyone see how the function above could be improved? (If we wanted all nested properties all the way down the hierarchy, we would do it using recursion, but there are some difficult edge cases, so it becomes inelegant pretty quickly.)

pyproject.toml
```diff
@@ -34,6 +34,7 @@ requests = "^2.32.3"
 [tool.poetry.group.test.dependencies]
 pytest = "^8.3.2"
 pytest-django = "^4.8.0"
+freezegun = "^1.5.1"
```
martonvago, line 37: For handling dates/times in tests consistently

sprout/core/create_properties_template.py
martonvago, line 1: I've pulled out the templates, leaving them in this file for now, but we can move them later. We can make a (data)class for them later if it seems useful, for here a constant seems fine.

sprout/core/create_properties_template.py
```diff
@@ -0,0 +1,58 @@
...
+    "description": "",
+    "homepage": "",
+    "sources": [],
+    "format": "csv",
+    "mediatype": "text/csv",
+    "encoding": "utf-8",
+}
+
+
+def package_template() -> dict:
```
martonvago, line 21: This needs to be a function for the timestamp to be generated at the right time.

sprout/core/create_properties_template.py
```diff
@@ -0,0 +1,58 @@
...
+        A package template object.
+    """
+    return {
+        "id": "",
+        "name": "",
+        "title": "",
+        "description": "",
+        "homepage": "",
+        "version": "0.1.0",
+        "created": datetime.now().astimezone().isoformat(timespec="seconds"),
```
martonvago, line 34: This matches the format they want.

sprout/core/package.py
martonvago, line 1: Re naming: 
- I chose to name the classes like this instead of `PackageProperties` and `ResourceProperties` to keep things short, but we may actually prefer the longer names to avoid using the same names as `frictionless` (it's not a problem, you just have to double check which one you're importing).
- I put `package` as the name of the file because it contains the classes that constitute a package. If anyone has a better idea, I'm happy to change it. Not sure if it would make sense to move these class definitions to separate files at this point.

A useful thing: you can get a dictionary of these by doing e.g. `dataclasses.asdict(package)`.

Default string representation: `Package(id='', name='', title='', description='', homepage='', version='0.1.0', created='2024-09-10T17:30:25+02:00', contributors=[], keywords=[], image='', licenses=[], resources=[Resource(id='', name='', path='', type='table', title='', description='', homepage='', sources=[], format='csv', mediatype='text/csv', encoding='utf-8')], sources=[])`
This seems okay to me, but we can override it easily.

sprout/core/package.py
```diff
@@ -0,0 +1,75 @@
...
+    name: str = ""
+    title: str = ""
+    description: str = ""
+    homepage: str = ""
+    version: str = "0.1.0"
+    created: str = field(
+        default_factory=lambda: datetime.now()
+        .astimezone()
+        .isoformat(timespec="seconds")
+    )
```
martonvago, line 69: I'm putting this as a string because I don't really see a scenario where we want to manipulate it as a date/time.

tests/core/test_package.py
martonvago, line 1: I looked up whether you are supposed to test object instantiation for dataclasses (i.e. `title = "my title"; package = Package(title=title); assert package.title == title`) and thankfully sounds like no, so I'm only testing the default case.

signekb:
@martonvago Will you rename this PR and the accompanying issue when you have an overview of which dataclasses we need? :)

sprout/core/package.py
```diff
@@ -0,0 +1,106 @@
...
+        "path",
+        "type",
+        "title",
+        "description",
+        "homepage",
+        "sources",
+        "format",
+        "mediatype",
+        "encoding",
+    }
```
martonvago, line 32: I’m putting this here so that it can be referenced when we verify that a dictionary / json object representing a package or resource has all required properties. If this set will always be the same as the set of all class attributes, then we can just pull that automatically from the class definition and this variable is not needed. However, if a dictionary / json object will be valid with only some class attributes (and this seems more likely?), then we need to list the required properties somewhere. I at least can’t see how we could pick out class attributes that are required properties for the purpose of verification when we use default values in the class definition.

tests/core/test_package.py
martonvago, line 1: I looked up whether you are supposed to test object instantiation for dataclasses (i.e. `title = "my title"; package = Package(title=title); assert package.title == title`) and thankfully sounds like no, so I'm only testing the default case.

sprout/core/package.py
lwjohnst86: For the name, it makes more sense to call it `properties`, since it contains properties of a package.

Also, really nice start here! There's some things that just... tickle my intuition about how we want this done and I can't yet identify what the tickle is yet 😆 So I'll return to this in a bit.

sprout/core/package.py
```diff
@@ -0,0 +1,106 @@
+from dataclasses import dataclass, field
+from datetime import datetime
+from typing import ClassVar
+
+
+@dataclass
+class Source:
```
lwjohnst86, line 7: Just thinking, can we nest dataclasses in a class? Like:

```
Properties.Source()
```

sprout/core/package.py
```diff
@@ -0,0 +1,106 @@
...
+    """Class representing the raw data sources for a package or resource."""
+
+    title: str = ""
+    path: str = ""
+    email: str = ""
+    version: str = ""
```
lwjohnst86, line 13: I think this is partly why I wanted the scraping done, to not only get the fields, but also the documentation. Since these dataclasses need some docstrings, and it would be nice to autogenerate the documentation from official sources rather than manually doing it (let's not waste that type of time 😛 )

sprout/core/package.py
lwjohnst86: I understand the utility of this, but maybe don't love that it is now attached to the class that is exposed to the user... Not sure of a good solution, need to think about it.

sprout/core/package.py
```diff
@@ -0,0 +1,106 @@
...
+    id: str = ""
+    name: str = ""
+    title: str = ""
+    description: str = ""
+    homepage: str = ""
+    version: str = "0.1.0"
+    created: str = field(
+        default_factory=lambda: datetime.now()
+        .astimezone()
+        .isoformat(timespec="seconds")
```
lwjohnst86, line 99: This would be better moved as its own function in a helper file or something.

sprout/core/package.py
martonvago: Hmm, that's a good point. Maybe moving these back to the files with the verification functions is a bit better then? Or any place that is not directly exposed to the user in the init file.

sprout/core/package.py
martonvago: Yeah, pretty much any way we want, I think. Are you thinking
```
Properties
   > Source
   > Contributor
   > License
   > Resource
```
or something more complicated?

sprout/core/package.py
martonvago: So call the file `properties` and the top-level class (if nesting) `Properties`?

sprout/core/package.py
martonvago: We could check if the libraries that promise to do something like convert a JSON schema to a Python class actually do what we need? There's [this one](https://github.com/koxudaxi/datamodel-code-generator) to be used as a CLI tool (so we could pre-generate the class definitions) or [this one](https://docs.pydantic.dev/latest/concepts/models/#dynamic-model-creation) that kinda allows generating classes at runtime (but this one would need more work to take a JSON schema as input).
I'm not super optimistic about the complexity these would add and the second option seems brittle and very opaque from a code clarity point of view.

martonvago:
I will close this because we've chosen the https://github.com/seedcase-project/seedcase-sprout/pull/692 road

PR #684, opened 2024-09-10
docs: 🔥 remove `create_properties_template()`

signekb:
## Description

Since we have decided to avoid scraping from the Data Package URL and do classes/dataclasses instead

## Reviewer Focus

This PR needs a quick review.

## Checklist

- [X] Build passed locally


K-Beicher (approved):
Does what it is supposed to, I think.

lwjohnst86 (approved):
🎉

PR #685, opened 2024-09-10
refactor: 🔥 remove some functions in `create_properties_template()` 

signekb:
## Description

Since we have decided to remove `create_properties_template()`, see #684, I thought we should remove the functions in it as well. This includes: `scrape_json_from_url()` and `drop_property_fields()`. 

I have left `edit_property_fields()` and `write_file()` which were also used in `create_properties_template()`, but I think we'll use elsewhere as well.

## Reviewer Focus

This PR needs a quick  review.

## Checklist

- [X] Added or updated tests
- [X] Tests passed locally


K-Beicher (approved):
Looks like it is removing what it is supposed to

lwjohnst86 (approved):

PR #686, opened 2024-09-10
docs: :building_construction: replace with `create_properties_object()`

signekb:
## Description

Replacing `create_package_properties()` used internally in `create_package_structure()`.

## Reviewer Focus

This PR needs a quick review.

## Checklist

- [X] Updated documentation


martonvago (commented):
Makes sense, just some questions/points that we don't necessarily have to answer for this PR:
- What is `create_properties_object` doing with the `path` argument?
- We'll have to think about whether we want to pass our properties around as dataclasses or as dictionaries (= json objects). My feeling is to go for dataclasses because then we benefit from their reliable structure and can put any package or resource-specific logic in those classes. But there may be arguments against this?
- I think it makes sense to include this step in the diagram as a conceptual step, but in terms of the actual code, this might just turn out to be a class instantiation (i.e. `package = Package(id=id)` or `package = asdict(Package(id=id))` if we want a dictionary)?

lwjohnst86 (changes requested):
@martonvago good observations!

For the second point, I agree, I think data classes would be right way to go and I don't see any strong arguments against it.

For your third point, I think that isn't necessary given the text of `create_properties_object()`, but it could be made a bit clearer to say "properties dataclass is initialized internally" or something.

docs/design/implementation/images/core/create-package-structure.puml
```diff
@@ -44,10 +44,10 @@ state "create_package_structure()" as cps {
   cd : - Output: A list of paths\n  of created directories
   cip --> cd
 
-  state "create_package_properties()" as cppt
+  state "create_properties_object()" as cppt
   cppt : - Required arg: path
   cppt : - Output: JSON object
-  cppt : - Uses create_properties_template()\n  internally
+  cppt : - Uses properties dataclasses\n  internally and sets default values
   cd --> cppt
```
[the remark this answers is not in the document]
signekb: @lwjohnst86 I'm not sure what you mean here. Should I remove `create_properties_object()` from this diagram completely? And then we'll just initialise the dataclass with default values within `create_package_structure()`? Or do we want a function for that initialisation?

docs/design/implementation/images/core/create-package-structure.puml
```diff
@@ -44,10 +44,10 @@ state "create_package_structure()" as cps {
   cd : - Output: A list of paths\n  of created directories
   cip --> cd
 
-  state "create_package_properties()" as cppt
+  state "create_properties_object()" as cppt
   cppt : - Required arg: path
```
lwjohnst86, line 48:
```suggestion
```

docs/design/implementation/images/core/create-package-structure.puml
```diff
@@ -44,10 +44,10 @@ state "create_package_structure()" as cps {
...
   cppt : - Output: JSON object
-  cppt : - Uses create_properties_template()\n  internally
+  cppt : - Uses properties dataclasses\n  internally and sets default values
   cd --> cppt
```
lwjohnst86, line 52:
```suggestion

```

lwjohnst86 (changes requested):
@signekb these suggestions hopefully make it clearer what I meant ☺️

signekb:
@lwjohnst86 Amazing, yes, that definitely made it clearer. Thanks! 
What about the outputs and inputs of JSON objects in this diagram? Should that be changed to the `Properties` class?

lwjohnst86:
@signekb for now, we'll refer to outputs/inputs as JSON, but this might change once we finalize and implement the properties as dataclasses

lwjohnst86 (approved):

PR #688, opened 2024-09-11
feat: ✨ add `get_root_envvar()`

### LINKED ISSUE #681: Create and test`get_root_envvar()` - a part of `path_sprout_root()`
Signature: `create_package_structure() -> Path | None:`

Returns: 
- Path if the `SPROUT_ROOT` environment variable is set
- None if `SPROUT_ROOT` isn't set

Context: see the `path_sprout_root()`diagram in the [path functions section](https://sprout.seedcase-project.org/docs/design/implementation/python-functions#path-functions) in the docs.

signekb:
## Description

Closes #681 

## Reviewer Focus

This PR needs in-depth review.

Focus on: 
- I'm not sure whether the first test will overwrite and remove the `SPROUT_ROOT` if it was set outside of the test? 
  If that's the case, we might need something like this to preserve the original root envvar: 

```
def test_returns_root_envvar_if_set(tmp_path):
    """Returns the root environment variable SPROUT_ROOT if it is set."""
    # Given
    original_root = os.environ.get("SPROUT_ROOT", None)
    root = tmp_path / "sprout"
    root.mkdir()
    os.environ["SPROUT_ROOT"] = str(root)

    try:
        # When
        root_path = get_root_envvar()

        # Then
        assert root_path == root
    finally:
        # Clean up
        if original_root is not None:
            os.environ["SPROUT_ROOT"] = original_root
        else:
            os.environ.pop("SPROUT_ROOT")

```

## Checklist

- [X] Added or updated tests
- [X] Tests passed locally


tests/core/test_get_root_envvar.py
martonvago, line 1: I think this is one of those cases where we would use some sort of tool to handle setting/resetting env vars safely as opposed to setting/resetting them ourselves. Or mock setting/resetting env vars as opposed to changing them at all.

E.g. with pytest we can do:
```python
def test_returns_root_envvar_if_set(monkeypatch):
    SPROUT_ROOT = "my/root"
    monkeypatch.setenv("SPROUT_ROOT", SPROUT_ROOT)

    assert get_root_envvar() == Path(SPROUT_ROOT)
```
This should revert changes automatically after the test runs.

martonvago (changes requested):
Great!! Just a comment about testing env vars

lwjohnst86 (approved):
Niiicccceeee 🎉

PR #692, opened 2024-09-12
feat: :sparkles: auto-generate property dataclasses based on spec

### LINKED ISSUE #625: Create dataclasses representing the top levels of the data package spec

martonvago:
## Description

This now generates dataclasses based on the Data Package spec using a template that produces better docstrings and appeases the linter for the most part. Not counting the fields and anything under those, I got it down to about 10 classes. 
I included both the generated file and a `properties.py` file with minor edits (adding a resource `id`, some formatting fixes, some missing docstrings).

What remains to be done:
- Some docstrings are missing in the spec, these will need to be filled in by hand. This could include a proofreading of the existing docstrings.
- We could add the default values from the other [PR](https://github.com/seedcase-project/seedcase-sprout/pull/676).
- We need to write a reasonable class for table schema fields and add this in.
- Have a think about whether we want to add any tests around the generation logic. Seeing that we need to proofread the classes every time we generate them anyway and that we probably won't be regenerating them all that often, I think testing the generation logic properly would be kinda misplaced effort. Maybe running it on a bare-bones (= fake) JSON schema and checking that it generates what we want / runs at all could work?

~~This is to show what it would look like if we used [datamodel-code-generator](https://github.com/koxudaxi/datamodel-code-generator) to extract class definitions from the [spec](https://datapackage.org/profiles/2.0/datapackage.json). I'm sure the output can be customised a bit more and where it puts the file would need to be decided etc., but I wanted to get some feedback on this rough draft to see if this is something we should work more on.~~

~~We can also have a script post-process the output if we think it's useful, but there are some things we want to rearrange.~~

## Reviewer Focus

Focus on the output generated by the library -- would this make our life easier?

## Checklist

- [?] Added or updated tests
- [x] Tests passed locally
- [x] Build passed locally
- [x] Updated documentation


properties.py
martonvago, line 1: These are the class definitions extracted from the JSON schema. 
As you can see, there are some questionable things, e.g.:
- The file is ridiculously long because the entire data package schema was processed, including the table schema and the different types of table fields. Plus, many objects in the schema can have a range of different types (with different sets of properties), which are represented as different classes here.
- Not all class attributes are documented and some explanations could be improved a bit / given more context. The docstrings come after the class attributes -- this seems to follow an older convention and it wasn't obvious to me if there's a newer one.
- There are many more or less duplicate definitions, such as `License`, `License1`, `License2`, `License3`.

signekb: I definitely see the potential in something like this! 

But it might depend on how much/how complex post processing it would need. For example:
- how do we want to structure the classes/dataclasses - which seems a bit up in the air right now
- how all-in do we want to go on having classes for the spec. Like do we actually want classes for different constraints for different kinds of fields (`Constraints` seems to be constraints for strings, while `Constraints10` are constraints for "geopoints")? (EDIT: I guess if we don't want to keep them, we can just remove a lot of them and it might still be useful to extract the documentation we need)
- how much work will it take to get the docstrings above the class attributes. It's very weird to me that they're below 🤔

lwjohnst86: Super nice! Is it possible to only select specific fields to extract and convert? For instance, only the package properties (excluding schema and resource properties)? And for the schema and resource, we run the codegen again on only those properties? Because yea, this is too big.

lwjohnst86 (commented):
Really like this idea!

properties.py
lwjohnst86: @signekb I agree about the point with docstrings below, but it does seem that is the way to write them 🤔 I think I prefer the use of Attributes in the docstrings for these documentation, but we haven't gotten to the doc generation yet, so that isn't relevant yet.

signekb: @lwjohnst86 When you say "attributes in the docstrings", do you mean something like the last part of this from [Example Google Style Python Docstring](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html): 

image

Bc I like that more too ☺️ I also think that will make the descriptions available when you do e.g., `help(Resource)` or hover over the class in VSCode.

martonvago:
@lwjohnst86 As for processing only specific fields, I think not directly with this tool, but we could pre-process the schema in a way similar to how we wanted to do it initially. So download the schema, drop the `resources` field on the package, and generate classes based on this modified schema. We can have a Python script for this and still call it from the justfile. (And we still have the option to post-process the dataclasses if we want.)

Just two (somewhat related) questions: 
1. What happens with the `resources` field? I think yesterday we said we wanted to have it on `PackageProperties` just not populate it with resources by default. But then the definition for `PackageProperties` would look like:
```python
@dataclass
class PackageProperties:
    ...
    resources: list[ResourceProperties] = field(default_factory=list)
```
Here, we're still referencing a `ResourceProperties` class in the type hint. Would we just add this in after the script has run?

2. We can run codegen on the resource schema and the table schema separately, but wouldn't we end up with the same very large number of dataclasses, just in different files? Because if we drop the table `schema` field from the resource schema, but then run the script on the table schema as well, then we end up processing the entire resource schema. Same for the package schema and the `resources` field.

properties.py
martonvago: I also think that looks nicer! It looks like it's possible to give codegen a custom template to follow when constructing class definitions, so maybe it could even create them like this 🤞

lwjohnst86:
@martonvago it sounds like some preprocessing will be needed on the JSON before giving it to the codegen CLI. Would that make sense?

sprout/core/codegen/dataclass.jinja2
```diff
@@ -0,0 +1,39 @@
...
+{% endfor -%}
+@dataclass
+{%- if base_class %}
+class {{ class_name }}({{ base_class }}):
+{%- else %}
+class {{ class_name }}:
+{%- endif %}
+{%- if description %}
+    """
+    {{ description | wordwrap(80) | indent(4) }}
```
martonvago, line 15: `wordwrap(80)` and `68` below are just what worked for getting the docstrings below the max line length. It's possible that if we play around with them a bit more we can get these numbers a bit higher.

sprout/core/codegen/generate_dataclasses.py
```diff
@@ -0,0 +1,71 @@
...
+        "--custom-template-dir", CUSTOM_TEMPLATE_DIR,
+        "--custom-formatters", CUSTOM_FORMATTERS_MODULE,
+        "--custom-file-header", "# ruff: noqa",
+        "--enum-field-as-literal", "all",
+        "--use-one-literal-as-default",
+        "--use-standard-collections",
+        "--use-title-as-name",
+        "--use-union-operator",
+        "--snake-case-field",
+        "--special-field-name-prefix", "ignore",
```
martonvago, line 35: This marks the `$schema` properties as to-ignore in the template. I'm ignoring them because they are not necessary for our classes and if we wanted to add them they would more properly be class attributes and not instance attributes.

sprout/core/codegen/generate_dataclasses.py
```diff
@@ -0,0 +1,71 @@
...
+    """
+    subprocess.run([
+        "datamodel-codegen",
+        "--input", SCHEMA_PATH,
+        "--input-file-type", "jsonschema",
+        "--output", OUTPUT_PATH,
+        "--output-model-type", "dataclasses.dataclass",
+        "--custom-template-dir", CUSTOM_TEMPLATE_DIR,
+        "--custom-formatters", CUSTOM_FORMATTERS_MODULE,
+        "--custom-file-header", "# ruff: noqa",
```
martonvago, line 28: Adding this because we don't care very much about linting the generated file if we have to make edits to it anyway before using it.

sprout/core/codegen/generated_properties.py
martonvago, line 1: There's an argument for not committing this at all, but for review purposes it's here now.

sprout/core/codegen/generate_dataclasses.py
```diff
@@ -0,0 +1,71 @@
...
+        "--use-standard-collections",
+        "--use-title-as-name",
+        "--use-union-operator",
+        "--snake-case-field",
+        "--special-field-name-prefix", "ignore",
+        "--use-field-description",
+        "--use-schema-description",
+        "--use-double-quotes",
+        "--reuse-model",
+        "--force-optional",
```
martonvago, line 40: This helps a lot with reducing the number of classes at the expense of losing the required - optional split coming from the spec.

sprout/core/fetch_json_from_url.py
martonvago, line 1: Adding this back in. Sorry for renaming to `fetch` from `scrape`. For me `scrape` means not just downloading some data as it was meant to be downloaded (what we are doing here) but also extracting specific bits from it (e.g. scraping usernames from a forum post).

sprout/core/write_json.py
martonvago, line 1: Adding this back in.

sprout/core/properties.py
```diff
@@ -0,0 +1,274 @@
...
+    )
+
+
+@dataclass
+class ResourceProperties:
+    """Data Resource.
+
+    Attributes:
+        - name (str | None): An identifier string.
+        - id (str | None): The unique identifier of this resource.
```
martonvago, line 189: I added this

sprout/core/properties.py
```diff
@@ -222,12 +213,11 @@ class ResourceProperties:
 
     name: str | None = None
     id: str | None = None
-    path: str | list[Path] | None = None
+    path: str | None = None
```
martonvago, line 216: I disabled multipath resources by not allowing a list here

sprout/core/properties.py
```diff
@@ -222,12 +213,11 @@ class ResourceProperties:
...
     data: Any | None = None
     type: Literal["table"] | None = None
     title: str | None = None
     description: str | None = None
-    homepage: str | None = None
```
martonvago: I deleted this because it is not listed on the website, only in the JSON schema, and there it is described as a Data Package property (which it is).

sprout/core/properties.py
```diff
@@ -179,7 +173,7 @@ class TableSchemaProperties:
 
     fields_match: list | None = None
     primary_key: list[str] | str | None = None
-    unique_keys: list[UniqueKey] | None = None
+    unique_keys: list[list[str]] | None = None
```
martonvago, line 176: Simplified this type because `list[None]` is meaningless as `unique_keys` and was probably added because we're forcing optional types.

justfile
```diff
@@ -62,3 +62,7 @@ build-website:
 add-function app part name:
   touch ./{{app}}/{{part}}/{{name}}.py
   touch ./tests/{{part}}/test_{{name}}.py
+
+# Extract dataclasses from the Data Package JSON schema
+generate-dataclasses:
+  poetry run python sprout/core/codegen/generate_dataclasses.py
```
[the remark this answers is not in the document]
lwjohnst86: Hmm, good question. We can leave it here for now.

sprout/core/codegen/custom_formatters.py
lwjohnst86, line 1: I think naming the folder `autogenerate_properties/` or `generate_properties/` or somthing compared to `codegen` would be clearer on its intent.

sprout/core/codegen/generate_dataclasses.py
```diff
@@ -0,0 +1,71 @@
+import subprocess
+from pathlib import Path
+
+from sprout.core.fetch_json_from_url import fetch_json_from_url
```
lwjohnst86, line 4: Since we already have the verb `get_` that implies the same thing as `fetch_`, let's use `get_` instead.

sprout/core/codegen/generate_dataclasses.py
lwjohnst86, line 28:
```suggestion
        "--custom-file-header", "# ruff: noqa\n# NOTE: This content was autogenerated, no need to edit it.",
```

Just to have a disclaimer somewhere about where the file came from.

sprout/core/codegen/generated_properties.py
```diff
@@ -0,0 +1,280 @@
...
+    """
+
+    title: str | None = None
+    path: str | None = None
+    email: str | None = None
+    version: str | None = None
+
+
+@dataclass
+class TableDialectProperties:
```
lwjohnst86, line 72: I'm not sure how much of this we'll need, but this is a great starting point!!

sprout/core/fetch_json_from_url.py
```diff
@@ -0,0 +1,20 @@
+import requests
+
+
+def fetch_json_from_url(url: str) -> dict:
```
lwjohnst86, line 4: As mentioned above, let's use `get_` instead.

sprout/core/properties.py
```diff
@@ -0,0 +1,278 @@
+from dataclasses import dataclass, field
```
lwjohnst86, line 1:
```suggestion
# NOTE: This content is modified from the auto-generated `codegen/generated_properties.py`
# file. Update the auto-generated properties file to add more dataclasses and move
# into this file.
from dataclasses import dataclass, field
```

codegen will need to be changed to something else, as per the comment above.

sprout/core/properties.py
```diff
@@ -0,0 +1,278 @@
...
+from typing import Any, Literal
+
+
+@dataclass
+class ContributorProperties:
+    """A contributor to this descriptor.
```
lwjohnst86, line 7: We'll eventually want to update this documentation a bit more as we use it and see what we might be missing, but this is a great starting point!!!

tests/core/test_write_json.py
```diff
@@ -0,0 +1,72 @@
...
+  "inner": {
+    "prop1": 123,
+    "prop2": [
+      1,
+      2,
+      null
+    ],
+    "prop3": true
+  }
+}"""
```
lwjohnst86, line 26: Woah, is this linted?? ugly with the indentation!! Any way to get it flush with everything else?

lwjohnst86 (changes requested):
Super nice!!! Just some suggestions :tada:

tests/core/test_write_json.py
martonvago: This is because it's testing the indentation as well, so we can't indent inside of `"""..."""`. I think the other option was 
```python
expected_contents = (
    '{\n'
    '  "outer": "value",\n'
    '  "inner": {\n'
    '    "prop1": 123,\n'
    '    "prop2": [\n'
    '      1,\n'
    '      2,\n'
    '      null\n'
    '    ],\n'
    '    "prop3": true\n'
    '  }\n'
    '}'
)
```

Would this be better?

sprout/core/properties.py
martonvago: https://github.com/seedcase-project/seedcase-sprout/pull/718 improves this a tiny bit

tests/core/test_write_json.py
lwjohnst86: Yeaaayyyy :star_struck:

lwjohnst86:
@martonvago I also forgot to say to include the dataclasses in the `__init__.py` file :relaxed:

lwjohnst86 (approved):
Amazing!!

PR #697, opened 2024-09-12
docs: :memo: expand on descriptions of path functions

### LINKED ISSUE #652: Create internal flow diagram for `path_resource_raw_files()`

### LINKED ISSUE #655: Create internal flow diagram for `path_package_database()`

lwjohnst86:
## Description

Closes #652, since I've removed it in favour of having a `list_files()` function instead (**Update**, keeping the `path_resource_raw_files()`). Also closes #655, since maybe we don't use the database. Or at least, it's feeling like it won't be necessary at this point.

This is a stacked PR.

## Reviewer Focus


docs/design/implementation/python-functions.qmd
```diff
@@ -342,17 +342,49 @@ fig-alt="A Plant UML schematic of the detailed code flow within the `extract_res
     `path_sprout_root()` location or all actual `resource_id` for a
     specific package.
 
+::: {.callout-note collapse="true"}
+### `path_*()` diagram
```
lwjohnst86, line 346: I considered whether this could be a class, but ended up deciding it doesn't make sense and would actually lead to more characters needing to type (e.g. `Path().package(1).resource(1)` vs `path_resources(1, 1)`).

docs/design/implementation/python-functions.qmd
```diff
@@ -232,8 +232,8 @@ path object of the created file.
 This function takes the files provided by `raw_files` and merges them
 into a `data.parquet` file provided by `path`. Use
 `path_resource_data()` to provide the correct path location for `path`
-and `path_resources_raw_files()` for the `raw_files` argument. Outputs
-the path object of the created file.
+and `path_resources_raw()` with `list_files()` for the `raw_files`
```
signekb, line 235: I actually think it's clearer to call this `path_resources_raw_files()`, so maybe we should change it to that? ☺️

docs/design/implementation/python-functions.qmd
```diff
@@ -342,17 +342,49 @@ fig-alt="A Plant UML schematic of the detailed code flow within the `extract_res
...
 ::: {.callout-note collapse="true"}
 ### `path_resource(package_id, resource_id)`
 
 Creates the absolute path to a specific resource directory within a
 package.
+
+#### Example usage
+
+``` python
+print(path_resources(1, 2))
```
signekb, line 380:
```suggestion
print(path_resource(1, 2))
```

signekb: The name of these two functions are super close, which isn't ideal, but I guess that's just how it has to be ☺️ 
At least, I can't seem to find a better alternative

docs/design/implementation/python-functions.qmd
```diff
@@ -362,37 +394,99 @@ Creates the absolute path to a resource's raw data directory within a
...
+[PosixPath('~/.sprout/packages/6/datapackage.json')]
+```
 :::
 
 ::: {.callout-note collapse="true"}
 ### `path_package_database(package_id)`
 
 Creates the absolute path to the package's database.
+
+<!-- TODO: Not sure if we will keep this database thing -->
```
signekb, line 452:
```suggestion
<!-- TODO: Not sure if we will keep the database -->
```

signekb (commented):
Nice! Just a comment on considering adding "files" to `path_resources_raw()` for clarity and a typo.

docs/design/implementation/python-functions.qmd
lwjohnst86: @signekb yea I agree. Should be easy to notice when typing that one doesn't take the resource ID arg. But yea, not super ideal.

docs/design/implementation/python-functions.qmd
```diff
@@ -362,37 +394,99 @@ Creates the absolute path to a resource's raw data directory within a
 package.
 :::
 
-::: {.callout-note collapse="true"}
-### `path_resource_raw_files(package_id, resource_id)`
+#### Example usage
```
martonvago, line 397: I think this might not be in the callout section now

docs/design/implementation/python-functions.qmd
```diff
@@ -362,37 +394,99 @@ Creates the absolute path to a resource's raw data directory within a
...
 ### `path_resource_data(package_id, resource_id)`
 
 Creates the absolute path to a resource's Parquet data file in a
-package.
+package. Internally also has a `verify_is_file()` check.
+
+#### Example usage
+
+``` python
+print(path_resources_data(4, 2))
```
martonvago, line 418:
```suggestion
print(path_resource_data(4, 2))
```
Sorry if I'm misunderstanding these

martonvago (changes requested):
Looks great, thanks 😁

docs/design/implementation/python-functions.qmd
lwjohnst86: Just fixed those before seeing your comment :relaxed:

PR #701, opened 2024-09-16
docs: :memo: remove `create_package_properties()`, use `edit_*()` instead

lwjohnst86:
## Description

Based on a comment by @martonvago, we don't actually need the `create_package_properties()` function, since the properties file is already created. So semantically, it makes more sense to use the `edit_` function. So this PR makes that change.

## Reviewer Focus

## Checklist

- [x] Updated documentation


martonvago (approved):
Makes sense!!

docs/design/implementation/python-functions.qmd
```diff
@@ -103,36 +103,26 @@ fig-alt="A Plant UML schematic of the detailed code flow within the `create_pack
...
-
-This is the second function to use to create a new data package. It is
-used to fill in the details in the `datapackage.json` file by using the
-`properties` argument which must be a JSON object. The `properties`
-argument must follow the Frictionless Data specification. To get a JSON
-object that follows the correct specification use
-`view_package_properties_template()`. Use `path_properties()` to provide
+### `edit_package_properties(path, properties)`
+
+Because there already exists an empty `datapackage.json` file, we use this function
```
signekb, line 108: Will the `datadata.json` file be completely empty though?

lwjohnst86: Yea sorta. Like, things like `created` will be added. But that's about it for now.

PR #717, opened 2024-09-19
feat: :sparkles: add `FieldProperties` and `ConstraintsProperties`

martonvago:
## Description

This adds dataclasses for `FieldProperties` and `ConstraintsProperties`.

The direct translation of the spec into classes would map each field type (for each data type) to a separate class. These classes are similar but some fields have more properties (e.g. `categories` is available only for integer and string fields) and some properties are structured differently for different fields (e.g. different constraints are applicable to string fields and integer fields). [Here](https://datapackage.org/standard/table-schema/#field)'s a nice description of the properties.

In `FieldProperties` I included all properties (except `rdfType`), accepting that this way the onus of not constructing meaningless fields (e.g. a string field with a maximum value of the integer 12) is on the user / downstream consumer. I constructed `ConstraintsProperties` similarly.

The docstrings were copied from the spec.

## Reviewer Focus

## Checklist

- [x] Added or updated tests
- [x] Tests passed locally
- [x] Build passed locally
- [x] Updated documentation


sprout/core/properties.py
```diff
@@ -161,11 +161,113 @@ class MissingValueProperties:
...
+        JSON Schema then this field value is valid.
+    """
+
+    required: bool | None = None
+    unique: bool | None = None
+    pattern: str | None = None
+    enum: list | None = None
+    min_length: int | None = None
+    max_length: int | None = None
+    minimum: str | float | int | None = None
```
martonvago, line 219: This is an example of allowing different types of values to express a minimum, without regard for the type of the field the constraint is for.

lwjohnst86 (changes requested):
Fantastic! Only comment is to add these to the `__init__.py` file.

lwjohnst86 (approved):
:tada:

PR #721, opened 2024-09-23
docs: :building_construction: diagram for `write_resource_data_to_raw()`

### LINKED ISSUE #641: Create internal flow diagram for `write_resource_data_to_raw()`

lwjohnst86:
## Description

Closes #641

## Reviewer Focus


docs/design/implementation/images/core/write-resource-data-to-raw.puml
```diff
@@ -0,0 +1,54 @@
...
+  vid : - Required arg: path
+  vid : - Output: path or Error
+  arg_path --> vid : path object
+
+  state "verify_is_file()" as vif
+  vif : - Required arg: path
+  vif : - Output: path or Error
+  arg_data_path --> vif : path object
+
+  state "create_raw_file_path()" as crfp
```
martonvago, line 35: Does this create the absolute path or only the file name for the compressed file? If the absolute path, should it also take the path to the resource folder as an argument?

martonvago (commented):
Nice!! ⭐  Just a question

signekb (commented):
Looks good to me 👍 

A question though: Are we sure we don't want to keep the original file name in the compressed file name? In my mind, this would just make it a lot more transparent which files they were compressed from from the perspective of a user (with access to the sprout folder structure)?

docs/design/implementation/images/core/write-resource-data-to-raw.puml
```diff
@@ -0,0 +1,54 @@
...
+  crfp : - Required arg: path
+  crfp : - Output: path object
+  crfp : - Internally creates\n  `timestamp` and `uuid`
+  crfp : - Uses `data_path` to\n  get extension
+  vif --> crfp : path object
```
lwjohnst86, line 40: Thinking about this, I wonder where we should put basic validation/verification steps... At this stage, or before being merged into Parquet... :thinking:

signekb: Maybe at this stage? I'm thinking that if we write "invalid" raw files that then fails validation/verification when writing to parquet, it might become messy?

martonvago (approved):
Nice, makes sense to me!

I'm approving, but I have a very small question:
Which bit would add in `raw` to the path between the resource id folder and the name of the compressed file? Would we use `path_resource_raw`?

K-Beicher (commented):
That looks good to me

lwjohnst86:
@martonvago yes, that's right, we'd use `path_resource_raw()`

PR #730, opened 2024-09-24
feat: ✨ add `path_*()` functions (with their `verify_*()`'ers)

### LINKED ISSUE #714: Create and test `path_*()`
Signature: `path_*() -> path:`

Context: See diagram in the [Core Python functions](https://sprout.seedcase-project.org/docs/design/implementation/python-functions) post. 

Note: `*` covers all path functions in the Path functions section in the post.

signekb:
## Description
  
I decided to have a script for path functions related to packages and one related to resources. I found that it became too long - for my taste at least - with the docstrings otherwise.   
  
To add existing ID's to the raised errors, I have added two verify functions.
  
To test the path functions, I have created some helper functions to create the directory structure of packages and resources.

Closes #714  

## Reviewer Focus
  

## Checklist

- [X] Added or updated tests
- [X] Tests passed locally
- [X] Build passed locally
- [X] Updated documentation (= docstrings)


tests/core/directory_structure_setup.py
```diff
@@ -0,0 +1,47 @@
...
+    """Creates a package file structure (with empty files) for path function tests.
+
+    Args:
+        root_path: Root path to create the package structure.
+        package_id: ID of the package to create.
+
+    Returns:
+        Path of package.
+    """
+    # TODO: Use `create_package_structure()` function here when it's implemented.
```
signekb, line 16:
```suggestion
    # TODO: Use `create_package_structure()` function here when has been implemented.
```

tests/core/test_path_package_functions.py
```diff
@@ -0,0 +1,67 @@
...
+    create_test_package_structure,
+)
+
+
+# Given one package
+@fixture
+def tmp_sprout_root(monkeypatch, tmp_path):
+    """Set up test package folder structure and return temp root directory"""
+    SPROUT_ROOT = str(tmp_path)
+    monkeypatch.setenv("SPROUT_ROOT", SPROUT_ROOT)
```
martonvago, line 22:
```suggestion
    monkeypatch.setenv("SPROUT_ROOT", str(tmp_path))
```
If you convert the path to string in here then you can return it as a `Path` and then you don't have to reconvert it to `Path` in the tests. Same in other test file.

martonvago (changes requested):
Really nice!! Just some comments 😊

sprout/core/path_error_context.py
```diff
@@ -0,0 +1,49 @@
...
+from sprout.core.verify_is_dir import verify_is_dir
+from sprout.core.verify_is_file import verify_is_file
+
+
+def verify_is_dir_or_raise_error_with_id_context(path: Path, ids_path: Path):
+    """Verifies if path is a directory or raises error with existing ids.
+
+    Args:
+        path: Path to verify.
+        ids_path: Path to find existing ID's.
```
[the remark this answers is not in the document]
signekb: I think you can do either? At least e.g., Merriam-Webster uses both ID's and IDs as plurals of ID 🤷  I'll change it.

sprout/core/path_package_functions.py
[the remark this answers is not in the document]
signekb: That makes a lot of sense, actually! 😅

tests/core/test_path_package_functions.py
signekb: Omg, ofc! Thanks 😁

sprout/core/path_package_functions.py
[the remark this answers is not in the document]
signekb: I have implemented this now, in a way that made sense to me. I didn't have the time to add separate tests for when the package doesn't exist and the package exist but the file is missing :)

sprout/core/path_package_functions.py
```diff
@@ -0,0 +1,57 @@
+from pathlib import Path
+
+from sprout.core.path_sprout_root import path_sprout_root
+from sprout.core.verify_is_dir import verify_is_dir
+from sprout.core.verify_is_file import verify_is_file
```
lwjohnst86, line 5: Paths for files need to `verify_is_file()`.

sprout/core/verify_is_package_dir.py
lwjohnst86, line 1: I split the original verify function into two rather than one, and simplified them a lot. The original confused me.

lwjohnst86 (dismissed):
@martonvago and @signekb I refactored the verify function into two, because the original one was confusing to me and had too many arguments for the purpose I interpreted it had. Makes the code a bit cleaner.

martonvago (changes requested):
Looks great! Just a tiny comment.

Only other thing I can think of is that `FileNotFoundError`s are not being tested, but we can add those tests later too.

lwjohnst86 (dismissed):
@martonvago since the `verify_is_file()` already has a unit test for it, I'm not too worried that we don't test that the files exist or not for these tests.

PR #731, opened 2024-09-24
feat: :sparkles: add `create_default_package_properties()`

### LINKED ISSUE #615: Create and test `create_properties_object()` - a part of `create_package_structure()`
Signature: `create_properties_object(path: Path) -> dict`

Arguments:
- `path`: path to the package folder

Returns: A dict with an initial properties object with required fields and default values

Uses the properties dataclasses internally and sets defauæt values.

Context: see the `create_package_structure` [diagram](https://github.com/seedcase-project/seedcase-sprout/blob/main/docs/design/images/core-python-functions/create-package-structure.svg) (updated in #686 )

martonvago:
## Description

This adds a function for creating package properties as a dictionary with some default values.

Closes #615 

## Reviewer Focus

## Checklist

- [x] Added or updated tests
- [x] Tests passed locally
- [x] Build passed locally
- [x] Updated documentation

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -35,6 +35,7 @@ platformdirs = "^4.3.2"
 [tool.poetry.group.test.dependencies]
 pytest = "^8.3.2"
 pytest-django = "^4.8.0"
+time-machine = "^2.15.0"
 
 [tool.poetry.group.dev.dependencies]
 ruff = "^0.6.2"
--- a/sprout/core/create_default_package_properties.py
+++ b/sprout/core/create_default_package_properties.py
@@ -1,0 +1,24 @@
+from dataclasses import asdict
+
+from sprout.core.properties import PackageProperties
+from sprout.core.utils import get_iso_timestamp
+
+
+def create_default_package_properties(id: int) -> dict:
+    """Creates a `PackageProperties` object with default values.
+
+    Args:
+        id: The id of the package.
+
+    Returns:
+        A dictionary representation of the package properties.
+    """
+    package_properties = PackageProperties(
+        id=id,
+        name=f"package-{id}",
+        title=f"Package {id}",
+        description=f"This is a description of Package {id}.",
+        version="0.1.0",
+        created=get_iso_timestamp(),
+    )
+    return asdict(package_properties)
--- a/sprout/core/create_default_package_properties.py
+++ b/sprout/core/create_default_package_properties.py
@@ -1,23 +1,18 @@
 from dataclasses import asdict
+from uuid import uuid4
 
+from sprout.core.get_iso_timestamp import get_iso_timestamp
 from sprout.core.properties import PackageProperties
-from sprout.core.utils import get_iso_timestamp
 
 
-def create_default_package_properties(id: int) -> dict:
+def create_default_package_properties() -> dict:
     """Creates a `PackageProperties` object with default values.
-
-    Args:
-        id: The id of the package.
 
     Returns:
         A dictionary representation of the package properties.
     """
     package_properties = PackageProperties(
-        id=id,
-        name=f"package-{id}",
-        title=f"Package {id}",
-        description=f"This is a description of Package {id}.",
+        id=str(uuid4()),
         version="0.1.0",
         created=get_iso_timestamp(),
     )
--- a/sprout/core/utils.py
+++ b/sprout/core/utils.py
@@ -1,3 +1,4 @@
+from datetime import datetime
 from re import sub
 
 
@@ -53,3 +54,12 @@ def convert_to_human_readable(string: str) -> str:
 
     # Capitalize first letter in words
     return string.title()
+
+
+def get_iso_timestamp() -> str:
+    """Generates a timestamp compliant with the Data Package spec.
+
+    Returns:
+        The timestamp as a string. E.g. `2024-05-14T05:00:01+00:00`.
+    """
+    return datetime.now().astimezone().isoformat(timespec="seconds")
--- a/sprout/core/utils.py
+++ b/sprout/core/utils.py
@@ -1,4 +1,3 @@
-from datetime import datetime
 from re import sub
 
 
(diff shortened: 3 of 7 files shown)


docs/design/implementation/images/core/create-package-structure.svg
martonvago, line 1: Oh I should have tried autogenerating this 🤦

pyproject.toml
```diff
@@ -35,6 +35,7 @@ platformdirs = "^4.3.2"
 [tool.poetry.group.test.dependencies]
 pytest = "^8.3.2"
 pytest-django = "^4.8.0"
+time-machine = "^2.15.0"
```
martonvago, line 38: Using this for managing dates/times in tests because `freezegun` was not behaving very nicely with non-UTC timezones

sprout/core/create_default_package_properties.py
```diff
@@ -0,0 +1,24 @@
...
+
+from sprout.core.properties import PackageProperties
+from sprout.core.utils import get_iso_timestamp
+
+
+def create_default_package_properties(id: int) -> dict:
+    """Creates a `PackageProperties` object with default values.
+
+    Args:
+        id: The id of the package.
```
martonvago, line 11: The issue says this should have a `Path` as argument, but it only needs the `id`, right? At least I wasn't sure what to do with the path.

sprout/core/create_default_package_properties.py
```diff
@@ -0,0 +1,24 @@
...
+        A dictionary representation of the package properties.
+    """
+    package_properties = PackageProperties(
+        id=id,
+        name=f"package-{id}",
+        title=f"Package {id}",
+        description=f"This is a description of Package {id}.",
+        version="0.1.0",
+        created=get_iso_timestamp(),
+    )
```
martonvago, line 23: Happy to have more / other defaults

sprout/core/create_default_package_properties.py
```diff
@@ -0,0 +1,24 @@
...
+    return asdict(package_properties)
```
martonvago, line 24: Returning a `dict` here that can be passed directly to `write_json` later. Alternatively, we can return a `PackageProperties`, but then I'm not sure if we need this function at all.

sprout/core/properties.py
```diff
@@ -298,7 +298,7 @@ class ResourceProperties:
 
     Attributes:
         - name (str | None): An identifier string.
-        - id (str | None): The unique identifier of this resource.
+        - id (int | None): The unique identifier of this resource.
```
martonvago, line 301: Changing this because we are passing `int`s as ids right now. Not super sure which way is better, but we can always change it later.

lwjohnst86 (changes requested):
Could you split this PR into 3? One for the changes to the docs, another for the typos and the last for the feature?

martonvago:
@lwjohnst86 Sure, but what typos?

sprout/core/create_default_package_properties.py
```diff
@@ -0,0 +1,24 @@
...
+    """Creates a `PackageProperties` object with default values.
+
+    Args:
+        id: The id of the package.
+
+    Returns:
+        A dictionary representation of the package properties.
+    """
+    package_properties = PackageProperties(
+        id=id,
```
lwjohnst86, line 17: Instead of using the package ID, generate a GUID/UUID, as described in https://datapackage.org/standard/data-package/#id

sprout/core/create_default_package_properties.py
```diff
@@ -0,0 +1,24 @@
+from dataclasses import asdict
+
+from sprout.core.properties import PackageProperties
+from sprout.core.utils import get_iso_timestamp
+
+
+def create_default_package_properties(id: int) -> dict:
```
lwjohnst86, line 7:
```suggestion
def create_default_package_properties() -> dict:
```

See comment below.

sprout/core/create_default_package_properties.py
lwjohnst86: Yup! :grin:

sprout/core/create_default_package_properties.py
lwjohnst86, line 11:
```suggestion
```

lwjohnst86: I think the issue is outdated, the current diagram doesn't include any input.

sprout/core/create_default_package_properties.py
```diff
@@ -0,0 +1,24 @@
...
+        id: The id of the package.
+
+    Returns:
+        A dictionary representation of the package properties.
+    """
+    package_properties = PackageProperties(
+        id=id,
+        name=f"package-{id}",
+        title=f"Package {id}",
+        description=f"This is a description of Package {id}.",
```
lwjohnst86, line 20:
```suggestion
```

We won't add this just yet, that will be later.

sprout/core/create_default_package_properties.py
lwjohnst86: This is a discussion we'll eventually have to have, but I think the JSON/dict output is the most flexible and requires the least from us in terms of extra work (I think so anyway, but we'll see)

sprout/core/utils.py
lwjohnst86, line 1: I'm not sure I like this file, particularly the name `utils`, see:

- https://www.reddit.com/r/Python/comments/g1dxfn/stop_naming_your_python_modules_utils/
- https://breadcrumbscollector.tech/where-to-put-all-your-utils-in-python-projects/
- https://dave.cheney.net/2019/01/08/avoid-package-names-like-base-util-or-common
- https://breadcrumbscollector.tech/stop-naming-your-python-modules-utils/

For now, the content inside will just stay inside until we refactor that. But in the meantime, could you move the timestamp function into it's own file?

lwjohnst86 (changes requested):
Lovely :grin: just some comments and changes requested :relaxed:

sprout/core/create_default_package_properties.py
martonvago: Ohhh, that explains the diagram!! Thanks!

lwjohnst86 (approved):
Super neat, nice!!

PR #734, opened 2024-09-24
refactor: :recycle: change type for id to int

martonvago:
## Description

This changes the type for ids on our classes from `str` to `int` because that is how we are passing them around right now.

## Reviewer Focus

This PR needs a quick review.

## Checklist

- [x] Added or updated tests
- [x] Tests passed locally
- [x] Build passed locally
- [x] Updated documentation

### DIFF
--- a/sprout/core/properties.py
+++ b/sprout/core/properties.py
@@ -298,7 +298,7 @@ class ResourceProperties:
 
     Attributes:
         - name (str | None): An identifier string.
-        - id (str | None): The unique identifier of this resource.
+        - id (int | None): The unique identifier of this resource.
         - path (str | None): A path pointing to the data for this resource.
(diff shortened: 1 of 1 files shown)


lwjohnst86 (commented):
The ID actually won't be the `package_id` we use else where. This ID will be a UUID/GUID to keep it a unique ID for the package.

martonvago:
Oh, I thought we would use `id` for both. If not, do we even want our package id to be saved as a property? (Same question for resource id.)

lwjohnst86:
The package id doesn't need to be saved, that's more of a helper for storing the package in relation to other packages on a system. The resource ID is different, it will be part of the path to the Parquet file in the resource field :relaxed:

martonvago:
I'll make a new PR for that because it's a completely different thing 😅

PR #743, opened 2024-10-01
feat: :sparkles: add sensible default values for package properties

martonvago:
## Description

This adds default values of the matching / correct type for fields on `PackageProperties` and in the classes nested under it.
I've left `ResourceProperties` alone for now because it was less straightforward, but it should be done as well to be consistent.

## Reviewer Focus

This PR needs a quick review.

## Checklist

- [x] Added or updated tests
- [x] Tests passed locally
- [x] Build passed locally
- [x] Updated documentation

### DIFF
--- a/sprout/core/properties.py
+++ b/sprout/core/properties.py
@@ -12,27 +12,25 @@ class ContributorProperties:
     """The people or organizations who contributed to this data package.
 
     Attributes:
-        title (str | None): The name of the contributor.
-        path (str | None): A fully qualified URL pointing to a relevant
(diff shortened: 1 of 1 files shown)


K-Beicher (commented):
Looks good to me, we are sure that we don't need to be able to determine whether a value has been left empty or hasn't been looked at by the creator?

lwjohnst86 (approved):
I like this approach :+1: @K-Beicher I think that this default value only really matters for the initial package setup. Once it gets into later stages of the setup, checks for the filled in fields will start showing up to the user

PR #745, opened 2024-10-02
feat: :sparkles: add computed property `asdict`

martonvago:
## Description

This adds a computed property `asdict` to all `*Properties` classes.
So with this we can write `package_properties.asdict` instead of `asdict(package_properties)`.
This was just something that felt useful when working on the verification functions, but I won't cry if we don't include it.

## Reviewer Focus

This PR needs a quick review.

## Checklist

- [x] Added or updated tests
- [x] Tests passed locally
- [x] Build passed locally
- [x] Updated documentation

### DIFF
--- a/sprout/core/properties.py
+++ b/sprout/core/properties.py
@@ -3,12 +3,26 @@
 # properties file to add more dataclasses and move them into this file.
 
 
-from dataclasses import dataclass, field
+from abc import ABC
+from dataclasses import asdict, dataclass, field
 from typing import Any, Literal
 
 
+class Properties(ABC):
+    """An abstract base class for all *Properties classes holding common logic."""
+
+    @property
+    def asdict(self) -> dict:
+        """Produces a dictionary representation of the object.
+
+        Returns:
+            The object as a dictionary.
+        """
+        return asdict(self)
+
+
 @dataclass
-class ContributorProperties:
+class ContributorProperties(Properties):
     """The people or organizations who contributed to this data package.
 
     Attributes:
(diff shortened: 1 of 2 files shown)


sprout/core/properties.py
```diff
@@ -3,12 +3,26 @@
 # properties file to add more dataclasses and move them into this file.
 
 
-from dataclasses import dataclass, field
+from abc import ABC
+from dataclasses import asdict, dataclass, field
 from typing import Any, Literal
 
 
+class Properties(ABC):
```
martonvago, line 11: (I'm marking it as abstract by inheriting `ABC` because it makes no sense to instantiate it, but in Python this alone doesn't actually prevent instantiation.)

lwjohnst86: Ooo neat! Though not sure I'm a huge fan of this, since now it seems like there's more code to provide the functionality of one function. But let's try it out, we can always remove it :relaxed:

tests/core/test_properties.py
```diff
@@ -0,0 +1,42 @@
...
+        ReferenceProperties,
+        TableSchemaForeignKeyProperties,
+        MissingValueProperties,
+        ConstraintsProperties,
+        FieldProperties,
+        TableSchemaProperties,
+        ResourceProperties,
+        PackageProperties,
+    ],
+)
```
lwjohnst86, line 37: I guess the disadvantage of a method vs a function is that we wouldn't need to test every single dataclass that we have. But I'm willing to test this method approach out :grin:

lwjohnst86 (approved):
Let's see how this goes!

Issue #711, opened 2024-09-17
Create and test `read_properties()` - a part of `edit_package_properties()`


signekb:
Signature: `read_properties(path: Path) -> json/dict:`

Returns: 
- json/dict with properties

~Should also verify that the properties it loads are valid (see [discussion](https://github.com/seedcase-project/seedcase-sprout/pull/664#discussion_r1745034511)).~ No verifying here, just reading. Since this is an internal only function, we can keep it to only have one action: reading.

Context: See diagram in the [Core Python functions](https://sprout.seedcase-project.org/docs/design/implementation/python-functions) post.

martonvago:
@lwjohnst86 I'm just wondering about verifying the properties loaded from `datapackage.json`: do we want to use the same function/logic as for the `properties` argument in `edit_package_properties`? This would make sense, except that a newly created `datapackage.json` will presumably not have all required fields (well, it will have them set to `null`), so it will be invalid in that sense.

lwjohnst86:
Yea, maybe at this stage, it doesn't validate it, only read it. I'll update the issue description.

martonvago:
But what do we want the behaviour to be if e.g. the contents of `datapackage.json` are not JSON? Exit the entire edit operation or recover somehow (e.g. by using only the values from the `properties` object in the argument to `edit_package_properties`)?

martonvago:
We can close this if we don't want to wrap `read_json` and `verify_properties_are_well_formed` into a separate `read_properties` function.

lwjohnst86:
Agreed, we don't need this function.

PR #746, opened 2024-10-02
feat: :sparkles: add `read_json()`

martonvago:
## Description

This adds a simple `read_json()` function, which loads the contents of a JSON file into a dictionary / list.
This is basically the `read_properties` function without any properties-specific validation. This way, it's not tied to the `datapackage.json` file in any way, so I used a more general name. If we do want to add properties-specific validation, we can still build on this function.

Not sure if closes but related to #711 

## Reviewer Focus

## Checklist

- [x] Added or updated tests
- [x] Tests passed locally
- [x] Build passed locally
- [x] Updated documentation

### DIFF
--- a/sprout/core/read_json.py
+++ b/sprout/core/read_json.py
@@ -1,0 +1,17 @@
+from json import loads
+from pathlib import Path
+
+
+def read_json(path: Path) -> list | dict:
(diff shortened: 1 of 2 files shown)


lwjohnst86 (approved):
Nice!

PR #751, opened 2024-10-03
docs: :building_construction: internal diagram for `write_resource_properties()`

### LINKED ISSUE #647: Create internal flow diagram for `write_resource_properties()`

lwjohnst86:
## Description

The internal flow is slightly different from the `write_package_properties()` in that the `datapackage.json` already has to exist because we need to join the existing properties to correctly add the resource to it.

I'm not certain if we need a check or function to put the resource in the right place in the spec. But let's see how the function gets developed and used.

Closes #647

## Reviewer Focus


docs/design/implementation/images/core/write-resource-properties.puml
martonvago, line 1: Will we still have an `edit_resource_properties` function? Because this seems to do that as well.

docs/design/implementation/images/core/write-resource-properties.puml
```diff
@@ -0,0 +1,61 @@
+@startuml write-resource-properties
+!theme seedcase from https://raw.githubusercontent.com/seedcase-project/seedcase-theme/main
+
+state "Input" as input {
+  input: - Use `path_*()` functions to help\ncreate correct paths
+' Arguments -----
+  properties : - Properties objects in JSON format
```
martonvago, line 7: So not `ResourceProperties`? Just comparing this to `edit_package_properties`, that is described (in the text, at least) as taking a `PackageProperties`.

docs/design/implementation/images/core/write-resource-properties.puml
```diff
@@ -0,0 +1,61 @@
...
+
+  state "verify_is_file()" as vif
+  vif : - Required arg: path
+  vif : - Output: path or Error
+  arg_path --> vif
+
+  state "read_properties()" as rp
+  rp : - Required arg: path
+  rp : - Output: JSON
+  vif --> rp : path object
```
martonvago, line 39: This would load the entire package properties, right?

martonvago (commented):
Makes sense, just questions 😊

docs/design/implementation/images/core/write-resource-properties.puml
lwjohnst86: Yea, I was thinking about. I couldn't think of an easy way of avoiding reading and joining existing datapackage. The edit function is only to edit or add properties, but not to write back to `datapackage.json`. Alternatively, we have a `write_properties()` (save to file) and `edit_properties()` (output an updated JSON/dict) that always form a full Data Package spec.... But then I feel it gets a bit big in the output... Thoughts?

docs/design/implementation/images/core/write-resource-properties.puml
lwjohnst86: Fixed in the text!

docs/design/implementation/images/core/write-resource-properties.puml
lwjohnst86: Yes :relaxed: I couldn't think of a better way to read back to just the lines needed. Ideas?

docs/design/implementation/images/core/write-resource-properties.puml
martonvago: Ah, I see, so if we split them up and kept the same arguments we would end up repeating ourselves a lot (loading the `datapackage.json`, validating). I think it makes sense to have `write_resource_properties` like this! Maybe there's some more optimal way of going about it, but I couldn't really come up with one.

martonvago (approved):
🌟 ⭐ 🌟

PR #760, opened 2024-10-09
style: :art: move WIP header into Quarto config, rather than as own file

lwjohnst86:
## Description

There is a config setting that's fairly new that allows you to put a header in all documents. So rather than use `include` shortcodes for putting the WIP in all documents, this simplifies it so the WIP callout is automatically added to all pages.

## Reviewer Focus

This PR needs a quick review.

### DIFF
--- a/_quarto.yml
+++ b/_quarto.yml
@@ -11,9 +11,9 @@ website:
   page-navigation: true
   body-header: |
     ::: {.callout-warning appearance="default" icon="true"}
-    🚧 Sprout is still in very active development and is evolving quickly, so
-    documentation and functionality may or may not work as described or intended
-    and could have substantial changes 🚧
+    🚧 Sprout is still in active development and evolving quickly, so the
+    documentation and functionality may not work as described
+    and could undergo substantial changes 🚧
     :::
   navbar:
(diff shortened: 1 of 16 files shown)


_quarto.yml
```diff
@@ -9,6 +9,12 @@ website:
   site-url: "https://sprout.seedcase-project.org/"
   repo-url: "https://github.com/seedcase-project/seedcase-sprout"
   page-navigation: true
+  body-header: |
+    ::: {.callout-warning appearance="default" icon="true"}
+    🚧 Sprout is still in very active development and is evolving quickly, so
+    documentation and functionality may or may not work as described or intended
+    and could have substantial changes 🚧
```
signekb, line 16:
```suggestion
    🚧 Sprout is still in active development and evolving quickly, so the
    documentation and functionality may not work as described
    and could undergo substantial changes 🚧
```
Just to simplify/shorten it a bit? :)

signekb (changes requested):
Ah, very nice to add it this this instead! 🎉  I have added a suggestion to simplify the warning a bit.

signekb (approved):

PR #762, opened 2024-10-09
chore: 🎨 move callout block to own file and use include instead

lwjohnst86:
## Description

I mistakenly made this a stacked PR.

Very small change, this callout block will be used in other docs, so moving it over into a preamble file and then using `include` shortcode.

## Reviewer Focus

This PR needs a quick review.

### DIFF
--- a/docs/guide/_preamble.qmd
+++ b/docs/guide/_preamble.qmd
@@ -1,13 +1,13 @@
 ::: callout-important
-For both the Python library and the CLI, Sprout assumes you have full
-control over the folders and files of the system, or at least your
+Whether you're using the Python library or the CLI, Sprout assumes you have full
+control over the system's folders and files, or at least your
 user's home directory. This includes being given space on a server that
-mostly has access through a Terminal, where you have control over the
+is primarily accessed through a terminal, where you have control over the
 directories you can write to.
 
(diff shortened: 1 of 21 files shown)


docs/guide/_preamble.qmd
```diff
@@ -0,0 +1,13 @@
...
+control over the folders and files of the system, or at least your
+user's home directory. This includes being given space on a server that
+mostly has access through a Terminal, where you have control over the
+directories you can write to.
+
+An easy example of this is if you install Sprout on your own computer
+because you want to create some data packages for research studies you
+are running. Another example would be if your research group has storage
+space on a server that you need to use a Terminal and SSH to be able to
+access.
```
signekb, line 12:
```suggestion
Whether you're using the Python library or the CLI, Sprout assumes you have full
control over the system's folders and files, or at least your
user's home directory. This includes being given space on a server that
is primarily accessed through a terminal, where you have control over the
directories you can write to.

For example, you might install Sprout on your own computer
because you want to create data packages for your research studies.
Alternatively, your research group might have storage
space on a server that requires the use of a terminal and SSH to
access.
```
Maybe? :)

signekb (changes requested):
👍 👍

PR #744, opened 2024-10-01
feat: :sparkles: add `verify_package_properties()`

martonvago:
## Description

This adds:
- `verify_package_properties_complete()`: checks if all required fields are present and not empty.
- `verify_package_properties_well_formed()`: checks if the object has the expected structure, ignoring blank fields. This is so that blank fields don't fail frictionless format checks.
- `verify_package_properties()`: checks both of the above.

I've also generalised `InvalidResourcePropertiesError` to `InvalidPropertiesError` because each individual error says whether it's a package or a resource error anyway.

Closes #712 

## Reviewer Focus

This PR needs an in-depth review.

Focus on whether this is the error-behaviour we want.

## Checklist

- [x] Added or updated tests
- [x] Tests passed locally
- [x] Linted and formatted code
- [x] Build passed locally
- [x] Updated documentation

Issue #712, opened 2024-09-17
Create and test `verify_package_properties()` - a part of `edit_package_properties()`


signekb:
Signature: `verify_package_properties(properties: json/dict) -> json/dict:`

- Checks that all fields and values match the spec 
- Checks that required fields are present

Args: 

- `properties`: JSON object of the package following the Data Package spec

Returns: 

- JSON object or Error 

Context: See diagram in the [Core Python functions](https://sprout.seedcase-project.org/docs/design/implementation/python-functions) post.

martonvago:
@lwjohnst86 Which package properties do we want to treat as required in the end?

martonvago:
And what do we want to do with `None` values? If a required field is `None` that should obviously fail validation. But what about other criteria given in the spec? Because a `None` field will fail the type check during `Package.validate_descriptor`. 

Options: 1) ignore these errors coming from frictionless, 2) remove these fields from the `dict`/json object, making it compliant with the spec. 
I think I’d prefer 2, but I’m not sure at which point in the flow we should remove them. My initial feeling is to have the `*Properties` classes spit out a `dict` representation without them, as this would nip the problem in the bud. 
This would mean not having `null` fields written to `datapackage.json` — would we be happy with this? After all, a `datapackage.json` with `null` values is not valid according to the spec.

lwjohnst86:
Hm, if a required field has none, we give an error and inform the user to add it. The only time there are `None`/`null` values in the `datapackage.json` is right after creating the structure with `create_package_structure()`. After that is the `edit_package_properties()`, which is the point in the flow that the user adds the correct fields. For items that are not package properties (e.g. resources that don't exist yet), we could potentially pass the validator only the package fields.

martonvago:
I guess I'm still a bit unclear about optional fields (e.g. `homepage` or `image`). We have all fields set to `None` by default, including optional ones, so when a user passes a `PackageProperties` object with only the required fields set to some value to `edit_package_properties`, then we still have the question of what to do with the unset fields. They are `None` in both the `current_properties` coming from `datapackage.json` and `new_properties` coming from the user, so the result of joining these two can only be `None`. But writing that out to the `datapackage.json` wouldn't give a valid package (even if we ignore the resources). That's why I'm wondering about dropping these unset optional fields. Does my question make sense? 😅 

As for the field `resources`, yeah, I think here we should just ignore that / expect it to be empty.

lwjohnst86:
[...]

PR #767, opened 2024-10-09
fix: :bug: sort ids before returning them

martonvago:
## Description

This sorts IDs before returning them. It seems to be the case that the order in which dirs are returned is not determined by Python, which can lead to some inconsistent behaviour. This is causing some tests to fail remotely, but not locally (for me at least). 

Even more interestingly, the tests are run for PRs where neither one of the participating branches even has the commit with the test (🤔). I have no idea why and this does nothing to resolve that.

## Reviewer Focus

This PR needs a quick review.

Focus on CHANGES.

## Checklist

- [x] Added or updated tests
- [x] Tests passed locally
- [x] Build passed locally
- [x] Updated documentation

### DIFF
--- a/sprout/core/get_ids.py
+++ b/sprout/core/get_ids.py
@@ -16,7 +16,7 @@ def get_ids(path: Path) -> list[int]:
     dirs = list(path.glob("*/"))
     ids = list(map(get_number_from_dir, dirs))
     # Drop any empty items
-    ids = list(filter(None, ids))
+    ids = sorted(filter(None, ids))
 
     return ids
 


signekb (commented):
Couldn't we sort the ids in the tests instead? 
Or is this bc it's just nice to ensure that the ids are always sorted when used in e.g., error messages? :)

signekb:
About the whole problem: Could this be due to a bump in dependencies that change the sorting behaviour somehow? It seems that #712 started failing after Luke merged `main` into it on 3 October, and I can see that we  bumped `frictionless` (among others) on `main` on 1 October? 
I think that the `test` action just runs all tests on the branch, independently of whether there are any commits related to that test 🌻

martonvago:
Yeah, I thought we might as well sort the ids before passing them to the error message, but I don’t feel very strongly about it.

martonvago:
Yeahhh, it could be that somehow the dependencies affected the sorting, but I think we’re only using `pathlib` for it, so it would need to be in some complicated way. At least I couldn’t figure out how or why when I looked at it last week. But I also think that the key to it lies in pinpointing when exactly main started failing.

If you look at [this merged PR](https://github.com/seedcase-project/seedcase-sprout/pull/743), this has the dependency updates from 1 Oct and yet all tests passed, so that confuses things a bit.

But `test` shouldn’t run tests that don’t exist at all on the two branches involved in the PR, right?

signekb:
Ah, you're right! My bad. Yeah, this requires some detective work 🕵️ 

I'm not sure what you mean with the last part? Which two branches are you referring to? 
Since #730 has been merged to `main` and `main` has been merged to #744, the failing tests do exist on that branch, right?

signekb:
BUT, maybe this discussion of the origin of the issue should be moved out of this PR?

signekb (approved):

signekb:
I think that this is PR is a non-controversial solution to the problem, so I'll merge it 👍

Issue #757, opened 2024-10-08
Create and test `write_resource_properties`


martonvago:
Signature: `write_resource_properties(path: Path, properties: dict) -> Path:`

Context: See diagram in the [Core Python functions](https://sprout.seedcase-project.org/docs/design/implementation/python-functions) post. 

Things to think about:
- is the behaviour of the existing `verify_resource_properties` function exactly what we want here?
- be careful which object exactly (package properties vs resource properties on the package) we're updating when we merge the new resource properties into the old ones

martonvago:
@lwjohnst86 @signekb This should have occurred to me sooner but: a package can have multiple resources, so when we update/edit a resource, we only want that specific resource changed and not the others, right? For new resources we would still want to check that the resource doesn't already exist. 

For this, we would need something unique to identify resources by. Earlier we decided we don't want to save our `resource_id` (i.e. the folder name) onto the resources in `datapackage.json`, so we can't really use that. The spec implies that the `name` field on resources is unique (it makes up foreign key references), so we could use that. Would we be happy with this?

martonvago:
Oh, but we only care about one resource for this iteration. Well, the question still stands I guess 😅

lwjohnst86:
The path field in the resource property contains the resource id 😁 so we can parse that!

martonvago:
Oh trueeeee, I forgot, thanks!!!

PR #769, opened 2024-10-10
feat: :sparkles: add default values to `ResourceProperties`

martonvago:
## Description

This adds default values to fields on `ResourceProperties` and some related objects.
With this, we'll have added defaults to almost all of our properties objects.

Related to #757 

## Reviewer Focus

This PR needs a quick review.

Focus on if these defaults are okay.

## Checklist

- [x] Added or updated tests
- [x] Tests passed locally
- [x] Build passed locally
- [x] Updated documentation

### DIFF
--- a/sprout/core/properties.py
+++ b/sprout/core/properties.py
@@ -90,63 +90,58 @@ class TableDialectProperties(Properties):
     and YAML, and spreadsheets like Microsoft Excel.
 
     Attributes:
-        header (bool | None): Specifies if the file includes a header row,
+        header (bool): Specifies if the file includes a header row,
             always as the first row in the file.
-        header_rows (list[int] | None): Specifies the row numbers for the header.
-        header_join (str | None): Specifies how multiline-header files have to join
+        header_rows (list[int]): Specifies the row numbers for the header.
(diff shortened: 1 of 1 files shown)


sprout/core/properties.py
```diff
@@ -353,47 +346,44 @@ class ResourceProperties(Properties):
     richer set of metadata.
 
     Attributes:
-        name (str | None): A simple name or identifier to be used for this resource.
+        name (str): A simple name or identifier to be used for this resource.
             Should consist only of lowercase English alphanumeric characters plus
             characters in `.-_`.
-        path (str | None): A path pointing to the data for this resource.
-        data (Any | None): Inline data for this resource.
```
martonvago, line 360: I deleted this because only one of `path` or `data` can be given according to the spec (and also this makes sense for our use case)

martonvago:
If we feel like no sensible default value can be given for a field and we don't expect to use/support that field anyway, we can always just delete it!

signekb (approved):
Yay 🎉

lwjohnst86 (approved):
:raised_hands:

PR #789, opened 2024-10-15
docs: :memo: run Python code chunks in docs

lwjohnst86:
## Description

This PR makes it so the documentation showing how to use Python also runs it, so we can always know that the output shown on the docs is the same as what we build.

This PR needs a fairly short review.


martonvago (dismissed):
Oh cool!!! 🚀

signekb (changes requested):
This is so nice! ⭐ 

However, every time I render the website, since the functions are run, they actually create the directories (e.g., `packages/1`) as well as the files `datapackage.json` and `README.md` in my local repo. Maybe there's a way to avoid that?

PR #790, opened 2024-10-16
feat: :sparkles: add `edit_package_properties()`

### LINKED ISSUE #715: Create and test `edit_package_properties()`
Signature: `edit_package_properties(path: Path, properties: json/dict) -> json/dict:`

Context: See diagram in the [Core Python functions](https://sprout.seedcase-project.org/docs/design/implementation/python-functions) post. 

Depends on:

- [x] #711 
- [x] #713 
- [x] #712

martonvago:
## Description

This adds the `edit_package_properties()` function.

Closes #715 

## Checklist

- [x] Added or updated tests
- [x] Tests passed locally
- [x] Build passed locally
- [x] Updated documentation

### DIFF
--- a/docs/design/implementation/python-functions.qmd
+++ b/docs/design/implementation/python-functions.qmd
@@ -103,19 +103,7 @@ fig-alt="A Plant UML schematic of the detailed code flow within the `create_pack
 ::: {.callout-note collapse="true"}
 ### `edit_package_properties(path, properties)`
 
-Because there already exists an empty `datapackage.json` file, we use
-this function to fill in the properties file with details specific to
-the package. We update the properties file by using the `properties`
-argument and giving it the `PackageProperties` object. See the help
-documentation of `PackageProperties` to know what should or needs to be
-added to the `datapackage.json` file. Use `path_properties()` to provide
-the correct path location. Outputs a JSON object only. Use
-`write_package_properties()` to save back to the `datapackage.json`
-file.
-
-Whenever you need to edit the `datapackage.json` properties, this is the
-function you will use again. It only edits metadata on the package
-itself, not on the data resources contained within the package.
+See the help documentation with `help(edit_package_properties())` for more details.
 
 ![Diagram showing the internal function flow of the
 `edit_package_properties()`
--- a/sprout/core/edit_package_properties.py
+++ b/sprout/core/edit_package_properties.py
@@ -16,21 +16,21 @@ def edit_package_properties(path: Path, properties: dict) -> dict:
     """Edits the properties of an existing package.
 
     Use this any time you want to edit the package's properties and particularly
-    after using `create_package_structure()`.  Because
+    after using `create_package_structure()`. Because
     `create_package_structure()` also creates an empty `datapackage.json` file,
-    you would use `edit_package_properties()` to fill in the properties file with
+    `edit_package_properties()` is used to fill in the properties file with
     details specific for the package.
 
     When you need to edit the `datapackage.json` properties, use this function
-    to ensure the properties are correctly structured and written.  It only
-    edits metadata on the package itself, not on the data resources contained
+    to ensure the properties are correctly structured and written. It only
+    edits the properties of the package itself, not on the data resources contained
     within the package.
 
-    Any value you give in `properties` will overwrite any preexisting values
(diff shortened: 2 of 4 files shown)


docs/design/implementation/python-functions.qmd
```diff
@@ -103,19 +103,7 @@ fig-alt="A Plant UML schematic of the detailed code flow within the `create_pack
 ::: {.callout-note collapse="true"}
 ### `edit_package_properties(path, properties)`
 
-Because there already exists an empty `datapackage.json` file, we use
```
lwjohnst86, line 106: I moved this text over into the function docstring.

lwjohnst86 (dismissed):
Very nicely done!! :stars: I only moved the docs over into the docstring, otherwise, amazing!

tests/core/test_edit_package_properties.py
```diff
@@ -0,0 +1,69 @@
...
+def test_throws_error_if_properties_file_cannot_be_read(tmp_path, properties):
+    """Should throw JSONDecodeError if the properties file cannot be read as JSON."""
+    file_path = tmp_path / "datapackage.json"
+    file_path.write_text(",,, this is not, JSON")
+
+    with raises(JSONDecodeError):
+        edit_package_properties(file_path, properties)
+
+
+def test_throws_error_if_current_package_properties_malformed(tmp_path, properties):
```
signekb, line 56:
```suggestion
def test_throws_error_if_current_package_properties_are_malformed(tmp_path, properties):
```
? 🤷‍♀️

tests/core/test_edit_package_properties.py
```diff
@@ -0,0 +1,69 @@
...
+    """Should throw NotPropertiesError if the current package properties are
+    malformed."""
+    package_properties = PackageProperties(name="invalid name with spaces").asdict
+    path = write_json(package_properties, tmp_path / "datapackage.json")
+
+    with raises(NotPropertiesError):
+        edit_package_properties(path, properties)
+
+
+def test_throws_error_if_new_package_properties_incorrect(properties_path):
```
signekb, line 66:
```suggestion
def test_throws_error_if_new_package_properties_are_incorrect(properties_path):
```
? 🤷‍♀️

tests/core/test_edit_package_properties.py
```diff
@@ -0,0 +1,69 @@
...
+    """Should throw NotPropertiesError if the new package properties are incorrect."""
```
signekb, line 67: Just noticing the difference in wording here with "incorrect" vs "malformed" on e.g., L58 above. Do you mean something different with these two words or do you use them as synonyms? If the latter, maybe use the same word? :)

signekb (changes requested):
Once again: Super nice 🌟 
I have added some suggestions and comments.

sprout/core/edit_package_properties.py
```diff
@@ -0,0 +1,57 @@
...
+from sprout.core.verify_is_file import verify_is_file
+from sprout.core.verify_package_properties import (
+    verify_package_properties,
+)
+from sprout.core.verify_properties_are_well_formed import (
+    verify_properties_are_well_formed,
+)
+
+
+def edit_package_properties(path: Path, properties: dict) -> dict:
```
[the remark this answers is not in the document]
martonvago: Hmm, that takes the `package_id` as argument, right? Which we don't really know in advance.
(Also, I think that function is called `path_package_properties` now, I'll change that in the docstring.)

tests/core/test_edit_package_properties.py
martonvago: Don't mind either way, but I do have a tendency to drop the "to be" in function names.

tests/core/test_edit_package_properties.py
martonvago: I'm trying to use "malformed" as the opposite of "well-formed" (so something that violates `verify_properties_are_well_formed`) and "incorrect" as something that violates `verify_package_properties`. So I guess this makes "incorrect" a broader term. Properties that are malformed or incomplete are also incorrect, but properties that are incorrect are not necessarily both malformed and incomplete. But I make no claims of having been super consistent on this.

lwjohnst86 (approved):

PR #816, opened 2024-10-25
docs: :memo: pseudo code and docstring for `write_resource_parquet()`

### LINKED ISSUE #642: Create internal flow diagram for `write_resource_parquet()`

lwjohnst86:
## Description

Based on @martonvago's suggestion, I'll write things in "pseudocode" from now on. But instead of pseudocode, I will write an outline of the Python function with how I think it might flow inside. Plus, I can write the full docstrings inside, so you all don't need and we don't need to move it over from the Quarto doc. **I have NOT ran this, tested it, or did any execution, this is purely how I think it might work**, hence "pseudo" :stuck_out_tongue:. I'll add some comments directly to the code in the PR.

Closes #642

## Checklist

- [x] Updated documentation


docs/design/implementation/python-functions.qmd
```diff
@@ -215,11 +215,7 @@ fig-alt="A Plant UML schematic of the detailed code flow within the `write_resou
 ::: {.callout-note collapse="true"}
 ### `write_resource_parquet(raw_files, path)`
 
-This function takes the files provided by `raw_files` and merges them
-into a `data.parquet` file provided by `path`. Use
-`path_resource_data()` to provide the correct path location for `path`
-and `path_resource_raw_files()` for the `raw_files` argument. Outputs
-the path object of the created file.
+See `help(write_resource_parquet)` for more details.
 :::
```
lwjohnst86, line 219: I might decide to delete this entirely, since we will soon have the docstrings auto-generated into the website for us :grin:

sprout/core/write_resource_parquet.py
```diff
@@ -0,0 +1,79 @@
+# ruff: noqa
```
lwjohnst86, line 1: Delete this later, but it just stops the linter from processing this file, since it is not valid Python code yet.

sprout/core/write_resource_parquet.py
```diff
@@ -0,0 +1,79 @@
...
+def write_resource_parquet(
+    raw_files_path: list[Path], parquet_path: Path, properties_path: Path
```
lwjohnst86, line 3: I couldn't decide how to handle the properties part, or even whether we should do any verification against the properties. My reasoning here is that may, just maybe, someone might make a change in the files outside of Sprout. So I thought, we can assume generally that the raw data are already verified, but can't guarantee it. Better to be safe than sorry!

sprout/core/write_resource_parquet.py
```diff
@@ -0,0 +1,79 @@
+# ruff: noqa
+def write_resource_parquet(
```
lwjohnst86, line 2: Maybe a better name would be `write_resource_raw_to_parquet()`? Alternatively, create another function that reads from the raw files and outputs a DataFrame, so that this functions first argument is `data`. Thoughts?

sprout/core/write_resource_parquet.py
```diff
@@ -0,0 +1,79 @@
...
+    that the files stored in the `resources/raw/` folder have already been
+    verified and validated, this function does some quick verification checks
+    of the data after reading it into Python from the raw file(s) by comparing
+    with the current properties given by the `properties_path`. All data in the
+    `resources/raw/` folder will be merged into one single data object and then
+    written back to the Parquet file. The Parquet file will be overwritten.
+
+    Examples:
+
+        ``` python
```
lwjohnst86, line 18: Based on the decision post for using quartodoc to generate the docstrings into the website, we can actually include executable code in the examples via using `{python}`. Without the `{}`, it will not execute. Just something to keep in mind, if we want to show users what it might look like when executed.

sprout/core/write_resource_parquet.py
```diff
@@ -0,0 +1,79 @@
...
+
+def write_parquet(data: DataFrame, path: Path) -> Path:
+    return path
+
+
+def read_raw_files(paths: list[Path]) -> DataFrame:
+    # Can read gzip files.
+    data_list = [polars.read_csv(path) for path in paths]
+    # Merge them all together.
+    data = polars.concat(data_list)
```
lwjohnst86, line 67: I added this real code here because I was curious how Polars might do it. It seems it can!

sprout/core/write_resource_parquet.py
```diff
@@ -0,0 +1,79 @@
...
+    # Confirms that the data matches the resource properties found in `datapackage.json`.
+    # Not sure if this is the best solution here to load in the properties file.
+    verify_is_file(properties_path)
+    properties = read_json(properties_path)
+
+    # This function could be several, e.g. verify_data_column_types(), verify_data_column_names()?
+    verify_data(data, properties)
+
+    # Could include validation here?
+    # validate_data(data, properties)
```
lwjohnst86, line 54: I think we can keep this as a TODO item and keep it commented, but not implement it for now.

martonvago (commented):
Very nice!! Just some questions.

I think I've developed some confusion about what the raw data files represent. Are they different versions of the data (with later versions overwriting earlier ones) or different sections of the data (e.g. one file for rows 1-100 and another one for rows 101-200)? Well, I guess there is no reason why they couldn't be used as both...

sprout/core/write_resource_parquet.py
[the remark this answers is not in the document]
lwjohnst86: Yea, I was thinking that, but at least within core, I'd really like to avoid having internal assumptions on where things are located in the folder. Hence why I wrote up the `path_` functions, so that there is that separation. In CLI and in the app, we should definitely have that assumption, but I'm not sure about in core. I'd like to keep it a bit more flexible, just in case.

sprout/core/write_resource_parquet.py
```diff
@@ -0,0 +1,79 @@
...
+        Outputs the path object of the created Parquet file.
+    """
+    # Not sure if this is the correct way to verify multiple files.
+    [verify_is_file(path) for path in raw_files_path]
+
+    data = read_raw_files(raw_file)
+
+    # Confirms that the data matches the resource properties found in `datapackage.json`.
+    # Not sure if this is the best solution here to load in the properties file.
+    verify_is_file(properties_path)
```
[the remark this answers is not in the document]
lwjohnst86: For the first question, yea, that makes sense, to have it the same as in `write_resource_data_to_raw()`. 

For the second question. Both `verify_data()` and `validate_data()` compare against properties. Validating compares against only the `constraints` field. Think of the difference being, with verify, it is checking that a given column is called `age` and that it is a `int` or `float` and that it has a given number of rows. With validate, it is checking that `age` is between 0 and 120 (since most humans don't live past 120), but only if that `constraints` field is provided by the user.

I don't know how frictionless handles this, based on how I read the docs, it seems the validate function does mostly verifying. Which is a bit annoying semantically, but :shrug:... Maybe I'll open an issue in their repo :stuck_out_tongue:

sprout/core/write_resource_parquet.py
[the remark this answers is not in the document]
martonvago: Okay, that makes sense! I'm just thinking that sometimes (not necessarily here) you want to go from one path to another one (e.g. you know the resource folder path and want the path to `datapackage.json`), and I'm not always sure what I'm allowed to assume / how I'm allowed to manipulate them. (Not really a question now, but I'll ask it properly when it comes up 😅 )

sprout/core/write_resource_parquet.py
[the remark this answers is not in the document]
martonvago: Ah okay, I get it now, thanks! Yeah, based on what I've seen, their `validate` does both: it checks the data against the type and number of fields specified in the schema (so verifies it) and also checks it against any constraints (so validates it). If we want to treat these separately, we could try filtering the error messages (like we do for `ResourceError` vs `PackageError`).

Should we treat these separately? I've looked quite a bit at this for [write_resource_data_to_raw](https://sprout.seedcase-project.org/docs/design/implementation/python-functions#fig-write-resource-data-to-raw), so could do it either way.

sprout/core/write_resource_parquet.py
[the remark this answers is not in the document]
martonvago: What would you think about adding some functions that are the inverse of the path functions? So for `path_package(package_id: int) -> Path` we could have `package_id_from_path_package(path: Path) -> int`. This wouldn’t add any new assumptions (other than the invertibility) and it would allow us to go from one path to another via the IDs. We could structure them analogously to the path functions and keep them all together.

lwjohnst86:
@martonvago I forgot to respond to your initial question.

Raw files are kept from the initial upload to keep a record just in case something happens.

A potential scenario might be, a first round of surveys are sent to people and that data gets uploaded to Sprout. That's one raw file. Maybe a few months later, the same survey is sent out and that data gets uploaded. That's another raw file. So those two raw files get merged together and saved as the `data.parquet` file, which would be the file that researchers actually use to do analyses.

sprout/core/write_resource_parquet.py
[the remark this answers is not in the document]
signekb: Probably a good idea to keep things separate

sprout/core/write_resource_parquet.py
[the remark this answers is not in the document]
signekb: In relation to the specifics of this, I'm awaiting the discussion in #826

sprout/core/write_resource_parquet.py
```diff
@@ -0,0 +1,79 @@
...
+
+def read_raw_files(paths: list[Path]) -> DataFrame:
+    # Can read gzip files.
+    data_list = [polars.read_csv(path) for path in paths]
+    # Merge them all together.
+    data = polars.concat(data_list)
+    return data
+
+
+def verify_data(data: DataFrame, properties: dict) -> Path:
```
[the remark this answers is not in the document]
signekb: Are we moving away from `frictionless-py` altogether, or do we want to use it for verification still? In relation to the discussion in #826

signekb (commented):
The overall picture of this makes sense to me as well 👍

docs/design/interface/functions.qmd
```diff
@@ -127,13 +127,21 @@ flowchart
     function --> out
 ```
 
-### {{< var wip >}} `write_resource_parquet(raw_files, path)`
+### {{< var wip >}} `build_resource_parquet(raw_files_path, resource_properties)`
```
lwjohnst86, line 130: I'm unsure of the naming here. And I'm unsure if it should output a DataFrame and have another function `write_resource_parquet()` that does the writing.

docs/design/interface/pseudocode/build_resource_parquet.py
```diff
@@ -0,0 +1,73 @@
...
+) -> Path:
+    """Merge all raw resource file(s) and write into a Parquet file.
+
+    This function takes the file(s) provided by `raw_files_path` and merges them into
+    a `data.parquet` file. The Parquet file will be stored at the path found in `ResourceProperties.path`.
+    While Sprout generally assumes
+    that the files stored in the `resources/raw/` folder have already been
+    verified and validated, this function does some quick verification checks
+    of the data after reading it into Python from the raw file(s) by comparing
+    with the current properties given by the `resource_properties`. All data in the
```
lwjohnst86, line 13:
```suggestion
    While Sprout generally assumes
    that the files stored in the `resources/raw/` folder are already correctly
    structured and tidy, it still runs checks to ensure the data are correct
    by comparing to the properties. All data in the
```

martonvago (changes requested):
This looks very sensible to me :grin:

martonvago (approved):
:partying_face:

Issue #801, opened 2024-10-22
Create and test `validate_raw_data` function (part of `write_resource_data_to_raw`)


martonvago:
- Signature: `validate_raw_data(path: Path) -> Path`
- Should validate the data file against the resource properties.
- See [diagram](https://sprout.seedcase-project.org/docs/design/implementation/python-functions#fig-write-resource-data-to-raw).

PR #824, opened 2024-10-29
feat: :sparkles: add `check_raw_data` function

### LINKED ISSUE #800: Create and test `verify_raw_data` function (part of `write_resource_data_to_raw`)
- Signature: `verify_raw_data(path: Path) -> Path`
- Should check the readability and structure of the data file the path points to. We should probably define what extactly we mean by this.

martonvago:
## Description

This PR adds a `check_raw_data` function. This is a combination of `verify_raw_data` and `validate_raw_data` because `frictionless` has one function for these. We can still split the function into two based on the error codes, but I wanted to put it up for discussion before making it more complicated.

Feel free to suggest better names for `check_raw_data` and the error it throws (if we're keeping them).

Note that I added `resource_properties: dict` as an argument instead of reading it from file inside `check_raw_data`. I'll put up another PR for loading the resource properties from `datapackage.json`, because this seemed like a separate task and raised its own questions.

Closes #800, #801

## Checklist

- [x] Added or updated tests
- [x] Tests passed locally
- [x] Build passed locally
- [x] Updated documentation

### DIFF
--- a/sprout/core/check_raw_data.py
+++ b/sprout/core/check_raw_data.py
@@ -1,0 +1,35 @@
+from pathlib import Path
+
+from frictionless import Resource
+
+from sprout.core.get_extension import get_extension
+from sprout.core.get_report_errors import get_report_errors
+from sprout.core.mismatched_raw_data_error import MismatchedRawDataError
+from sprout.core.path_sprout_root import path_sprout_root
+
+
+def check_raw_data(resource_properties: dict, data_path: Path) -> Path:
+    """Checks a raw data file against a set of resource properties.
+
+    Args:
+        resource_properties: The resource properties to check against.
+        data_path: The path to the raw data file.
+
+    Returns:
+        The path to the raw data file, if the data match the properties.
+
+    Raises:
+        MismatchedRawDataError: If the data do not match the properties.
+    """
+    resource = Resource.from_descriptor(resource_properties)
+    resource.format = get_extension(data_path)
+    resource.scheme = "file"
+    resource.basepath = str(path_sprout_root())
+    resource.path = str(data_path.relative_to(path_sprout_root()))
+
+    report = resource.validate()
+    errors = get_report_errors(report)
+    if errors:
+        raise MismatchedRawDataError(errors, data_path)
+
+    return data_path
--- a/sprout/core/get_extension.py
+++ b/sprout/core/get_extension.py
@@ -1,0 +1,19 @@
+from pathlib import Path
+
+from sprout.core.verify_is_file import verify_is_file
+
+
+def get_extension(path: Path) -> str:
+    """Returns the extension of the file `path` points to without a leading `.`.
+
+    Args:
+        path: The path to the file.
+
+    Returns:
+        The extension of the file.
+
+    Raises:
+        FileNotFoundError: If `path` does not point to a file.
+    """
+    verify_is_file(path)
+    return path.suffix[1:]
--- a/sprout/core/get_report_errors.py
+++ b/sprout/core/get_report_errors.py
@@ -1,0 +1,17 @@
+from frictionless import Error, Report
+
+
+def get_report_errors(report: Report) -> list[Error]:
+    """Returns all errors in a Frictionless report.
+
+    Args:
+        report: The report to get the errors from.
+
+    Returns:
+        The errors in the report.
+    """
+    return [
+        error
+        for error in report.errors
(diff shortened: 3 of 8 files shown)


sprout/core/check_raw_data.py
```diff
@@ -0,0 +1,35 @@
...
+        resource_properties: The resource properties to check against.
+        data_path: The path to the raw data file.
+
+    Returns:
+        The path to the raw data file, if the data match the properties.
+
+    Raises:
+        MismatchedRawDataError: If the data do not match the properties.
+    """
+    resource = Resource.from_descriptor(resource_properties)
```
martonvago, line 24: When `frictionless` validates a resource it checks both the resource properties and the table schema, including any constraints on the fields. It locates the data file to compare with the schema by following the `path` property of the resource. So here I'm constructing a resource that has the resource properties loaded from `datapackage.json`, except that the path is pointed to the raw data file. This doesn't affect the original resource properties.

sprout/core/check_raw_data.py
```diff
@@ -0,0 +1,35 @@
...
+    resource.format = get_extension(data_path)
+    resource.scheme = "file"
```
martonvago, line 26: These are needed for being able to open the file.

sprout/core/check_raw_data.py
```diff
@@ -0,0 +1,35 @@
...
+    resource.basepath = str(path_sprout_root())
+    resource.path = str(data_path.relative_to(path_sprout_root()))
```
martonvago, line 28: I'm doing this because `path` cannot be an absolute path. Since it has to be made into a relative path, it made sense to make it relative to sprout root. To ensure that `frictionless` can find the data file, I'm setting sprout root to be the `basepath` of the resource. This has the added advantage of working with the tests (for which I am infinitely grateful).

sprout/core/get_extension.py
martonvago, line 1: A wondrously complex function

sprout/core/get_report_errors.py
martonvago, line 1: I just pulled this out from `verify_properties_are_well_formed.py`

tests/core/test_check_raw_data.py
```diff
@@ -0,0 +1,122 @@
...
+from sprout.core.mismatched_raw_data_error import MismatchedRawDataError
+from sprout.core.properties import ResourceProperties, TableSchemaProperties
+from sprout.core.write_file import write_file
+
+
+@fixture
+def sprout_root(tmp_path, monkeypatch) -> Path:
+    ROOT = tmp_path / "root"
+    monkeypatch.setenv("SPROUT_ROOT", str(ROOT))
+    return ROOT
```
martonvago, line 15: Setting this so everything can happen inside temporary dirs

tests/core/test_check_raw_data.py
```diff
@@ -0,0 +1,122 @@
...
+    message = str(error.value)
+    for code in error_codes:
+        assert code in message
+
+
+def test_accepts_matching_data_with_constraints(path_to_data, resource_properties):
+    """Should accept a data file that conforms to schema constraints."""
+    resource_properties["schema"]["fields"] += [
+        {"name": "id", "type": "integer"},
+        {"name": "name", "type": "string", "constraints": {"maxLength": 10}},
```
martonvago, line 106: I've discovered that `frictionless` expects constraint names in camel case (hence `maxLength`), so we'll have to have a think about what to do with our class attribute names.

sprout/core/check_raw_data.py
```diff
@@ -0,0 +1,35 @@
...
+
+    report = resource.validate()
```
lwjohnst86, line 30: Only going to make a small comment here, but looking through the frictionless source code (which, personally, isn't well designed), it isn't clear how they do the [validation](https://github.com/frictionlessdata/frictionless-py/blob/main/frictionless/validator/validator.py) (I had to dig for finally finding how `resource.validate()` works).

My instinct is that we may have to either 1) write our own checks because it gives me anxiety looking through and trying to understand their code, or 2) look for other verify/validate packages available. My guess, given this is a very common need, there will be better designed tools out there for these purposes.

signekb: Just tagging the relevant issue: #826

signekb:
@martonvago @lwjohnst86 should I hold off on reviewing this until #826 is closed?

martonvago:
Replaced by #1060, so closing

PR #827, opened 2024-10-30
chore: :hammer: add pre-commit config with commit linting

### LINKED ISSUE #8: Use commitlint to check that commits match use for semantic release
[commitlint](https://github.com/conventional-changelog/commitlint) can be used to check that commits follow a pattern in order to effectively use (python) semantic release. Using the action is described [here](https://github.com/conventional-changelog/commitlint/blob/master/docs/guides/ci-setup.md)

lwjohnst86:
## Description

This PR implements and tests out the commit linting, which I've also made a PR on `template-python-project` repo for too.

Pre-commit recommends to run pre-commit right after installing it, which resulted in a lot of files being changed. They were only whitespace and end of file changes so I will open a new PR stacked on this one for that.

Closes seedcase-project/template-python-project#8


signekb (commented):
I feel like I might have missed something: Do we want to use pre-commit now? Also locally? How is the pre-commit hooks here (trailing white space and end of file fixer) related to commit linting? :)

martonvago (dismissed):
Looks good to me! (Would we want to include ruff linting/formatting in the pre-commit hook as well?)

lwjohnst86:
@signekb commit linting is most effective *before* the message is committed. So the only way to achieve this is with a pre-commit hook. So that's why I started setting up the pre-commit. Plus, the best way that the CI is implemented for commit linting is through the pre-commit CI, otherwise I would have had to write something myself. 

The other stuff added related to the whitespace etc were from the default setup of pre-commit, so I assumed they were needed. Maybe they weren't but they are good basic checks anyway (something I assumed VSCode did already with the formatting, but it seems not).

I'll do a demo of this once everything is merged for using it locally. We don't *have* to use it locally, but I think we should :stuck_out_tongue:

lwjohnst86:
@martonvago yes, I think adding ruff and other linting would be fantastic! I wanted to start with this first though :grin:

PR #828, opened 2024-10-30
style: :art: ran pre-commit, only fixes trailing whitespaces and end of file newlines

lwjohnst86:
## Description

This is a stacked PR. There's a lot that was changed. Thankfully, it is only trailing whitespaces on lines and fixes end of file newlines.

This PR needs a quick review.


signekb (dismissed):
I understand the removal of whitespaces, but why are the svgs deleted and added again?

martonvago (dismissed):

lwjohnst86:
@signekb the SVGs had trailing whitespace/incorrect file ending, which were fixed by the pre-commit. They weren't deleted and re-added :relaxed: SVG is plain text, that's why.

Issue #830, opened 2024-10-31
[discussion]: Verifying package and resource properties


martonvago:
This is just to keep track of some questions about how we verify package and resource properties. This is not to say that big changes are necessary, only to make sure that we consider them fully.
- How will the move to a new library for validation/verification affect the way we verify package and resource properties? Do we want to keep using `frictionless-py` for verification (currently in `verify_properties_are_well_formed`)? Related to #826.
- How will the new design for the `*Properties` classes affect the way we verify package and resource properties? Related to #808.
- What forms can a `*Properties` object take (e.g. any `*Properties` instance, default `*Properties` instance, `compact_dict` representation of the instance) and when do we want to consider each of these as correct?

martonvago:
These questions have either been answered or are no longer relevant 🎉

PR #832, opened 2024-11-01
fix: 🩹  update `verify_properties_are_well_formed()` based on changes to the `Properties` classes

signekb:
## Description

NB this is a stacked PR merging to #808 

This PR changes the behaviour of `verify_properties_are_wellformed()` based on the changes to the `Properties` classes in #808 

This work was done in collaboration with @martonvago 🌷 It's not pretty, but it gets the work done for now.
During this, we talked about #830 and how we think not using `frictionless-py` could simplify the work here as well, since the `validate()`function keeps surprising us by not working as we expect it to.

## Checklist

- [X] Added or updated tests
- [X] Tests passed locally

### DIFF
--- a/sprout/core/verify_properties_are_well_formed.py
+++ b/sprout/core/verify_properties_are_well_formed.py
@@ -24,7 +24,7 @@ def verify_properties_are_well_formed(properties: dict, error_type: str) -> dict
     non_empty_properties = {
         key: value for key, value in properties.items() if value != ""
     }
-    report = validate(non_empty_properties)
+    report = validate(non_empty_properties, type=error_type.replace("-error", ""))
 
     errors = [
         error
--- a/sprout/core/verify_properties_are_well_formed.py
+++ b/sprout/core/verify_properties_are_well_formed.py
@@ -1,4 +1,5 @@
 from frictionless import validate
+from frictionless.errors import Error
 
 from sprout.core.not_properties_error import NotPropertiesError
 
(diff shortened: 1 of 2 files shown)


sprout/core/verify_properties_are_well_formed.py
```diff
@@ -24,14 +25,17 @@ def verify_properties_are_well_formed(properties: dict, error_type: str) -> dict
...
     }
-    report = validate(non_empty_properties)
+    report = validate(non_empty_properties, type=error_type.replace("-error", ""))
 
     errors = [
         error
         for error in report.errors
-        + [error for task in report.tasks for error in task.errors]
-        if error.type == error_type
+        if "required" not in error.message and error.type == error_type
```
signekb, line 33: This workaround, for example, is very non-ideal. 
1. To filter the required field errors, we have to go into the error message 
2. We have to filter on error type twice (here and on L28) to ensure we only catch the errors we intend to (either `PackageError` or `ResourceError)`

sprout/core/verify_properties_are_well_formed.py
```diff
@@ -24,14 +25,17 @@ def verify_properties_are_well_formed(properties: dict, error_type: str) -> dict
     non_empty_properties = {
         key: value for key, value in properties.items() if value != ""
     }
-    report = validate(non_empty_properties)
+    report = validate(non_empty_properties, type=error_type.replace("-error", ""))
```
signekb, line 28: Also not pretty 🥀

martonvago (approved):
Thanks!! Approving but agreeing with all your comments

lwjohnst86 (approved):
Yea, agreed. Too bad about frictionless's `validate()` :unamused:

PR #836, opened 2024-11-04
docs: :lipstick: update landing page with an image

lwjohnst86:
## Description

I found an image a week ago that looked good and though to update the landing page with it. Finally got around to finishing that.

This PR needs a quick review.


K-Beicher (dismissed):
That does look rather nice!

martonvago (dismissed):
Niiice! (I first thought the cloud was watering that potted plant, which would have been very on brand)

index.qmd
```diff
@@ -1,23 +1,33 @@
...
-image-title: Image from Storyset
-image-alt: Illustration of two people looking at data charts together.
 about:
   id: hero-heading
-  template: solana
-  image-width: 10em
+  template: jolla
+  image: "docs/images/home.svg"
+  image-title: Image from Storyset.
+  image-alt: Illustration of a woman interacting with digital files and folders, with database objects on the ground in front of her.
```
signekb, line 12:
```suggestion
  image-alt: Illustration of a woman interacting with digital files and folders, with databases and a folder on the ground in front of her.
```
Mostly bc I'm not sure about "database objects"

docs/images/home.svg
signekb, line 1: I do find this illustration a bit too busy for my taste. 
Maybe we could remove the background, like this? 
Data points-rafiki (2)

signekb: If you prefer the other one, let's just go with that 👍

PR #841, opened 2024-11-07
feat: ✨ add `create_raw_file_name()`

### LINKED ISSUE #802: Create and test `create_raw_file_name` function (part of `write_resource_data_to_raw`)
- Signature: `create_raw_file_name(path: Path) -> Path`. Or could this return `str`?
- Should create a file name of the form `<timestamp>-<uuid>.<ext>.gz`.

signekb:
## Description 

This PR adds the function `create_raw_file_name()` with a test. 
For this, I also added a helper function `get_compact_iso_timestamp()` that generates a compact version of the iso timestamp from `get_iso_timestamp(), by removing characters as ":", "T", and "-". It does, however, add a "-" between the data and time for readability.

Closes #802 

## Checklist

- [X] Added or updated tests
- [X] Tests passed locally

### DIFF
--- a/sprout/core/get_iso_timestamp.py
+++ b/sprout/core/get_iso_timestamp.py
@@ -1,5 +1,4 @@
 from datetime import datetime
-from re import sub
 
 
 def get_iso_timestamp() -> str:
@@ -17,4 +16,4 @@ def get_compact_iso_timestamp() -> str:
     Returns:
         The timestamp as a string. E.g. `20240514-05000100`.
     """
-    return sub(r"\W", "", get_iso_timestamp()).replace("T", "-")
+    return datetime.now().strftime("%Y-%m-%dT%H%M%SZ")
--- a/sprout/core/get_iso_timestamp.py
+++ b/sprout/core/get_iso_timestamp.py
@@ -2,7 +2,7 @@ from datetime import datetime
 
 
 def get_iso_timestamp() -> str:
-    """Generates a timestamp compliant with the Data Package spec.
+    """Generates an ISO timestamp compliant with the Data Package spec.
 
     Returns:
         The timestamp as a string. E.g. `2024-05-14T05:00:01+00:00`.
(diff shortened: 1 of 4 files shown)


sprout/core/get_iso_timestamp.py
```diff
@@ -8,3 +9,12 @@ def get_iso_timestamp() -> str:
         The timestamp as a string. E.g. `2024-05-14T05:00:01+00:00`.
     """
     return datetime.now().astimezone().isoformat(timespec="seconds")
+
+
+def get_compact_iso_timestamp() -> str:
+    """Generates a compact timestamp.
+
+    Returns:
+        The timestamp as a string. E.g. `20240514-05000100`.
```
lwjohnst86, line 18: Good to convert timestamp to something that is more friendly, especially since some computers don't like `:` in the filename.

But, this isn't ISO 8601 compliant, e.g. lots of Stack Exchange posts on this topic of filenames and timestamps like [here](https://serverfault.com/questions/292014/preferred-format-of-file-names-which-include-a-timestamp/370766#370766). 

A format that adheres to ISO 8601 would be something like `yyyy-mm-ddThhmmssZ` or `yyyymmddThhmmssZ`. Both the `T` and `Z` are necessary to maintain compliance. The `Z` being to indicate UTC timezone, which we can for now set as the default (I don't think the exact time they uploaded will be an issue, but maybe?). I kinda prefer the one with the hyphens because it is a bit easier to read, but it mixes the ISO 8601 format a bit (without `:` but with `-`) so is less compliant the the one with no `-` or `:`. You decide which of the two you like.

lwjohnst86 (changes requested):
Nice :relaxed:

sprout/core/get_iso_timestamp.py
signekb: Oh, yes! You're right! Sorry about that. 
Good idea with just adding the `Z` for UTC. I also don't think that will be an issue; I would think that if anything happens, we'll be mostly concerned with "relative time" - e.g., which file was uploaded first/last. 
I'm 100 % for the `-` version 👍

lwjohnst86 (approved):
:grin:

PR #848, opened 2024-11-11
refactor: 🔥 remove everything Django-related

### LINKED ISSUE #845: Simplify development and repo by removing all Django stuff
The development feels weird to keep a lot of that stuff even though we don't use it, it no longer reflects the way we do things, and it adds extra files and unnecessary things to the development process (e.g. the GitHub actions).

So maybe it's time to remove all Django content? Especially if we end up not using Django in the end.

signekb:
## Description

This PR removes all folders, files, tests, dependencies, just recipes, etc. that are related to our use of Django. We are not currently using Django and if we decide to use it again later on, the work removed here would need a complete refactor/rebuild anyway, since the Django app was created before we decided to rely on the Data Package standard. 

Closes #845 

This PR needs an in-depth review. Is there something I have forgotten to remove? 🔥 

## Checklist

- [X] Added or updated tests
- [ ] Tests passed locally (no - see #851  for fix of the failing tests) 


Dockerfile
```diff
@@ -13,14 +13,11 @@ RUN pip install --no-cache-dir poetry==1.7.1
...
+RUN poetry install --no-root --no-interaction
 
 # Copy all code to image
 COPY . /code
 
-# Create Django migrations (You may remove this when migrations are included in git)
-RUN poetry run python manage.py makemigrations --no-input
-
 # Move static assets to STATIC_ROOT
 RUN poetry run python manage.py collectstatic --no-input
```
lwjohnst86, line 22:
```suggestion
```

You could even delete this whole file for now.

lwjohnst86 (changes requested):
:fire: Very small comment.

lwjohnst86 (approved):
:tada:

PR #851, opened 2024-11-11
fix: 🐛  Add UTC timezone to tests using timemachine

signekb:
## Description

This PR fixes the tests failing due to the timezone not being set. I *think* they started failing bc of the switch to "winter time", but I'm actually not entirely sure.  

This PR needs a quick review.

For simplicity (i.e., to avoid having to look at all the currently failing tests on `main`), this PR is stacked on #848 

## Checklist

- [X] Added or updated tests
- [X] Tests passed locally


K-Beicher (dismissed):
Looks like that is the standard fix to this. Can't see anything wrong.

martonvago:
Ah, look:
Screenshot 2024-11-12 at 15 18 50
I played around with it a bit and my guess is that Django was setting the timezone to UTC when no other time zone was provided. That would explain why these tests pass on main but not here.

martonvago (dismissed):
Nice!! 👍👍👍

lwjohnst86 (dismissed):

PR #872, opened 2024-11-13
docs: :memo: describe split between `lib` and `core`

### LINKED ISSUE #850: Document split between low level core and high level core
E.g. lower level core Python code is flexible, while higher level is more abstracted out and similar to CLI (since there's a lot of small little things that some functionality does that would be nice to have bundled as a higher level function).

### LINKED ISSUE #869: Add component called `lib` (?) that differentiates from lower level, flexible core
As I have been writing and thinking about the data resource side of things, plus with some comments @martonvago made about knowing when to assume certain path locations or not, some things felt a bit "ragged"/"rough". I thought, taking inspiration for how the CLI might behave, maybe it would be better to have a very low level, very flexible, less assumption-based (not dependent on specific path locations at all) `core` and a higher level, more opinionated and assumption based `lib` for working with data packages and our data organization in Python (not as the CLI). 

I'll write a draft in the design section to make these thoughts clearer.

lwjohnst86:
## Description

As I was writing the guide on creating data resources and late night swirling thoughts, combined with potential users who would code in Python, I thought that there should actually be Python code that mimicks what the CLI will do... meaning code that has a lot more assumptions and opinions tied to it. Based on discussions we've had over the months, I see a lot of value in having a highly flexible `core`, that would allow us to easily integrate e.g. plugins/extensions, not be tied to the specific folder and file structure we prescribe, and allow others to customize things a lot more to their needs. But there is also a lot of value in having some code that is very opinionated and simple. For instance, not needing to pass around `path_*()` for all functions and just assume "properties will always be the root of the package", etc.

Closes #850 and closes #869

### DIFF
--- a/docs/design/architecture/modular-design.qmd
+++ b/docs/design/architecture/modular-design.qmd
@@ -59,10 +59,3 @@ each split is used.
     imports the required functions from `seedcase_sprout.core`, with
     functions from other Python packages that convert that code into a
     web app.
-
--   The web API is in `seedcase_sprout/api/`. Each endpoint imports the
-    required functions from `seedcase_sprout.core` with functions from
-    other Python packages that convert that code into a web API. This
-    particular module may not be necessary as many Python packages that
-    create web apps are also able to easily create web APIs from the
-    same or slightly modified code.
--- a/docs/design/architecture/modular-design.qmd
+++ b/docs/design/architecture/modular-design.qmd
@@ -14,7 +14,7 @@ In order to achieve our aims, the main functionality is designed and
 implemented as a Python package and can be used as a Python package.
 This code is stored in the Python module (folder location)
 `seedcase_sprout/core/` and is accessible via the Python code
-`import seedcase_sprout.core` statement. Only external-facing functions
+`import seedcase_sprout.core as sp` statement. Only external-facing functions
 would be exposed to the user and to other programs.
 
 The `core` functions and classes aim to have as few external assumptions
@@ -40,9 +40,9 @@ interfaces to Sprout that suit their particular needs.
 Our [Guide](/docs/guide/index.qmd) has examples and tutorials on how
 each split is used.
 
--   A higher-level, opinionated Python implementation of `core`, named
+-   A higher-level, opinionated Python abstraction of `core`, named
     `lib` and stored in `seedcase_sprout/lib/`. This module abstracts
-    away and simplified many of the steps and functions available in
+    away and simplifies many of the steps and functions available in
     `core`, with the consequence of being less flexible and less
     customizable. This module is intended for users who want to use
     Sprout without having to worry about the finer details of the exact
--- a/docs/design/architecture/modular-design.qmd
+++ b/docs/design/architecture/modular-design.qmd
@@ -13,36 +13,56 @@ the context of Sprout.
 In order to achieve our aims, the main functionality is designed and
 implemented as a Python package and can be used as a Python package.
 This code is stored in the Python module (folder location)
-`sprout/core/` and is accessible via the Python code
-`import sprout.core` statement. Only external-facing functions would be
-exposed to the user and to other programs.
+`seedcase_sprout/core/` and is accessible via the Python code
+`import seedcase_sprout.core` statement. Only external-facing functions
+would be exposed to the user and to other programs.
+
+The `core` functions and classes aim to have as few external assumptions
+and dependencies as possible. For instance, input and output don't
+depend on the state of any other function. This way, we can have a
+highly flexible, modular, and easily testable `core` that other
+developers (and potentially users) can easily configure and customize to
+their own needs. This modular and flexible "lower level" design allows
+us to create a variety of interfaces and extensions to Sprout.
 
 Any extensions would then only need to incorporate and depend on the
-`sprout.core` package to get Sprout's functionality. Each extension
-would be its own module within Sprout, with modules named like
-`sprout/extension_name/`, for instance, `sprout/cli/`. That way, within
-the extension's module, the logic that we implement is only specific to
-creating the CLI and not to the actual functionality of Sprout.
+`seedcase_sprout.core` package to get Sprout's functionality. Each
+extension would be its own module within Sprout, with modules named like
+`seedcase_sprout/extension_name/`, for instance, `seedcase_sprout/cli/`.
+That way, within the extension's module, the logic that we implement is
+only specific to creating the CLI and not to the actual functionality of
+Sprout.
 
 Other potential extensions would follow the same or similar pattern.
 This allows other developers to create their own extensions and
 interfaces to Sprout that suit their particular needs.
 
-Our [Guide](../../guide/index.qmd) has examples and tutorials on how
+Our [Guide](/docs/guide/index.qmd) has examples and tutorials on how
 each split is used.
 
--   For our CLI, it is placed in the module `sprout/cli/`. Each command
-    imports the relevant functions from `sprout.core`, along with
-    decorators (from other Python packages) to convert the Python code
-    into a CLI program.
+-   A higher-level, opinionated Python implementation of `core`, named
+    `lib` and stored in `seedcase_sprout/lib/`. This module abstracts
+    away and simplified many of the steps and functions available in
+    `core`, with the consequence of being less flexible and less
+    customizable. This module is intended for users who want to use
+    Sprout without having to worry about the finer details of the exact
+    implementation. The internals of the `lib` functions would be strict
+    and opinionated steps for how we envision a data package to be
+    created and managed.
 
--   The web app is in `sprout/app/`. Each page of the web app imports
-    the required functions from `sprout.core`, with functions from other
-    Python packages that convert that code into a web app.
+-   For our CLI, it is placed in the module `seedcase_sprout/cli/`. Each
+    command imports the relevant functions from `seedcase_sprout.core`,
+    along with decorators (from other Python packages) to convert the
+    Python code into a CLI program.
 
--   The web API is in `sprout/api/`. Each endpoint imports the required
-    functions from `sprout.core` with functions from other Python
-    packages that convert that code into a web API. This particular
-    module may not be necessary as many Python packages that create web
-    apps are also able to easily create web APIs from the same or
-    slightly modified code.
+-   The web app is in `seedcase_sprout/app/`. Each page of the web app
+    imports the required functions from `seedcase_sprout.core`, with
+    functions from other Python packages that convert that code into a
+    web app.
+
+-   The web API is in `seedcase_sprout/api/`. Each endpoint imports the
+    required functions from `seedcase_sprout.core` with functions from
+    other Python packages that convert that code into a web API. This
+    particular module may not be necessary as many Python packages that
+    create web apps are also able to easily create web APIs from the
+    same or slightly modified code.
--- a/docs/design/architecture/modular-design.qmd
+++ b/docs/design/architecture/modular-design.qmd
@@ -48,12 +48,12 @@ each split is used.
     Sprout without having to worry about the finer details of the exact
     implementation. The internals of the `lib` functions would be strict
     and opinionated steps for how we envision a data package to be
-    created and managed.
+    created, structured, and managed.
 
--   For our CLI, it is placed in the module `seedcase_sprout/cli/`. Each
+-   The CLI is placed in the module `seedcase_sprout/cli/`. Each
     command imports the relevant functions from `seedcase_sprout.core`,
     along with decorators (from other Python packages) to convert the
-    Python code into a CLI program.
+    Python code into a CLI.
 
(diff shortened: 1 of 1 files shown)


docs/design/architecture/modular-design.qmd
```diff
@@ -13,36 +13,56 @@ the context of Sprout.
...
-    Python packages that convert that code into a web app.
-
--   The web API is in `sprout/api/`. Each endpoint imports the required
-    functions from `sprout.core` with functions from other Python
-    packages that convert that code into a web API. This particular
-    module may not be necessary as many Python packages that create web
-    apps are also able to easily create web APIs from the same or
-    slightly modified code.
+-   A higher-level, opinionated Python implementation of `core`, named
+    `lib` and stored in `seedcase_sprout/lib/`. This module abstracts
```
lwjohnst86, line 44: Not sure what the right name for this is, but `lib` sounded right. Open to ideas.

docs/design/architecture/modular-design.qmd
```diff
@@ -13,36 +13,56 @@ the context of Sprout.
 In order to achieve our aims, the main functionality is designed and
 implemented as a Python package and can be used as a Python package.
 This code is stored in the Python module (folder location)
-`sprout/core/` and is accessible via the Python code
-`import sprout.core` statement. Only external-facing functions would be
-exposed to the user and to other programs.
+`seedcase_sprout/core/` and is accessible via the Python code
+`import seedcase_sprout.core` statement. Only external-facing functions
```
signekb, line 17:
```suggestion
`import seedcase_sprout.core as sp` statement. Only external-facing functions
```
To include that we use (and hopefully others will start using) shorthand?

docs/design/architecture/modular-design.qmd
signekb: Just so make sure I understand: You want to create `lib` that includes all the assumptions about e.g., `paths` that you wanted to avoid in `core`? Will `lib` then depend on the functionality build in `core` to avoid having the functionality duplicated?

signekb: I'm thinking: Do we *need* core to be without assumptions? Couldn't `lib` (i.e., an opinionated package more closely related to the CLI) just be `core`?

docs/design/architecture/modular-design.qmd
```diff
@@ -13,36 +13,56 @@ the context of Sprout.
...
-
--   The web API is in `sprout/api/`. Each endpoint imports the required
-    functions from `sprout.core` with functions from other Python
-    packages that convert that code into a web API. This particular
-    module may not be necessary as many Python packages that create web
-    apps are also able to easily create web APIs from the same or
-    slightly modified code.
+-   A higher-level, opinionated Python implementation of `core`, named
+    `lib` and stored in `seedcase_sprout/lib/`. This module abstracts
+    away and simplified many of the steps and functions available in
```
lwjohnst86, line 45:
```suggestion
    away and simplifies many of the steps and functions available in
```

docs/design/architecture/modular-design.qmd
```diff
@@ -13,36 +13,56 @@ the context of Sprout.
...
-    the required functions from `sprout.core`, with functions from other
-    Python packages that convert that code into a web app.
-
--   The web API is in `sprout/api/`. Each endpoint imports the required
-    functions from `sprout.core` with functions from other Python
-    packages that convert that code into a web API. This particular
-    module may not be necessary as many Python packages that create web
-    apps are also able to easily create web APIs from the same or
-    slightly modified code.
+-   A higher-level, opinionated Python implementation of `core`, named
```
lwjohnst86, line 43:
```suggestion
-   A higher-level, opinionated Python abstraction of `core`, named
```

docs/design/architecture/modular-design.qmd
lwjohnst86: You got it!

signekb: @lwjohnst86 what do you think about my comment above?

martonvago (commented):
If you think there will be enough stuff in `core` and `lib` separately then I don’t mind! 
- What would be another example of `lib`-specific functionality? Other than the paths / IDs, I mean. I think you mentioned something about how flexible the data upload/adding process could be.
- If the higher-level components (`cli`, `app`) depend directly on `core` and not on `lib`, will we not end up reimplementing some of the functionality of `lib` in these? I’m just thinking that for the same reasons that it would be nice to have these simplifications now, it will be nice to have them in the higher-level components too?

lwjohnst86:
@signekb and @martonvago here are some examples of why we can't simply have `lib` be used in `cli`/`app` and why only `core` would make the experience of coding in Python as an user (not developer) not enjoyable.

Let's take a simple example of making an edit to the description field of the package properties. In `core`, the steps to take currently are (might not be precisely what we have, but it's the idea):

``` python
current_properties = read_properties(path_properties(package_id=1))
updated_properties = edit_package_properties(
	current_properties,
	PackageProperties(description="Blah blah")
)
update_readme(updated_properties, path_readme(1))
write_package_properties(updated_properties, path_package_properties(package_id=1))
```

It would be nicer if it could be something like:

``` python
update_package_properties(
	package_id=1,
	properties=PackageProperties(description=...)
)
```

And internally it will read the current properties, update the README, write back to the properties. 

But this code above can't be used in `cli`, because an assumption of Python code is *no interaction from the user*, while `cli` code often has interaction from the user (e.g. "do you want to delete this field? (y/n)"). Instead, internally, `cli` would use the distinct steps from `core`. For instance, if a user does:

``` bash
sprout update package properties --package_id 1 --description "blah blah"
```

1. Read properties
2. Check that all required fields are included in the properties, if no, ask user to include some text for those fields.
3. Update the README with the new properties.
4. Write back to the package properties.
5. Output some basic info back to the user about what was done.

There will be even more checks and code for the `app`, on top of sending output back to the user's screen as they type stuff. None of that would be possible if we used `lib` rather than `core`.

This gets even more complicated with adding resources, e.g. we want to add more raw data files using a plugin for that data format/machine. In `core`, it might be (some functions are not described in design because I realize as writing the `write_resource_parquet()` and the creating resource guide that we are missing several steps and actions):

``` python
write_resource_data_to_raw(
	path_resource_raw(package_id=1, resource_id=1),
	# Path to raw data file on users computer.
)

data = read_raw_files(path_resource_raw_files(1, 1))
cleaned = custom_plugin_for_data_from_specific_machine(data)

resource_properties = read_resource_properties(path_properties(1))
checked = check_data(cleaned, resource_properties)

write_resource_parquet(checked, path_resource_data(1, 1))

updated_properties = update_resource_properties(
	path_properties(1),
	ResourceProperties(
		schema=TableSchemaProperties(
			fields=list[FieldProperties(
				bytes=file_bytes(path_resource_data(1, 1))
			)
		)
	)
)

write_resource_properties(
	updated_properties,
	path_properties(1)
)

update_readme(read_properties(path_properties(1)), path_readme(1))
```

And this is just off the top of my head. That's not a fun coding experience for an non-developer Python user. It would be nicer if it was:

``` python
add_resource_data(
	package_id=1,
	resource_id=1,
	data_path=path....,
	plugin=custom_plugin_action
)
```

Is this clear enough to describe the issue? :face_with_spiral_eyes:

martonvago:
Thanks for the detailed explanation!

I think having `lib` on top of `core` is very well justified! If the user flows in `cli` will be different from the flows in `lib` because there will be more interactivity, then I can see how `cli` couldn't directly use `lib`. 

As for `app`, it's harder for me to say without knowing how the backend and frontend will interact / how integrated they will be, especially now that maybe we won't use Django. If the backend implementation will have lots of extra bits (like we would have had with Django), then I can see how it couldn't rely on `lib`. If, on the other hand, the backend logic is largely the same, then I can imagine hooking a frontend onto functions like `update_package_properties` in `lib`, provided that they throw helpful errors.

martonvago (dismissed):

signekb:
@lwjohnst86 Yes, thanks for elaborating! 
I think that I just needed some time to get some of these new pieces to fit together, and to me, it actually makes sense to introduce `lib` now as well. Especially in terms of the interactivity we want in the CLI and the app and with having a smoother experience when using the Python package.
I definitely don't have an overview of how everything will look down the line - but this seems to be an appropriate step 👍  
And not as much of a switch of direction as I initially thought 😅

docs/design/architecture/modular-design.qmd
```diff
@@ -13,36 +13,56 @@ the context of Sprout.
 In order to achieve our aims, the main functionality is designed and
 implemented as a Python package and can be used as a Python package.
 This code is stored in the Python module (folder location)
-`sprout/core/` and is accessible via the Python code
-`import sprout.core` statement. Only external-facing functions would be
-exposed to the user and to other programs.
+`seedcase_sprout/core/` and is accessible via the Python code
+`import seedcase_sprout.core as sp` statement. Only external-facing functions
+would be exposed to the user and to other programs.
```
signekb, line 18: In terms of external-facing functions, will this be mostly/only `lib` functions then? Or where do we draw the internal/external line?

docs/design/architecture/modular-design.qmd
```diff
@@ -13,36 +13,56 @@ the context of Sprout.
...
+
+The `core` functions and classes aim to have as few external assumptions
+and dependencies as possible. For instance, input and output don't
+depend on the state of any other function. This way, we can have a
+highly flexible, modular, and easily testable `core` that other
+developers (and potentially users) can easily configure and customize to
+their own needs. This modular and flexible "lower level" design allows
```
signekb, line 25: How do we imagine developers and potentially users would want to configure/customise to their own needs? Would that be in the form of extensions?

docs/design/architecture/modular-design.qmd
```diff
@@ -13,36 +13,56 @@ the context of Sprout.
...
-    slightly modified code.
+-   A higher-level, opinionated Python abstraction of `core`, named
+    `lib` and stored in `seedcase_sprout/lib/`. This module abstracts
+    away and simplifies many of the steps and functions available in
+    `core`, with the consequence of being less flexible and less
+    customizable. This module is intended for users who want to use
+    Sprout without having to worry about the finer details of the exact
+    implementation. The internals of the `lib` functions would be strict
+    and opinionated steps for how we envision a data package to be
+    created and managed.
```
signekb, line 51:
```suggestion
    created, structured, and managed.
```
I'm thinking in terms of folder structure here

docs/design/architecture/modular-design.qmd
```diff
@@ -13,36 +13,56 @@ the context of Sprout.
...
+
+-   For our CLI, it is placed in the module `seedcase_sprout/cli/`. Each
```
signekb, line 53:
```suggestion
-   The CLI is placed in the module `seedcase_sprout/cli/`. Each
```

docs/design/architecture/modular-design.qmd
```diff
@@ -13,36 +13,56 @@ the context of Sprout.
...
+    command imports the relevant functions from `seedcase_sprout.core`,
+    along with decorators (from other Python packages) to convert the
+    Python code into a CLI program.
```
signekb, line 56:
```suggestion
    Python code into a CLI.
```

docs/design/architecture/modular-design.qmd
```diff
@@ -13,36 +13,56 @@ the context of Sprout.
...
+
+-   The web app is in `seedcase_sprout/app/`. Each page of the web app
+    imports the required functions from `seedcase_sprout.core`, with
+    functions from other Python packages that convert that code into a
+    web app.
+
+-   The web API is in `seedcase_sprout/api/`. Each endpoint imports the
```
signekb, line 63: This might be my confusion/ignorance, but couldn't we just remove this `api` split for now?

signekb (changes requested):
Just have some additional questions/suggestions :)

docs/design/architecture/modular-design.qmd
lwjohnst86: Yea, extensions, not using the path structure we've decided on, adding extra processing steps that might not be suitable for a full extension/plugin, or if they just want to use specific functions within sprout without using our full workflow/process. A bit like how we'd like individual/finer-grained functionality from frictionless, but it isn't possible to do.

docs/design/architecture/modular-design.qmd
lwjohnst86: That's fair! We can always add it back later.

docs/design/architecture/modular-design.qmd
```diff
@@ -13,36 +13,56 @@ the context of Sprout.
...
+    required functions from `seedcase_sprout.core` with functions from
+    other Python packages that convert that code into a web API. This
+    particular module may not be necessary as many Python packages that
+    create web apps are also able to easily create web APIs from the
+    same or slightly modified code.
```
lwjohnst86, line 68:
```suggestion

```

PR #875, opened 2024-11-13
chore: :fire: remove last remains of older code and deploy files

lwjohnst86:
## Description

These are the last bits of files that we don't current need, nor will need over the next year (more likely). So just removing them for now.

This PR needs a quick review.

## Checklist

- [x] Added or updated tests
- [x] Tests passed locally
- [x] Build passed locally
- [ ] Updated documentation


martonvago (dismissed):
👍👍👍

signekb (dismissed):
👍 
I was also thinking whether we should remove the two functions `convert_to_human_readable()` and `convert_to_snake_case()` in `core/utils.py` along with their tests?

lwjohnst86:
@signekb good idea, I will also remove those.

PR #877, opened 2024-11-13
chore: :fire: no longer need these files

lwjohnst86:
## Description

This PR removes most of the PlantUML files and docs that focused on the web app. They were super useful at the time and help evolve our understanding of what we wanted to do and build, but are now very out-dated with our current state. So, like the Django removal, these remove all these docs related to the Django and web app stuff.

This PR needs a quick review.


martonvago (approved):

signekb (approved):
👍

PR #881, opened 2024-11-15
build: :wrench: add GitHub workflow to keep track of Data Package versions

### LINKED ISSUE #862: Investigate how to get releases of data package spec from official source
There doesn't seem to be a clear and straightforward way to download the latest release of the spec.

martonvago:
## Description

This adds a GitHub workflow for checking for updates in the Data Package standard. If there is an update, an issue is raised.

I tried this in a test repo and it worked, but we'll see if the permissions will work here as well.

I thought this was a not particularly intrusive way of getting updates from the Data Package project, but we can do something else as well. Feel free to change the wording or add more settings to the issue.

Closes #862 

## Checklist

- [x] Added or updated tests
- [x] Updated documentation
- [x] Ran `just run-all`

### DIFF
--- a/.github/workflows/check-data-package-version.yml
+++ b/.github/workflows/check-data-package-version.yml
@@ -2,8 +2,8 @@ name: Check Data Package standard version
 
 on:
   schedule:
-    # Runs at midnight on Sundays
-    - cron: 0 0 * * 0
+    # Runs every two months
+    - cron: 0 0 1 */2 *
   workflow_dispatch:
 
 env:
@@ -16,9 +16,11 @@ jobs:
     steps:
       - name: Fetch latest version
         run: |
-          latest_version=$(curl -s https://api.github.com/repos/$DP_REPO/releases/latest | jq -r .tag_name)
+          latest_version=$(gh release view --repo $DP_REPO --json tagName -q .tagName)
           echo "latest_version=$latest_version" >> $GITHUB_ENV
           echo "Latest version is $latest_version"
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
       - name: Create an issue if a new version is available
         if: env.latest_version != env.CURRENT_VERSION
@@ -28,10 +30,10 @@ jobs:
             Please check if updating Sprout's copy of the Data Package standard is necessary.
 
             If so:
-            - Replace the old JSON schema of the Data Package standard with the new one downloaded from the website.
-            - Update the current version of \`$DP_REPO\` in the '$GH_WORKFLOW' GitHub workflow."
+              - [ ] Replace the old JSON schema of the Data Package standard with the new one downloaded from the website.
+              - [ ] Update the current version of \`$DP_REPO\` in the '$GH_WORKFLOW' GitHub workflow."
           gh issue create --title "$title" --body "$body"
         env:
-          GH_TOKEN: ${{ secrets.ADD_TO_BOARD }}
+          GH_TOKEN: ${{ secrets.UPDATE_DATAPACKAGE }}
           GH_REPO: ${{ github.repository }}
           GH_WORKFLOW: ${{ github.workflow }}
--- a/.github/workflows/check-data-package-version.yml
+++ b/.github/workflows/check-data-package-version.yml
@@ -2,12 +2,13 @@ name: Check Data Package standard version
 
 on:
   schedule:
-    - cron: '0 0 * * 0' # Runs at midnight on Sundays
+    # Runs at midnight on Sundays
+    - cron: 0 0 * * 0
   workflow_dispatch:
 
 env:
-  CURRENT_VERSION: "v2.0"
-  DP_REPO: "frictionlessdata/datapackage"
+  CURRENT_VERSION: v2.0
+  DP_REPO: frictionlessdata/datapackage
 
 jobs:
   check-latest-version:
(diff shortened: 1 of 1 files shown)


.github/workflows/check-data-package-version.yml
```diff
@@ -0,0 +1,36 @@
...
+          title="New \`$DP_REPO\` version available: $latest_version"
+          body="A new version of [\`$DP_REPO\`](https://github.com/$DP_REPO) is available: $latest_version.
+            Please check if updating Sprout's copy of the Data Package standard is necessary.
+
+            If so:
+            - Replace the old JSON schema of the Data Package standard with the new one downloaded from the website.
+            - Update the current version of \`$DP_REPO\` in the '$GH_WORKFLOW' GitHub workflow."
+          gh issue create --title "$title" --body "$body"
+        env:
+          GH_TOKEN: ${{ secrets.ADD_TO_BOARD }}
```
martonvago, line 34: It looks like in order for the creation of this issue to be picked up by other workflows (like the one that adds it to the project), the token needs to  include the `project` scope. I think our `ADD_TO_BOARD` token has this, but I can't actually look at it, so this is just a guess. Alternatively, we can add it to the project as part of `gh issue create`, but it will also need the `project` scope for that. We can also use the API endpoint for adding the issue and then `GITHUB_TOKEN` seems to be enough.

martonvago:
We can test this by e.g. setting the current version to `v1.0` and triggering it manually.

.github/workflows/check-data-package-version.yml
```diff
@@ -0,0 +1,37 @@
...
+  CURRENT_VERSION: v2.0
+  DP_REPO: frictionlessdata/datapackage
+
+jobs:
+  check-latest-version:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Fetch latest version
+        run: |
+          latest_version=$(curl -s https://api.github.com/repos/$DP_REPO/releases/latest | jq -r .tag_name)
```
lwjohnst86, line 19:
```suggestion
          latest_version=$(curl -s https://api.github.com/repos/$DP_REPO/releases/latest | jq -r .tag_name)
          latest_version=$(gh release list -R $DP_REPO --limit 1 --json name -q ".[].name"
```

Just a slightly clearer, slightly shorter version :stuck_out_tongue: mainly so I can use gh cli in more places :stuck_out_tongue: So this takes the top list of releases (latest) and gives back it's name.

.github/workflows/check-data-package-version.yml
```diff
@@ -0,0 +1,37 @@
...
+      - name: Create an issue if a new version is available
+        if: env.latest_version != env.CURRENT_VERSION
+        run: |
+          title="New \`$DP_REPO\` version available: $latest_version"
+          body="A new version of [\`$DP_REPO\`](https://github.com/$DP_REPO) is available: $latest_version.
+            Please check if updating Sprout's copy of the Data Package standard is necessary.
+
+            If so:
+            - Replace the old JSON schema of the Data Package standard with the new one downloaded from the website.
+            - Update the current version of \`$DP_REPO\` in the '$GH_WORKFLOW' GitHub workflow."
```
lwjohnst86, line 32:
```suggestion

            - [ ] Replace the old JSON schema of the Data Package standard with the new one downloaded from the website.
            - [ ] Update the current version of \`$DP_REPO\` in the '$GH_WORKFLOW' GitHub workflow."
```

Just so they are checklists.

.github/workflows/check-data-package-version.yml
```diff
@@ -0,0 +1,37 @@
+name: Check Data Package standard version
+
+on:
+  schedule:
+    # Runs at midnight on Sundays
+    - cron: 0 0 * * 0
```
lwjohnst86, line 6: This probably doesn't need to be ran that often, maybe even once a month or every 2 months.

lwjohnst86 (changes requested):
Very nice!! Just some comments :grin:

.github/workflows/check-data-package-version.yml
```diff
@@ -0,0 +1,37 @@
...
+          title="New \`$DP_REPO\` version available: $latest_version"
+          body="A new version of [\`$DP_REPO\`](https://github.com/$DP_REPO) is available: $latest_version.
+            Please check if updating Sprout's copy of the Data Package standard is necessary.
+
+            If so:
+            - Replace the old JSON schema of the Data Package standard with the new one downloaded from the website.
+            - Update the current version of \`$DP_REPO\` in the '$GH_WORKFLOW' GitHub workflow."
+          gh issue create --title "$title" --body "$body"
+        env:
+          GH_TOKEN: ${{ secrets.ADD_TO_BOARD }}
```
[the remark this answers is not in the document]
martonvago: Okay! We just need to remember to add it :smile: I don't think I can see the settings

.github/workflows/check-data-package-version.yml
martonvago: Okay! I did it with `view` and `tagName`, but otherwise the same

lwjohnst86 (approved):

lwjohnst86:
I don't know how the synching will work, I *think* that this file won't be affected, but it might get deleted, with the way I've set the synching up. If it does, we'll move this action over into the `.github` repo I think.

signekb (approved):

PR #887, opened 2024-11-20
fix: :bug: compact form of nested properties

martonvago:
## Description

This fixes a problem where `compact_dict` was not removing `None` values from nested objects.

This PR needs a quick review.

## Checklist

- [x] Added or updated tests
- [x] Updated documentation
- [x] Ran `just run-all`

### DIFF
--- a/seedcase_sprout/core/properties.py
+++ b/seedcase_sprout/core/properties.py
@@ -30,9 +30,9 @@ class Properties(ABC):
         """
         return asdict(
             obj=self,
-            dict_factory=lambda tuples: dict(
-                tuple for tuple in tuples if tuple[1] is not None
-            ),
+            dict_factory=lambda tuples: {
+                key: value for key, value in tuples if value is not None
+            },
         )
 
 
--- a/seedcase_sprout/core/properties.py
+++ b/seedcase_sprout/core/properties.py
@@ -3,7 +3,7 @@
 # properties file to add more dataclasses and move them into this file.
 
 from abc import ABC, abstractmethod
-from dataclasses import dataclass, fields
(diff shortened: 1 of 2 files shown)


seedcase_sprout/core/properties.py
```diff
@@ -21,12 +21,19 @@ def default(cls: type[Self]) -> Self:
...
         Returns:
             A dictionary representation of the object with only non-None values.
         """
-        return {key: value for key, value in asdict(self).items() if value is not None}
+        return asdict(
+            obj=self,
+            dict_factory=lambda tuples: dict(
+                tuple for tuple in tuples if tuple[1] is not None
+            ),
+        )
```
martonvago, line 36: This is basically moving the `None`-check inside `asdict`. `asdict` walks the entire data structure, turning all elements into dictionaries recursively. The problem was that we were only filtering `None` values on the outermost layer, leaving nested `None` values in place. Turns out we can pass a custom function to `asdict` for it to use as it processes the data structure. If we add the `None`-check here, it will be applied everywhere recursively.

signekb, line 36: Would this be a bit cleaner, so we avoid the `tuple[1]`? 
```suggestion
        return asdict(
            obj=self,
            dict_factory=lambda tuples: {
                key: value for key, value in tuples if value is not None
            },
        )
```

signekb (changes requested):
Ah! Nice catch! Thanks 🎉 
I have made one suggestion to make it a bit cleaner. But, if you like the way you implemented it more, I will not object ☺️

lwjohnst86 (approved):
:clinking_glasses:

PR #890, opened 2024-11-25
feat: write script to capture overdue reviews of documents

K-Beicher:
added review dates as a test to three documents, two overdue and one not for testing

## Description

This PR adds a first draft of a script which will help us to maintain our documentation and ensure that it is reviewed on a regular basis.

The premise is that we add a comment to all the documents that we would like to include in our regular reviews with a date for next review. This adds a level of flexibility to the review schedule as we can decide for newly created documents (where we are not entirely sure that we have captured everything) that they should be reviewed in three or six months, and for more established documents that they may be fine with a yearly check.

The alternative would be to maintain a list of documents in all repos that we would like to review, or run through everything every half year.

Closes 

## Focus for reviewer

This is a first draft, it will need to be changed so that it points to the repo that the script is stored in (I assuming we can't have a script in one repo which checks all of them).

### DIFF
--- a/docs/design/architecture/c4-models.qmd
+++ b/docs/design/architecture/c4-models.qmd
@@ -58,3 +58,5 @@ This diagram shows the components that make up the database and storage containe
 ![C4 Component diagram showing the parts that make up the database and
 storage containers.](images/c4/component-backend.svg)
 :::
+
+<!--Next Review date: 2024-01-01-->
--- a/docs/design/architecture/modular-design.qmd
+++ b/docs/design/architecture/modular-design.qmd
@@ -46,3 +46,5 @@ each split is used.
     module may not be necessary as many Python packages that create web
     apps are also able to easily create web APIs from the same or
     slightly modified code.
+
+<!--Next Review date: 2025-01-01-->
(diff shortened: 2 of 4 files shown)


lwjohnst86 (changes requested):
Hmm, even though I know you used ChatGPT because you mentioned to me, I still would have guessed it was ChatGPT code. In general, ChatGPT is useful for very basic things, but it has very clear limitations and tendencies to do certain, less than ideal things. For instance the nested for loops, adding extra code to print things (like in the end of the script), adding things that aren't even necessary (e.g. the `except ValueError`), using out-dated ways of working with code (e.g. using `os` when the current way is using `pathlib`), or not knowing about basic functionality (that might be more recent additions like `pathlib.glob()` which would make nearly half the code redundant). It also has no way of understanding *context*, which is something only a human would be able to see. So, while it is useful to get a general idea for where to start, in general, avoid using it as-is and be very very cautious about thinking it does what you want in an effective, clear way.

Because I know/see the code was (mainly) produced by ChatGPT, my comments and suggestions may sound harsher than their intent. This is mainly feedback on why ChatGPT isn't too great to rely on :stuck_out_tongue: 

Also, this script should ideally be in the repo `spaid` or maybe even converted into a GitHub workflow so we can automate it on all repos :grin:

lwjohnst86:
We'll put this on hold for now.

PR #904, opened 2024-12-03
docs: 📝 update hero-text to describe modules and include lib

signekb:
## Description

As well as update the header to fit the text better (in my opinion) :) 

This PR needs a quick review.

### DIFF
--- a/index.qmd
+++ b/index.qmd
@@ -36,7 +36,7 @@ div.quarto-about-jolla {
 ::: {.hero-banner}
 ::: {.landing-page-block}
 ::: {.hero-text}
-## Manage your data in a consistent way
+## Streamlined data management for every workflow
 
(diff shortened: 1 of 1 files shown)


index.qmd
```diff
@@ -36,18 +36,21 @@ div.quarto-about-jolla {
...
 to structure and manage your data.
 
-Sprout has three key components:
+Sprout is a **Python package** with four key modules:
 
--   A **Python package** for organising and managing data
+-   A flexible `core` encapsuling Sprout's core functionalities,
+    directed at developers.
+-   An opinionated `lib` for organising and managing data
     programmatically.
```
signekb, line 50: Not sure this is the best way to describe them. Thoughts?

lwjohnst86 (approved):
That sounds nice to me :relaxed:

PR #924, opened 2024-12-19
feat: :sparkles: Add custom `CheckError` for check functions

### LINKED ISSUE #921: [discussion] How should the `check` package return and display errors?
I've been thinking about how the `check` package should return and display errors.

`jsonschema` creates `ValidationError`s to represent schema violations. These are large objects with lots of information and sometimes include consituent errors/causes. E.g., a resource's `path` can be an array or a string matching a particular pattern (in Sprout it can only be a string, but that doesn't matter for now). If the `path` is a string but it doesn't match the pattern, `jsonschema` will create an error tree like:
```
Error1 triggered by `oneOf` validator (i.e. fixing one of the errors below would make the check pass)
    Error2 triggered by `pattern` validator (i.e. the string doesn't match the pattern)
    Error3 triggered by `type` validator (i.e. the string is not an array)
```

1. Do we want to expose `ValidationError` (a big and complex object) or pare it down to a custom `CheckError` (a small and simple object)?
2. Do we want to keep the nesting (`[Error1: [Error2, Error3]]`) or flatten the list of errors (`[Error1, Error2, Error3]`)?
3. Do we even want to include high-level errors like Error1 (`oneOf`) in our output?

If we choose the simplest option for all 3 questions (custom `CheckError`, no nesting, no high-level errors), we will end up with a non-nested list of simple errors about concrete validators being violated. The upside is that this list will be easy to process and filter in Sprout, and displaying these errors in a human-readable way will also be easy.

The downside is that we lose some information about the relationship between the errors (i.e. the hierarchy) as well as any error attributes we don't include in our custom error.

Does anyone see any use cases where these downsides matter? Maybe other users of the `check` package will want access to the full error information?

martonvago:
## Description

This PR adds a custom `CheckError` to represent violations of individual rules / validators. This error corresponds roughly to `jsonschema`'s `ValidationError`, but it is simpler and includes a bit more information for errors violating required-checks.

Some more detail on why this is nice: [here](https://github.com/seedcase-project/seedcase-sprout/issues/921).
But the main benefits are:
  - Users of the `check` package will not need to depend on `jsonschema` directly.
  - It will be a LOT easier to process errors downstream (so in Sprout).
  - We have more control over what to include in the errors and how to display them.

Closes #921 

Focus on whether this is a reasonable way of structuring and displaying errors.

## Checklist

- [x] Added or updated tests
- [x] Updated documentation
- [x] Ran `just run-all`


seedcase_sprout/core/checks/__init__.py
martonvago, line 1: This is a WIP

seedcase_sprout/core/checks/check_error.py
```diff
@@ -0,0 +1,89 @@
+from functools import total_ordering
+
+
+@total_ordering
```
martonvago, line 4: This makes it so that `CheckError`s can be compared

seedcase_sprout/core/checks/check_error.py
```diff
@@ -0,0 +1,89 @@
...
+        message: str,
+        json_path: str,
+        validator: str,
+    ):
+        """Initialises CheckError.
+
+        Args:
+            message: The error message.
+            json_path: The path to the JSON field within the enclosing JSON object where
+                the error occurred.
```
martonvago, line 19: E.g. `$.resources[2].name`

martonvago: For now I put it as a string, but we can change it to something easier to manipulate later if we need to.

seedcase_sprout/core/checks/check_error.py
```diff
@@ -0,0 +1,89 @@
...
+        Returns:
+            The result of the comparison.
+        """
+        if not isinstance(other, CheckError):
+            return NotImplemented
+        return (self.json_path, self.validator, self.message) < (
+            other.json_path,
+            other.validator,
+            other.message,
+        )
```
martonvago, line 59: So order the errors alphabetically, first by `json_path`, then breaking ties by `validator`, then by `message`

seedcase_sprout/core/checks/get_full_json_path_from_error.py
martonvago, line 1: By default `jsonschema` doesn't add the field name to the path for errors triggered by missing required fields. I think for us / Sprout this would be useful, so I'm adding the field name in our `CheckError`s.

seedcase_sprout/core/checks/to_check_errors.py
```diff
@@ -0,0 +1,38 @@
...
+)
+from seedcase_sprout.core.checks.unwrap_errors import unwrap_errors
+
+COMPLEX_VALIDATORS = {"allOf", "anyOf", "oneOf"}
+
+
+def to_check_errors(validation_errors: Iterator[ValidationError]) -> list[CheckError]:
+    """Transforms `jsonschema.ValidationError`s to more compact `CheckError`s.
+
+    The list of errors is
```
lwjohnst86, line 17:
```suggestion
    The list of errors is:
    
```

seedcase_sprout/core/checks/unwrap_errors.py
```diff
@@ -0,0 +1,18 @@
...
+        errors: A nested list of errors.
+
+    Returns:
+        A flat list of errors.
+    """
+    unwrapped = []
+    for error in errors:
+        unwrapped.append(error)
+        if error.context:
+            unwrapped.extend(unwrap_errors(error.context))
```
lwjohnst86, line 17: There's something about this that I feel is off, but I can't pinpoint it. I'll come back to it another time. I think it's the nested use of the same function plus the for loop in general. :thinking:

seedcase_sprout/core/checks/check_error.py
```diff
@@ -0,0 +1,89 @@
...
+        if not isinstance(other, CheckError):
+            return NotImplemented
+        return (
+            self.message == other.message
+            and self.json_path == other.json_path
+            and self.validator == other.validator
+        )
+
+    def __lt__(self, other: object) -> bool:
+        """Checks if this error is less than an object.
```
lwjohnst86, line 45: I'm a bit confused, what does "less than an object" mean?

lwjohnst86 (changes requested):
It's a bit of a big PR, so will review more of this again tomorrow :relaxed:

seedcase_sprout/core/checks/__init__.py
```diff
@@ -0,0 +1,7 @@
+"""Check functions for the Frictionless Data Package standard."""
+
+from .check_package_properties import check_package_properties
+from .check_properties import check_properties
+from .check_resource_properties import check_resource_properties
+
+__all__ = [check_properties, check_package_properties, check_resource_properties]
```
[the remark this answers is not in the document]
martonvago: You're right, they do!!

seedcase_sprout/core/checks/check_error.py
```diff
@@ -0,0 +1,89 @@
...
+            json_path: The path to the JSON field within the enclosing JSON object where
+                the error occurred.
+            validator: The name of the validator that failed.
+        """
+        self.message = message
+        self.json_path = json_path
+        self.validator = validator
+        super().__init__(self.__str__())
+
+    def __eq__(self, other: object) -> bool:
```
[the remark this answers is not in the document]
martonvago: It's the root of the class hierarchy, so everything is an `object`. And here I want to be able to compare everything with a `CheckError`.

seedcase_sprout/core/checks/check_error.py
martonvago: As in: when I compare a `CheckError` with any other object (like `check_error < 123`), I want `__lt__` to decide whether the error counts as less than that object. Here `__lt__` says that if the other object is not a `CheckError` then it cannot do the comparison. But if the other object is a `CheckError` it can meaningfully compare them and decide which one should come first when they are sorted.

seedcase_sprout/core/checks/check_error.py
```diff
@@ -0,0 +1,89 @@
...
+        if not isinstance(other, CheckError):
+            return NotImplemented
+        return (self.json_path, self.validator, self.message) < (
+            other.json_path,
+            other.validator,
+            other.message,
+        )
+
+    def __hash__(self) -> int:
+        """Returns a hash for this error.
```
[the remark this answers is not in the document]
martonvago: It's like a companion piece to `__eq__`. Some built-in Python functions (e.g. `set`) rely on computing a hash (a unique integer) for the items they operate on. So if I want to be able to do e.g. `set([error1, error2, error1])`, I need `CheckError` to be hashable.

seedcase_sprout/core/checks/unwrap_errors.py
martonvago: It'll probably help that the structure of `ValidationError`s is like:
```json
{
    "some_field": "some value",
    "context": [
        {
            "some_field": "some value",
            "context": [
                {"some_field": "some value", "context": [...]},
                {"some_field": "some value", "context": [...]},
            ],
        },
        {"some_field": "some value", "context": []},
    ],
}
```
So each error can have a `context` property that lists more errors, where, again, each of those may have a `context` with yet more errors. Here we want to traverse this structure to the end and put all nested errors into one flat list.

seedcase_sprout/core/checks/__init__.py
martonvago, line 7:
```suggestion
__all__ = ["check_properties", "check_package_properties", "check_resource_properties"]
```

tests/core/checks/test_to_check_errors.py
```diff
@@ -0,0 +1,214 @@
...
+            message="Low-level error",
+            json_path="$.user.address.street",
+            validator="c-validator",
+        ),
+    ]
+
+
+@mark.parametrize("validator", COMPLEX_VALIDATORS)
+def test_ignores_complex_validators(validator):
+    """Should ignore errors from validators in the COMPLEX_VALIDATORS set."""
```
signekb, line 148: Maybe write this in the function docstring and why it ignores these errrors? :)

tests/core/checks/test_unwrap_errors.py
```diff
@@ -0,0 +1,54 @@
...
+    """Should unwrap a flat list of errors."""
+    assert unwrap_errors(errors) == errors
+
+
+def test_unwraps_nested_errors():
+    """Should unwrap a nested list of errors."""
+    assert unwrap_errors([nested_error]) == [nested_error, simple_error]
+
+
+def test_unwraps_deeply_nested_errors():
```
signekb, line 38: Is there any reason why `unwrap_errors` doesn't deduplicate the errors as well?

signekb (commented):
Sorry for the long wait. 
Great work! 🌟 I only have some minor comments/suggestions/questions.

tests/core/checks/test_to_check_errors.py
martonvago: It says "filtered for summary-type errors"?

signekb (approved):
Nice work 👍 👍

lwjohnst86 (approved):
Very nice! A bit big of a PR, but I'm done reviewing it :grin: I think it would help a lot to have this moved over into it's own package though!! Conceptually makes a lot of sense, but we'll do that maybe next iteration or two.

PR #931, opened 2025-01-03
docs: :memo: revise output design to default to working dir

### LINKED ISSUE #894: Expand on outputs doc to include situation of having data package be created in the working directory
Like `git init`. And maybe specify that `lib` will default to local, since people will likely be using venvs if they work with Python anyway, so might as well assume the venv is a directory (e.g. managed via poetry) and that the directory will be the data package.

lwjohnst86:
## Description

Given that we've moved towards the default behaviour of creating data packages in the working directory, I've revised the output design docs to reflect that change.

Closes #894

This PR needs a quick review.

## Checklist

- [x] Updated documentation
- [ ] Ran `just run-all`


signekb:
@lwjohnst86  and these changes are also in #931 ? :)

lwjohnst86:
@signekb ugh annoying. This shouldn't be in #936 so review this one first :relaxed:

docs/design/interface/outputs.qmd
```diff
@@ -25,104 +23,100 @@ resource (see Note below).
...
+demographics could be one data resource, while the blood sample data
+could be another data resource within the data package.
 :::
 
-### Folders
-
-The folders that Sprout creates and uses are:
-
-`SPROUT_ROOT/`
+## Single data package in working directory
```
martonvago, line 45:
```suggestion
## Single-user local environment
```
Maybe? To match "Multi-user server environment" below. But I don't feel strongly about this :P

martonvago (commented):
Looks good!!

docs/design/interface/outputs.qmd
lwjohnst86: Hmm, it doesn't exactly match the functionality though. The structure below doesn't have to be in a single user nor local environment. Just like Git repos could be on a multi-user server or on a single-user local.

PR #935, opened 2025-01-03
refactor: :truck: rename `SPROUT_ROOT` to `SPROUT_GLOBAL`

lwjohnst86:
## Description

Root has a bunch of different meanings in operating systems (root of a folder, root of the OS, etc). So to help differentiate it more, we'll use `SPROUT_GLOBAL` to indicate the global OS location for Sprout.

I had to rename a lot of Python files, so there are a lot of changes. Git also unfortunately recognized many file name changes as deletions, not file naming differences.

Closes #895

This PR needs a quick review.

## Checklist

- [x] Added or updated tests
- [x] Updated documentation
- [x] Ran `just run-all`

PR #936, opened 2025-01-03
docs: :memo: explain that we follow Seedcase's overall design

lwjohnst86:
## Description

This PR adds a very small edit to the design landing page to explain that we follow principles and patterns (that will eventually be written) found in the design repo.

This PR needs a quick review.

## Checklist

- [ ] Added or updated tests
- [x] Updated documentation
- [ ] Ran `just run-all`


docs/design/interface/outputs.qmd
```diff
@@ -150,24 +145,43 @@ Sprout creates and uses these files:
...
+    reasons, even though they contain the same data.
+
+## Multi-user server environment
+
+When Sprout is used within a server environment where multiple users and
+potentially multiple data packages are stored, the file system structure
+is different. The main difference is that the data packages are stored
+in a central location, rather than in the working directory of the user.
+Instead of a newly created data package being stored in `.`, it will
+default to the below paths.
```
signekb, line 171: I feel like this (and the header in the section above "Single data package in working directory") makes it sound like we don't expect users to work on multiple data packages when Sprout is used in a non-server environment? 
I could definitely see a scenario where someone collects data for multiple, unrelated projects?

signekb:
@lwjohnst86 seems like there's some changed here related to #935. Should this be a stacked PR instead?

docs/design/interface/outputs.qmd
```diff
@@ -150,24 +145,43 @@ Sprout creates and uses these files:
...
+
+`SPROUT_GLOBAL/`
```
signekb, line 173: Will `SPROUT_GLOBAL` not be used *at all* when working in a non-server environment? Only describing it under this header makes it seem so. Was that intentional?

lwjohnst86:
I am not sure what happened to this branch :thinking: :confused:

docs/design/interface/outputs.qmd
lwjohnst86: Yea, if they create multiple data packages, they would create new folders for them first. So no more `<id>` for packages in this particular context.

docs/design/interface/outputs.qmd
lwjohnst86: We can't control what users will do, only what type of assumptions/designs we make for the code. In this case, we make the assumption that local situations don't use GLOBAL so we can write the code with that assumption/design in mind.

martonvago (dismissed):
:+1: :+1: :+1:

Issue #927, opened 2025-01-03
Investigate issues around quartodoc's build of reference docs


lwjohnst86:

signekb:
List of identified issues: 
- [ ] Docstrings with multiline Returns are incorrectly split <img width="601" alt="Image" src="https://github.com/user-attachments/assets/cbe22a62-6640-487a-8b16-f8df253bd458" />
- [ ] When a referencing a function inside a file with the same name, quartodoc incorrectly builds the referene based on the file, not the function <img width="767" alt="Image" src="https://github.com/user-attachments/assets/9fe7fe14-505e-431a-9cd8-a2c9a5b6d424" />
- [ ] Writes out the entire function reference (too long/not pretty and not necessary since they are added in `__init__.py`) <img width="777" alt="Image" src="https://github.com/user-attachments/assets/5cae3651-1128-42b7-9af8-ee4d681dda9d" />

signekb:
This was a tracking issue to identify issues around Quartodoc and our docstrings. For now, I have added the issues found to the list above. 

If I find more issues, I will add them as separate issues to the repo and not add them to the list above.

PR #942, opened 2025-01-06
docs: 🎨 add package to each subsection and remove core before functions

### LINKED ISSUE #945: Remove "core" from reference headers
This is not necessary bc of the section split in the sidebar and on the reference index page.

signekb:
## Description

This PR prettifies the reference docs a bit by removing the `core.` before each function. This is achieved by adding the package `seedcase_sprout.core" at the subsection level. 

The entire reference will still be shown when looking at the individual function documentation like so: 
image

For reference, this is what it looks like on the website currently: 
image

Related to #927 
Closes #945 

This PR needs a quick review.

## Checklist

- [X] Updated documentation
- [ ] Ran `just run-all`

### DIFF
--- a/_quarto.yml
+++ b/_quarto.yml
@@ -79,6 +79,7 @@ quartodoc:
     table_style: description-list
   sections:
     - title: "Core functions"
+      desc: "Core functions that support the creation and management of data packages and data resources."
 
     - subtitle: "Data package functions"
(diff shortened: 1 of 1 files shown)


lwjohnst86 (approved):
Oh so much nicer!!

PR #948, opened 2025-01-06
docs: 🐛  enable quartodoc dynamic lookup

### LINKED ISSUE #943: Quartodoc shows helper functions we don't want to include in the reference docs
Like so: 

Image

Related to #944

### LINKED ISSUE #944: When referring to a function inside a file of the same name, Quartodoc refers to the file not the function
When referencing a function inside a file with the same name, quartodoc refers - incorrectly - to the file, not the function.

This is why, the reference docs for e.g., `create_package_structure` (in the file `create_package_structure.py`) looks like this: 
Image 

While the reference docs for e.g., `path_package` (in the file `path_package_functions.py`)  looks like this: 

Image

signekb:
## Description

This fixes the issue of quartodoc documenting the module/file instead of the function within the file with the same name. 

By actually loading the package (what they call "dynamic lookup"), I think `core`'s namespace (i.e., `__all__` in `__init__.py`) becomes available for quartodoc. This is instead of quartodoc using static analysis to look up the objects we want to document (which is the default).

It now looks like this for `create_package_structure` locally: 
image
<img https://github.com/user-attachments/assets/fccc6419-ac6e-4b8d-ba56-c6d7dc334a5a)

Instead of: 
image

Closes #944 
Closes #943 

Note: This is a stacked PR on #942 

## Checklist

- [X] Updated documentation
- [ ] Ran `just run-all`

### DIFF
--- a/_quarto.yml
+++ b/_quarto.yml
@@ -74,6 +74,7 @@ quartodoc:
   dir: "docs/reference"
   package: "seedcase_sprout"
   parser: google
+  dynamic: true
   renderer:
     style: markdown
     table_style: description-list
--- a/_quarto.yml
+++ b/_quarto.yml
@@ -79,6 +79,7 @@ quartodoc:
     table_style: description-list
   sections:
     - title: "Core functions"
+      desc: "Core functions that support the creation and management of data packages and data resources."
 
     - subtitle: "Data package functions"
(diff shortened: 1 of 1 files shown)


lwjohnst86 (approved):
:tada:

PR #962, opened 2025-01-10
feat: :sparkles: add simple helper functions

martonvago:
## Description

This PR adds some simple helper functions that will be used when running more complex checks in Sprout.

As a rule, I tried to write simple checks in such a way that they only check a single thing. So if a function checks whether a field is blank, it won't fail if the field is missing or is of the wrong type, because there are separate checks for that. That said, it's always possible to argue that this doesn't make sense for some functions -- so feel free to argue if you think any checks should behave differently!

## Checklist

- [x] Added or updated tests
- [x] Updated documentation
- [x] Ran `just run-all`


seedcase_sprout/core/sprout_checks/check_no_inline_data.py
```diff
@@ -0,0 +1,32 @@
+from seedcase_sprout.core.checks.check_error import CheckError
+from seedcase_sprout.core.sprout_checks.get_json_path_to_resource_field import (
+    get_json_path_to_resource_field,
+)
+
+
+def check_no_inline_data(
+    properties: dict, index: int | None = None
+) -> list[CheckError]:
```
signekb, line 9: What does it return if no error is found?/should that be added here?

tests/core/sprout_checks/test_check_no_inline_data.py
```diff
@@ -0,0 +1,25 @@
...
+
+
+@mark.parametrize("index", [None, 2])
+def test_passes_if_data_not_set(index):
+    """Should pass if inline data is not set."""
+    assert check_no_inline_data({}, index) == []
+
+
+@mark.parametrize("index", [None, 2])
+def test_fails_if_data_is_set(index):
```
signekb, line 16: In general in this PR, there's something with the naming of the tests. Can their names be more specific? 
E.g., it's not like the test below (L16) itself fails, and the function doesn't "fail" either - it returns an error.  Like instead of `test_fails_if_data_is_set()` then `test_returns_error_if_data_is_set()`

signekb (changes requested):
Still loving the smaller PRs 👌 🌟

seedcase_sprout/core/sprout_checks/remove_not_sprout_related_resource_errors.py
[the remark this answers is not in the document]
martonvago: I'm also not a big fan of the name. I wanted it to express what we're removing, not just that we're removing something.
Maybeeee `get_sprout_specific_resource_errors`?

seedcase_sprout/core/sprout_checks/remove_not_sprout_related_resource_errors.py
```diff
@@ -0,0 +1,28 @@
...
+
+    Errors filtered out:
+      - inline `data` required but missing
+      - `path` is not of type array
+
+    Args:
+        errors: The full error list.
+
+    Returns:
+        The filtered error list.
```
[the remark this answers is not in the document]
martonvago: I think it's a bit too long? Especially because it says it in a nicely formatted way a couple of lines above

seedcase_sprout/core/sprout_checks/remove_not_sprout_related_resource_errors.py
```diff
@@ -0,0 +1,28 @@
...
+    return [
+        error
+        for error in errors
+        if not (error.validator == "required" and error.json_path.endswith(".data"))
+        and not (
+            error.validator == "type"
+            and error.json_path.endswith(".path")
+            and error.message.endswith("not of type 'array'")
+        )
+    ]
```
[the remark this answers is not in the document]
martonvago: Hmmm, to be honest, I think it's not too bad the way it is, because it's clearly ANDing two NOT-clauses.

If I were to simplify it, I would be tempted to write a separate function for each check (`is_data_required_error`, `is_path_type_error`), and then do `return [error for error in errors if not is_data_required_error(error) and not is_path_type_error(error)]` in the main function. Because then the work would be more evenly split and it would be more modular.

But these are just my feelings :woman_shrugging:

tests/core/sprout_checks/test_check_no_inline_data.py
martonvago: Sure! (I meant that the check itself fails in a more conceptual way lol)

tests/core/sprout_checks/test_get_json_path_to_resource_field.py
```diff
@@ -0,0 +1,13 @@
+from seedcase_sprout.core.sprout_checks.get_json_path_to_resource_field import (
+    get_json_path_to_resource_field,
+)
+
+
+def test_returns_correct_path_without_index():
```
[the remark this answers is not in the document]
martonvago: I'm happy with "expected"!

seedcase_sprout/core/sprout_checks/remove_not_sprout_related_resource_errors.py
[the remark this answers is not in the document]
martonvago: Actually, that would be too similar to the function that actually collects the errors that are triggered by Sprout-specific requirements. So how about `exclude_non_sprout_resource_errors`?

seedcase_sprout/core/sprout_checks/get_blank_value_for_type.py
```diff
@@ -0,0 +1,19 @@
...
+    Returns:
+        The corresponding blank value.
+    """
+    match type:
+        case RequiredFieldType.str:
+            return ""
+        case RequiredFieldType.list:
+            return []
+        case _:
+            return None
```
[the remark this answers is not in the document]
signekb: Hahahha, fair :p

seedcase_sprout/core/sprout_checks/remove_not_sprout_related_resource_errors.py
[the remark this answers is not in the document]
signekb: Naming is hard. Yes, let's go with `exclude_non_sprout_resource_errors` 👍

seedcase_sprout/core/sprout_checks/remove_not_sprout_related_resource_errors.py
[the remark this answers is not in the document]
signekb: Yeah, that's true. Lets keep it as is 👍

tests/core/sprout_checks/test_check_resource_id_in_data_path.py
```diff
@@ -43,8 +43,8 @@ def test_passes_if_path_of_wrong_type(index, data_path):
         Path("resources", "1", "data.parquet", "1"),
     ],
 )
-def test_fails_if_path_malformed(index, data_path):
-    """Should fail if the path does not contain a resource ID."""
+def test_returns_error_if_data_path_malformed(index, data_path):
```
signekb, line 46:
```suggestion
def test_returns_error_if_data_path_is_malformed(index, data_path):
```

signekb (dismissed):
🌟

seedcase_sprout/core/sprout_checks/check_data_path_string.py
```diff
@@ -0,0 +1,29 @@
...
+    Returns:
+        A list of errors. An empty list if no error was found.
+    """
+    path = properties.get("path", "")
+    if isinstance(path, str):
+        return []
+
+    return [
+        CheckError(
+            message=f"{path} is not of type 'string'",
```
lwjohnst86, line 25:
```suggestion
            message=f"{path} needs to be a 'string'.",
```

seedcase_sprout/core/sprout_checks/check_no_inline_data.py
```diff
@@ -0,0 +1,31 @@
...
+        A list of errors. The empty list if no error was found.
+    """
+    if properties.get("data") is None:
+        return []
+
+    return [
+        CheckError(
+            message=(
+                "'data' should not be set. Sprout expects data in separate "
+                "data files specified by 'path'."
```
lwjohnst86, line 26:
```suggestion
                "Sprout doesn't use the 'data' field, instead it expects data "
                "in separate files that are given in the 'path' field."
```

seedcase_sprout/core/sprout_checks/check_resource_id_in_data_path.py
```diff
@@ -0,0 +1,37 @@
...
+    if not isinstance(data_path, str):
+        return []
+
+    data_path = Path(data_path)
+    if len(data_path.parts) == 3 and data_path.parts[1].isdigit():
+        return []
+
+    return [
+        CheckError(
+            message="'path' should contain the resource ID",
```
lwjohnst86, line 33:
```suggestion
            message="The 'path' field needs to contain the resource ID.",
```

tests/core/sprout_checks/test_check_data_path_string.py
```diff
@@ -0,0 +1,35 @@
...
+
+@mark.parametrize("index", [None, 2])
+def test_error_found_if_path_not_string(index):
+    """Should find an error if the path is not of type string."""
+    properties = {"path": 123}
+
+    errors = check_data_path_string(properties, index)
+
+    assert len(errors) == 1
+    assert errors[0].message == "123 is not of type 'string'"
```
lwjohnst86, line 33: Can you modify these type of assertions find patterns rather than the exact message? Given I made a change in the message above. Or just check for the existence of the message? Seems very brittle to write the exact text used for the message.

tests/core/sprout_checks/test_check_no_inline_data.py
```diff
@@ -0,0 +1,25 @@
...
+
+@mark.parametrize("index", [None, 2])
+def test_error_found_if_data_is_set(index):
+    """Should find an error if inline data is set."""
+    properties = {"data": "some data"}
+
+    errors = check_no_inline_data(properties, index)
+
+    assert len(errors) == 1
+    assert "'data' should not be set" in errors[0].message
```
lwjohnst86, line 23: Same as comment above.

lwjohnst86 (changes requested):
Very nice! Just some small comments and suggestions :grin:

seedcase_sprout/core/sprout_checks/check_data_path_string.py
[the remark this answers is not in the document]
martonvago: Yeah, that's a good point!

seedcase_sprout/core/sprout_checks/check_resource_id_in_data_path.py
```diff
@@ -0,0 +1,37 @@
+from pathlib import Path
+
+from seedcase_sprout.core.checks.check_error import CheckError
+from seedcase_sprout.core.sprout_checks.get_json_path_to_resource_field import (
+    get_json_path_to_resource_field,
+)
+
+
+def check_resource_id_in_data_path(
```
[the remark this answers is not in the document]
martonvago: Or, with the above change, `check_resource_id_in_resource_path`? Could even shorten it to `check_id_in_resource_path`?

seedcase_sprout/core/sprout_checks/check_data_path_string.py
martonvago: I would agree, except that I want to match the exact phrasing of `jsonschema`'s `ValidationError`s, because I want type errors for the same field and the same wrong type from `jsonschema` and from Sprout to look exactly the same. Because then it's easy to de-duplicate them.

tests/core/sprout_checks/test_check_data_path_string.py
martonvago: I'll have a look at the rest, but for this one I wanted to check that it matches the phrasing of `jsonschema` exactly (see comment above). We could maybe put that phrasing in a constant and reference it like that to make it obvious?

seedcase_sprout/core/sprout_checks/check_data_path_string.py
lwjohnst86: hmmm, gotcha. Could we change them in the final output? These error messages are not the most helpful (error messages never seem to be!).

lwjohnst86: Could you also leave a comment there to indicate it needs to match the `jsonschema`? So that we don't modify it later.

tests/core/sprout_checks/test_check_data_path_string.py
lwjohnst86: I think that sounds good to me :+1: it will be less brittle that way :relaxed:

seedcase_sprout/core/sprout_checks/check_resource_id_in_data_path.py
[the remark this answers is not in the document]
lwjohnst86: I like the shorter one!

seedcase_sprout/core/sprout_checks/check_data_path_string.py
martonvago: Yeah, we could change them at 2 points!
1. In the `checks` package. So any check functions from the `checks` package would already return errors with this message. 
2. In Sprout. Change them either directly after calling a check function from the `checks` package or at the very end.

I think 1 would be a lot cleaner!

tests/core/sprout_checks/test_check_data_path_string.py
martonvago: Okay, I'll do that. And then we can change the error messages exposed by `checks` (and copied by Sprout) in another PR.

seedcase_sprout/core/sprout_checks/check_data_path_string.py
lwjohnst86: Yea, I agree, 1 is cleaner.

seedcase_sprout/core/sprout_checks/check_resource_path_string.py
```diff
@@ -3,8 +3,10 @@
     get_json_path_to_resource_field,
 )
 
+CHECKS_TYPE_ERROR_MESSAGE = "{field_value} is not of type '{field_type}'"
```
martonvago, line 6: This could arguably go into a separate file or even be exposed by the `checks` package. As #978 will rework this, I'll just leave it like this for now.

lwjohnst86 (approved):
:+1:

PR #964, opened 2025-01-10
feat: :sparkles: collect all Sprout-specific property errors

martonvago:
## Description

This PR adds functions that collect Sprout-specific errors (i.e. violations of constraints other than those in the Data Package standard) after checking package and resource properties. These functions will be used together with check functions from the `checks` package to run all checks.

## Checklist

- [x] Added or updated tests
- [x] Updated documentation
- [x] Ran `just run-all`


tests/core/sprout_checks/test_get_sprout_resource_errors.py
```diff
@@ -0,0 +1,123 @@
...
+        title="Resource 1",
+        description="A resource.",
+    ).compact_dict
+
+
+@fixture
+def properties_partial():
+    return ResourceProperties(
+        path=str(Path("resources", "1", "data.parquet")),
+    ).compact_dict
```
signekb, line 34: Following #970 , should `get_sprout_resource_errors()` also take a `Properties` class instead of a `dict`, or do we at this stage want to keep using `dicts`?

seedcase_sprout/core/sprout_checks/get_sprout_package_errors.py
```diff
@@ -0,0 +1,29 @@
...
+from seedcase_sprout.core.sprout_checks.check_required_package_properties_not_blank import (  # noqa: E501
+    check_required_package_properties_not_blank,
+)
+from seedcase_sprout.core.sprout_checks.required_fields import (
+    PACKAGE_SPROUT_REQUIRED_FIELDS,
+)
+
+
+def get_sprout_package_errors(
+    properties: dict, check_required: bool
```
signekb, line 14: At this point, I'm not entirely sure why we need the `check_required` arg? Is that for after `create_package_structure()` that currently creates the properties without some of the required fields?

tests/core/sprout_checks/test_get_sprout_resource_errors.py
```diff
@@ -0,0 +1,123 @@
...
+
+
+# Check required fields
+
+
+def test_passes_full_resource_properties(properties):
+    """Should pass with a full set of resource properties."""
```
signekb, line 41: The same with test naming here as in one of the previous PRs #962 . Would you mind updating these as well? :)

signekb (changes requested):
Nice work ⭐ Looking forward to getting to the next PRs.

seedcase_sprout/core/sprout_checks/get_sprout_package_errors.py
martonvago: Partly yes! 
Another scenario is when we edit a set of properties: the incoming set of properties probably won't be a full set of properties because it will only contain the field or fields we want to update. But we still want to be able to check that these properties are otherwise correct/well-formed.

tests/core/sprout_checks/test_get_sprout_resource_errors.py
martonvago: Yes! But I thought maybe let's allow this current set of PRs (1-7) to go through with everything being `dict` and then make a new one that moves them over to `Properties`. Because then we just decide where we do the `Properties` > `dict` conversion and switch to `Properties` in every function above that point.

tests/core/sprout_checks/test_get_sprout_resource_errors.py
martonvago: Will do!

seedcase_sprout/core/sprout_checks/get_sprout_package_errors.py
signekb: Ah, ofc! Makes sense 👍 
An alternative scenario would be if we passed along the entire properties with the edited field(s)? That an edit properties function would 1) load the existing properties, 2) edit the field(s), 3) check that the entirety of the properties are well-formed?

martonvago: Yes and I think our update function has that check as well. So currently we actually do 3 checks when we update: check the existing properties (may be incomplete), check the incoming properties (may be incomplete), check the result of the update (must be complete).
(((I have a vague memory about discussing if we should keep all 3 checks and deciding yes)))

tests/core/sprout_checks/test_get_sprout_resource_errors.py
martonvago: I actually changed my mind, sorry! I put a [comment](https://github.com/seedcase-project/seedcase-sprout/issues/970#issuecomment-2593080327) on that issue about it.

seedcase_sprout/core/sprout_checks/get_sprout_package_errors.py
signekb: I also have a vague memory of this, actually. I guess it's to differentiate whether the check fails due to problems in the existing properties, in the edited field, or in the merge of the two?

signekb: @martonvago based on what you said at the status meeting today, will you change the `check_required` in another PR or add that change throughout the existing PRs?

martonvago: @signekb New PR, less messy that way I think

signekb: Super. I'll just link to it here for clarity #972

lwjohnst86 (dismissed):
:+1:

lwjohnst86:
I'll merge this after the PR this is stacked on gets resolved.

PR #967, opened 2025-01-10
feat: :sparkles: add `check_properties()`

martonvago:
## Description

This PR adds the top-level check function for package and resource properties combined (in Sprout).

## Checklist

- [x] Added or updated tests
- [x] Updated documentation
- [x] Ran `just run-all`


tests/core/sprout_checks/test_check_properties_sprout.py
```diff
@@ -0,0 +1,160 @@
...
+                description="A second resource.",
+            ),
+        ],
+    ).compact_dict
+
+
+@mark.parametrize("check_required", [True, False])
+def test_passes_full_properties(properties, check_required):
+    """Should pass if all required fields are present and correct."""
+    assert check_properties(properties, check_required=check_required) == properties
```
signekb, line 50: Naming of tests :)) 

Could also be something like:

```suggestion
def test_check_passes_full_properties(properties, check_required):
    """Should pass if all required fields are present and correct."""
    assert check_properties(properties, check_required=check_required) == properties
```

signekb (changes requested):
Nice work! 😁

tests/core/sprout_checks/test_check_properties_sprout.py
martonvago: I realllllly don't mind, but there's an argument that "check" is implied in all test names because it's the only function we're testing.

seedcase_sprout/core/sprout_checks/check_properties.py
```diff
@@ -0,0 +1,48 @@
...
+    errors += get_sprout_package_errors(properties, check_required)
+
+    for index, resource in enumerate(properties.get("resources", [])):
+        errors += get_sprout_resource_errors(resource, check_required, index)
+
+    errors = sorted(set(errors))
+
+    if errors:
+        raise ExceptionGroup(
+            f"Properties check failed on properties\n{properties}", errors
```
lwjohnst86, line 45:
```suggestion
            f"The follow checks failed on the properties:\n{properties}", errors
```

lwjohnst86 (changes requested):
Like with other PRs, only a small suggestion

lwjohnst86 (dismissed):
Nice to see it all tied together at the end!

seedcase_sprout/core/sprout_checks/check_package_properties.py
```diff
@@ -21,10 +20,10 @@ def check_package_properties(properties: dict, check_required=True) -> dict:
             fields. Defaults to True.
 
     Returns:
-        `properties`, if all checks passed.
+        `properties` if all checks pass.
 
     Raises:
-        FailedCheckError: If at least one check failed.
+        ExceptionGroup: A group of `CheckError`s, one for each check that failed.
```
signekb, line 26:
```suggestion
        ExceptionGroup: A group of `CheckError`s, one error per failed check.
```
Same for other docstrings with this error :)

signekb (changes requested):
Very nice! ⭐  Just a minor thing (+ a workflow failed)

PR #968, opened 2025-01-13
refactor: :recycle: use `check_package_properties`

martonvago:
## Description

This PR refactors `edit_package_properties` to use `check_package_properties`.

## Checklist

- [x] Added or updated tests
- [x] Updated documentation
- [x] Ran `just run-all`


seedcase_sprout/core/edit_package_properties.py
```diff
@@ -43,16 +38,18 @@ def edit_package_properties(path: Path, properties: dict) -> dict:
...
         FileNotFound: If the `datapackage.json` file doesn't exist.
-        NotPropertiesError: If the new package properties are not correct or the current
-            package properties are not well-formed.
+        FailedCheckError: If there is an error in the current, incoming or resulting
+            package properties.
         JSONDecodeError: If the `datapackage.json` file couldn't be read.
     """
     check_is_file(path)
-    verify_properties_are_well_formed(properties, PackageError.type)
+    properties.pop("resources", None)
```
martonvago, line 46: This just removes `resources` from the incoming properties object. I added this because this function is about updating package properties, so we don't want to overwrite or update resource properties.
Maybe it can be argued that it would make sense to allow setting `resources` to `[]`, for example, but that could also be achieved by deleting all resources.

But I'm open to doing it in a different way!

seedcase_sprout/core/edit_package_properties.py
```diff
@@ -43,16 +38,18 @@ def edit_package_properties(path: Path, properties: dict) -> dict:
...
+
+    check_package_properties(properties, check_required=False)
```
martonvago, line 48: No required check because incoming properties are likely to be partial properties, e.g. just updating one field.

seedcase_sprout/core/edit_package_properties.py
```diff
@@ -43,16 +38,18 @@ def edit_package_properties(path: Path, properties: dict) -> dict:
...
 
     current_properties = read_json(path)
-    verify_properties_are_well_formed(current_properties, PackageError.type)
+    check_package_properties(current_properties, check_required=False)
```
martonvago, line 51: No required check because current properties may be incomplete. E.g., when first created, package properties will not have all fields and it is in this step that missing fields are populated.

seedcase_sprout/core/edit_package_properties.py
```diff
@@ -43,16 +38,18 @@ def edit_package_properties(path: Path, properties: dict) -> dict:
...
 
     current_properties.update(properties)
 
-    return verify_package_properties(current_properties)
+    return check_package_properties(current_properties, check_required=True)
```
martonvago, line 55: Required check done because at this point all required fields should be present.

Note though that this check is only for package properties. So if `resources` are set (because they were on the original properties) these will not be checked here. Should we check them? We could have a check that checks resources if they are present, but doesn't fail if `resources` are missing.

tests/core/test_edit_package_properties.py
```diff
@@ -64,41 +119,75 @@ def test_throws_error_if_path_points_to_nonexistent_file(tmp_path):
...
 
+    assert "a name with spaces" in error.value.message
+    errors = error.value.errors
+    assert len(errors) == 1
+    assert errors[0].json_path == "$.name"
+    assert errors[0].validator == "pattern"
 
-def test_throws_error_if_new_properties_are_empty(properties_path):
-    """Should throw NotPropertiesError if the new properties are empty."""
-    with raises(NotPropertiesError):
```
martonvago, line 103: I thought it was more consistent to accept `{}` for incoming properties. These pass `check_package_properties` when required fields are not checked, and having a "neutral element" for the edit operation seemed okay.

martonvago:
Opened https://github.com/seedcase-project/seedcase-sprout/pull/974 instead

PR #971, opened 2025-01-16
docs: ✨ align table colwidths on reference index page

### LINKED ISSUE #946: Align column widths in the function overview tables to avoid splitting function names
It would be nice if the function overview table had aligned and more fitting col widths, so function names aren't split to two lines like they are currently:

 Image

Can be achieved by adding a yml header to `reference/index.qmd` like below: 
```reference/index.qmd
---
tbl-colwidths: [40, 60]
---
```

I just need to figure out if this is possible within Quartodocs or some kind of workaround is needed.

EDIT: Might be possible with https://machow.github.io/quartodoc/get-started/extending.html#using-a-custom-index-page

signekb:
## Description

This PR customises the `summarize` function for `layout.Section`s by adding column widths to the tables.
This change only affects the tables created on the reference index page. 

Full disclosure: I don't really like the way the `render` and `summarize` function(s) are set up in `quartodoc`, but I kept it as close to the original as possible (basically only adding the `+ ': {tbl-colwidths="[40,60]"}'` in L70 here). 

An alternative solution would be to change the output directory of quartodoc to e.g., `out_index: docs/reference/index-quartodoc.qmd` and then add a [custom index page](https://machow.github.io/quartodoc/get-started/extending.html#using-a-custom-index-page) `docs/reference/index.qmd` that looks something like this: 

```
---
tbl-colwidths: [40,60]
---

{{< include /docs/reference/index-quartodoc.qmd >}}

```

## Effect of this PR

| Before/current | After |
| ------------- | ------------- |
|    |  |

Closes #946 

## Checklist

- [X] Updated documentation
- [X] Ran `just run-all`


lwjohnst86 (approved):
Yea, I don't love how quartodoc is internally implemented, but this looks fine with me :grin: we aren't building quartodoc, and it isn't high priority to make elegant, well designed code for it. Just whatever works :stuck_out_tongue:

PR #972, opened 2025-01-16
feat: :sparkles: add `CheckErrorMatcher` object

martonvago:
## Description

This PR adds a `CheckErrorMatcher` object. This is/will be used to filter out unwanted errors easily.
Example usage:
```python
errors = exclude_errors(
        errors,
        [
            CheckErrorMatcher(validator="required", json_path="data"),
            CheckErrorMatcher(validator="type", json_path="path", message="not of type 'array'"),
        ],
    )
```
I'm going to use this to replace the `check_required` flag in Sprout's check functions.

## Checklist

- [x] Added or updated tests
- [x] Updated documentation
- [x] Ran `just run-all`


seedcase_sprout/core/checks/check_error_matcher.py
martonvago, line 1: This is a very simple class for encapsulating matching logic. The motivation behind this is that filtering `CheckError`s or deciding if a `CheckError` fulfills any of a set of conditions comes up pretty often.

martonvago: I put this class and `exclude_errors.py` into `checks` because they are generally applicable, but we can move them to `sprout_checks` if we think they should be restricted to Sprout.

seedcase_sprout/core/checks/check_error_matcher.py
```diff
@@ -0,0 +1,66 @@
...
+
+        Returns:
+            If there was a match.
+        """
+        return self.message is None or self.message in error.message
+
+    def json_path_matches(self, error: CheckError) -> bool:
+        """Determines if this matcher matches the `json_path` of the given `CheckError`.
+
+        Matching on the full `json_path` and matching on the field name are supported.
```
martonvago, line 43: This can obviously be refined further (e.g. to allow matching fields with a specific name on resources only), but we can add that as/if we need it. Same goes for `message_matches`.

lwjohnst86 (dismissed):
I have a hard time seeing what this does or it's usefulness, but I think it would help once we move these functionality over into a standalone package with some guides showing the functions in action. Or seeing the functions in action in other functions within Sprout. But you're very much knowledgeable about what's needed and the challenges in this check area, so :+1:

signekb (commented):
Nice work 👍  Just a suggestion and a question to help me understand 🔍

tests/core/checks/test_check_error_matcher.py
```diff
@@ -0,0 +1,99 @@
...
+        )
+        is expected
+    )
+
+
+@mark.parametrize(
+    "json_path,matcher,expected",
+    [
+        ("$.match", None, True),
+        ("$.no.match", "", False),
```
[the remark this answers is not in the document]
martonvago: Not very solid reasons :sweat_smile:! I was thinking:
- `message_matches` should check if the matcher message is a substring of the error message. Doesn't matter that much if `""` matches or not?
- `json_path_matches` should check if the matcher json path is exactly the full path or exactly the very last field. `""` is neither, so I made it not match anything.
- `validator_matches` should check if the matcher validator is exactly the validator of the error. `""` is not a validator, so I made it not match.

We can refine the behaviour further for sure!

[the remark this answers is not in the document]
signekb: Ah, I think that makes sense! Thanks for elaborating 👍

PR #973, opened 2025-01-17
refactor: :recycle: use `CheckErrorMatcher` in Sprout checks

martonvago:
## Description

This PR replaces the `is_required` flag in Sprout's check functions with a list of `CheckErrorMatcher`s to ignore.
Many files were changed, but the changes are very similar and don't affect the behaviour.

## Checklist

- [x] Added or updated tests
- [x] Updated documentation
- [x] Ran `just run-all`


lwjohnst86 (dismissed):
Nice streamlining of some things!

seedcase_sprout/core/sprout_checks/check_properties.py
```diff
@@ -10,16 +9,15 @@
 )
 
 
-def check_properties(properties: dict, check_required=True) -> dict:
+def check_properties(properties: dict, ignore: list[CheckErrorMatcher] = []) -> dict:
```
signekb, line 12: I might have asked this before, but how come `check_properties()` doesn't use `check_package_properties()` and `check_resource_properties()`? :)

Just saw that some of the logic is duplicated (L34-43 in his file and L37-46 in `check_resource_properties`

signekb (commented):
Nice 👍 Just a question

seedcase_sprout/core/sprout_checks/check_properties.py
martonvago: Yeah, they are very similar!

It's mirroring the structure of the `checks` package. There, the initial problem with using `check_package_properties` and `check_resource_properties` in `check_properties` was that
1.  `check_package_properties` deliberately removes some constraints about the `resources` field, but in `check_properties` we want these, so we would need to add them right back.
2. `check_resource_properties` knows only about the resource properties it's given, so it creates `json_path`s to errors relative to this root. In `check_properties` we want the root to be the package properties.

That said, based on how we use these, I think we will be able to simplify / commonise things somewhat!

PR #974, opened 2025-01-17
refactor: :recycle: use check functions in `edit_package_properties`

martonvago:
## Description

This PR refactors `edit_package_properties` to use the check functions.

Once https://github.com/seedcase-project/seedcase-sprout/pull/976 is merged in, I will update this PR with those changes. <-- Done! :heavy_check_mark: 

## Checklist

- [x] Added or updated tests
- [x] Updated documentation
- [x] Ran `just run-all`


seedcase_sprout/core/edit_package_properties.py
```diff
@@ -43,16 +40,27 @@ def edit_package_properties(path: Path, properties: dict) -> dict:
...
-            package properties are not well-formed.
         JSONDecodeError: If the `datapackage.json` file couldn't be read.
+        ExceptionGroup: If there is an error in the current, incoming or resulting
+            package properties. A group of `CheckError`s, one error for each failed
+            check.
     """
     check_is_file(path)
-    verify_properties_are_well_formed(properties, PackageError.type)
+
+    properties.pop("resources", None)
```
martonvago, line 50: This just removes `resources` from the incoming properties object. I added this because this function is about updating package properties, so we don't want to overwrite or update resource properties.
Maybe it can be argued that it would make sense to allow setting `resources` to `[]`, for example, but that could also be achieved by deleting all resources.

But I'm open to doing it in a different way! E.g., we could throw an error if `resources` are set.

seedcase_sprout/core/edit_package_properties.py
```diff
@@ -43,16 +40,27 @@ def edit_package_properties(path: Path, properties: dict) -> dict:
...
+    check_package_properties(
+        properties, ignore=[CheckErrorMatcher(validator="required")]
+    )
```
martonvago, line 53: No required check because incoming properties are likely to be partial properties, e.g. just updating one field.

seedcase_sprout/core/edit_package_properties.py
```diff
@@ -43,16 +40,27 @@ def edit_package_properties(path: Path, properties: dict) -> dict:
...
 
     current_properties = read_json(path)
-    verify_properties_are_well_formed(current_properties, PackageError.type)
+    check_properties(
+        current_properties,
+        ignore=[CheckErrorMatcher(validator="required")],
+    )
```
martonvago, line 59: No required check because current properties may be incomplete. E.g., when first created, package properties will not have all fields and it is in this step that missing fields are populated.

I decided to do a full check of both package and resource properties here (`check_properties` and not just `check_package_properties`) because the current properties may have resources, and maybe we shouldn't allow editing properties if there's anything invalid/incorrect  about them. But let me know if we should not check resource properties here!

seedcase_sprout/core/edit_package_properties.py
```diff
@@ -43,16 +40,27 @@ def edit_package_properties(path: Path, properties: dict) -> dict:
...
 
     current_properties.update(properties)
 
-    return verify_package_properties(current_properties)
+    return check_properties(
+        current_properties,
+        ignore=[CheckErrorMatcher(validator="required", json_path="resources")],
+    )
```
martonvago, line 66: Required check done because at this point all required fields should be present -- that is, all required fields except resources.

However, as above, resources may be present (if they were on the original properties), so I'm using `check_properties` for a full check.

tests/core/test_edit_package_properties.py
```diff
@@ -64,41 +118,102 @@ def test_throws_error_if_path_points_to_nonexistent_file(tmp_path):
...
-        == current_properties | new_properties
-    )
 
+def test_throws_error_if_both_current_and_incoming_properties_empty(tmp_path):
+    """Should throw a group of `CheckError`s if both current and incoming properties are
+    empty."""
+    properties_path = get_properties_path(tmp_path, {})
 
-def test_throws_error_if_new_properties_are_empty(properties_path):
-    """Should throw NotPropertiesError if the new properties are empty."""
```
martonvago, line 102: I thought it was more consistent to accept `{}` for incoming properties. These pass `check_package_properties` when required fields are not checked, and having a "neutral element" for the edit operation seemed okay.

seedcase_sprout/core/edit_package_properties.py
```diff
@@ -46,20 +43,31 @@ def edit_package_properties(
...
+        ignore=[CheckErrorMatcher(validator="required")],
+    )
 
     current_properties.update(properties)
 
-    verify_package_properties(current_properties)
+    check_properties(
+        current_properties,
+        ignore=[CheckErrorMatcher(validator="required", json_path="resources")],
+    )
```
martonvago, line 71: Required check done because at this point all required fields should be present -- that is, all required fields except resources.

However, as above, resources may be present (if they were on the original properties), so I'm using `check_properties` for a full check.

lwjohnst86 (dismissed):
Lovellyyy :star_struck:

PR #976, opened 2025-01-20
refactor: ♻️  user `Properties` classes in user facing functions

### LINKED ISSUE #970: Use `*Properties` classes as argument types in Sprout
[This only affects Sprout, not the `checks` package.]

Change all argument types for properties arguments from `dict` to the approperiate `*Properties` class.
Starting at top-level functions, we should change to using `*Properties` classes and use them as far down in the function stack as possible.

A natural point for transitioning to `dict` (so calling `compact_dict` on the properties) would be when the properties are fed to a check function or the `write_json` function.

signekb:
## Description

This PR changes the input type (and output type, when it's relevant) to a `Properties` class in the user facing functions.

This means that the functions take a `Properties` class as input, but transforms it to a `dict` before handling it to the `checks`. Before returning, the `properties` object is transformed back to a `Properties` class using a new method `from_dict` based on the work by @martonvago. For this method, we needed a new dependency (`dacite`).

Closes #970 

## Checklist

- [X] Added or updated tests
- [X] Updated documentation
- [X] Ran `just run-all`


tests/core/test_create_resource_properties.py
```diff
@@ -1,40 +1,44 @@
...
+
+
+def test_creates_properties_correctly(tmp_path, resource_properties):
     """Given valid inputs, should create properties object correctly."""
     resource_path = tmp_path / "resources" / "1"
     resource_path.mkdir(parents=True)
-    properties = {
-        "name": "test",
-        "path": "",
-    }
```
signekb, line 16: Added a fixture, `resource_properties()`, instead of defining identical `properties` here and in L28-31.

tests/core/test_edit_package_properties.py
```diff
@@ -48,20 +48,22 @@ def test_edits_only_changed_package_properties(properties_path, properties):
     current_properties = read_json(properties_path)
 
     # When, Then
-    expected_properties = current_properties | properties
-    assert edit_package_properties(properties_path, properties) == expected_properties
+    expected_properties = current_properties | properties.compact_dict
```
signekb, line 51: The easiest way to "merge" the properties are still by using the `dict` representation. So, that's why I compare `dicts` here.

tests/core/test_edit_package_properties.py
```diff
@@ -83,22 +85,19 @@ def test_throws_error_if_current_package_properties_are_malformed(tmp_path, prop
...
-    new_properties = {"custom-field": "custom-value"}
+    new_properties = PackageProperties.from_dict({"custom-field": "custom-value"})
 
     # When, Then
-    assert (
-        edit_package_properties(properties_path, new_properties)
-        == current_properties | new_properties
-    )
+    with raises(NotPropertiesError):
+        assert edit_package_properties(properties_path, new_properties)
```
signekb, line 97: Note : This is a change of behaviour. 

When we used a dict, it was easy to add a custom field (as allowed by the Data Package standard). This, however, becomes tricker when we use the `Properties` classes. 

Are we fine with not supporting custom fields for now and then add it later, if we ever need/want to support them?

tests/core/test_properties.py
```diff
@@ -99,3 +99,41 @@ def test_creates_package_properties_with_correct_defaults(mock_uuid):
     assert properties.id == str(mock_uuid())
     assert properties.version == "0.1.0"
     assert properties.created == "2024-05-14T05:00:01+00:00"
+
+
+@mark.parametrize(
+    "dict, expected_properties",
```
signekb, line 105: Is there a more elegant way of testing this method for all `Properties` classes or is this fine?

tests/core/test_edit_package_properties.py
lwjohnst86: I'm fine with that. If we need to add custom fields, I think there's a section in the data package spec that talks about adding extensions, which we could do.

tests/core/test_properties.py
lwjohnst86: Not sure, but my feeling is no :laughing: This is fine though, it is just test code.

lwjohnst86 (dismissed):
Nice! Looks good :relaxed:

tests/core/test_edit_package_properties.py
```diff
@@ -83,22 +85,19 @@ def test_throws_error_if_current_package_properties_are_malformed(tmp_path, prop
         edit_package_properties(path, properties)
 
 
-def test_adds_custom_fields(
+def test_throws_error_when_adding_custom_fields(
```
martonvago, line 88: My only problem with this test is that what this is actually testing is that an empty properties object will lead to an error (so, the next test). This is because `from_dict` ignores unknown key-value pairs, so here it ignores all key-value pairs. But if we added a known key-value pair, `from_dict` would create an acceptable `PackageProperties` and no error would be raised.

Now that the input to the function is a `PackageProperties` (which cannot contain custom fields), I think we can just remove this test.

tests/core/test_properties.py
martonvago: I also think it's fine. I think the nested tests are very useful, the others are basically testing that all subclasses inherit the superclass method.

martonvago (changes requested):
Yay! :rocket: 

I will use this as a base for https://github.com/seedcase-project/seedcase-sprout/pull/974

tests/core/test_edit_package_properties.py
signekb: You're absolutely right! :star2: I'll remove it.

lwjohnst86: @signekb when you do this, I can approve and merge :relaxed:

signekb: Done 🔥

martonvago (dismissed):

Issue #980, opened 2025-01-21
Remove `path_package_database()`


signekb:
Since with the current setup, we wont create a db.

PR #982, opened 2025-01-21
docs: 📝  add examples to user facing functions

### LINKED ISSUE #951: Add examples to docstrings with Quarto syntax

signekb:
## Description

This PR adds examples in the docstrings of user facing functions. 

Closes #951 

## Checklist

- [X] Updated documentation
- [X] Ran `just run-all`


seedcase_sprout/core/path_package_functions.py
```diff
@@ -28,6 +41,19 @@ def path_package_database(package_id: int) -> Path:
...
+        ```{python}
+        #| output: true
+        import os
+
+        import seedcase_sprout.core as sp
+
+        # Set global path for the example
+        os.environ["SPROUT_GLOBAL"] = ".storage/"
+
+        # sp.path_package_database(package_id=1)
```
signekb, line 55: Commented out bc there is no package database in the current setup. This function will soon be removed altogether #980

seedcase_sprout/core/write_resource_properties.py
```diff
@@ -31,6 +31,50 @@ def write_resource_properties(
...
+            path=sp.path_package_properties(package_id=1),
+            properties=sp.PackageProperties(
+                name="new-package-name",
+                title="New Package Title",
+                description="This is a package description",
+            ),
+        ).compact_dict
+
+        write_json(edited_properties, path=sp.path_package_properties(package_id=1))
+        ```
```
signekb, line 57: Had to add this bc `write_resource_properties()` would fail otherwise bc the existing properties/`data-package.json` are missing required fields (name, description, title)

seedcase_sprout/core/path_resource_functions.py
```diff
@@ -29,6 +42,25 @@ def path_resource_data(package_id: int, resource_id: int) -> Path:
 
     Returns:
         The absolute path the specified resource's data file.
+
+    Examples:
+        ```{python}
+        #| output: false
+        #| echo: false
+        f = open(".storage/packages/1/resources/1/data.parquet", "x")
+        ```
```
signekb, line 51: Added this so there was a data file to get the path for.

seedcase_sprout/core/create_package_structure.py
```diff
@@ -32,6 +32,19 @@ def create_package_structure(path: Path) -> list[Path]:
         NotADirectoryError: If the directory in the path doesn't exist.
         FileNotFoundError: If a file could not be created.
         TypeError: If `datapackage.json` could not be created.
+
+    Examples:
+        ```{python}
+        #| output: true
+        import os
+
+        import seedcase_sprout.core as sp
```
lwjohnst86, line 41: We really have to import each time???

seedcase_sprout/core/create_resource_properties.py
```diff
@@ -39,6 +39,25 @@ def create_resource_properties(
...
+        # Set global path for the example
+        os.environ["SPROUT_GLOBAL"] = ".storage/"
+
+        sp.create_resource_properties(
+            path=sp.path_resource(package_id=1, resource_id=1),
+            properties=sp.ResourceProperties(
+                name="new-resource-name",
+                path="data.parquet"
+            ),
+        )
```
lwjohnst86, line 59: Examples should be self-contained. This example would need to have the properties created first.
```suggestion
				# Need to create a property object first
				sp.create_package_structure(path=sp.path_packages())
				
				# Now can add a resource property
        sp.create_resource_properties(
            path=sp.path_resource(package_id=1, resource_id=1),
            properties=sp.ResourceProperties(
                name="new-resource-name",
                path="data.parquet"
            ),
        )
```

seedcase_sprout/core/create_resource_structure.py
```diff
@@ -28,6 +28,19 @@ def create_resource_structure(path: Path) -> list[Path]:
...
+         ```{python}
+         #| output: true
+         import os
+
+         import seedcase_sprout.core as sp
+
+         # Set global path for the example
+         os.environ["SPROUT_GLOBAL"] = ".storage/"
+
+         sp.create_resource_structure(path=sp.path_resources(package_id=1))
```
lwjohnst86, line 42: See above.

_quarto.yml
```diff
@@ -2,6 +2,10 @@ project:
   type: seedcase-theme
   pre-render: rm -rf .storage
   render:
+    - "docs/reference/create_package*.qmd"
+    # create_resource_structure must run after create_package_structure
+    # but before create_resource_properties for examples to run correctly
+    - "docs/reference/create_resource_structure.qmd"
```
lwjohnst86, line 8: A well designed example code shouldn't depend on other example codes from running, so these settings actually highlight something we need to consider/be aware of. And on how to write our example code.

lwjohnst86 (changes requested):
Nice start! I didn't finish reviewing it since the main thing to resolve here is that example code should be self-contained. It's a good exercise for us to identify what functions depend on others to run without anything else involved.

_quarto.yml
signekb: @lwjohnst86 So, you would like the example for e.g., `create_resource_structure` to include the creation of a package?

seedcase_sprout/core/create_package_structure.py
signekb: I think so? 
Bc when Quarto renders, it starts a new Python kernel per document? 
image

I also think it's not too bad to be very explicit and have the import in the examples?

_quarto.yml
signekb: Based on the comment further down, that's a yes!

seedcase_sprout/core/path_resource_functions.py
```diff
@@ -29,6 +49,36 @@ def path_resource_data(package_id: int, resource_id: int) -> Path:
...
+            #   package_id=1,
+            #   resource_id=1,
+            #   data="path/to/data.csv")
+
+            # sp.write_resource_parquet(
+            #     raw_files=sp.path_resource_raw_files(package_id=1, resource_id=1),
+            #     path=sp.path_resource_data(package_id=1, resource_id=1))
+
+            # Get the path to the resource data
+            # sp.path_resource_data(package_id=1, resource_id=1)
```
signekb, line 80: Some examples, such as this one, requires functionality that doesn't exist yet, so I have commented it out and added a TODO item.

seedcase_sprout/core/write_resource_properties.py
```diff
@@ -31,6 +31,44 @@ def write_resource_properties(
...
+        #     path=temp_dir / "1" / "datapackage.json",
+        #     resource_properties=sp.ResourceProperties(
+        #         name="new-resource-name",
+        #         title="New resource name",
+        #         description="This is a new resource",
+        #         path="data.parquet",
+        #     ),
+        # )
+        ```
     """
```
signekb, line 72: Same here. Currently, we can't write resource properties on a newly created package bc the (package) properties are missing required fields.

lwjohnst86 (approved):
Super nice!!! :star_struck:

PR #984, opened 2025-01-21
feat: :sparkles: add `write_package_properties()`

martonvago:
## Description

This PR adds the `write_package_properties` function. This will be used to complete the package guide.

## Checklist

- [x] Added or updated tests
- [x] Updated documentation
- [x] Ran `just run-all`


seedcase_sprout/core/write_package_properties.py
```diff
@@ -0,0 +1,34 @@
...
+from seedcase_sprout.core.checks.check_error_matcher import CheckErrorMatcher
+from seedcase_sprout.core.properties import PackageProperties
+from seedcase_sprout.core.sprout_checks.check_properties import check_properties
+from seedcase_sprout.core.write_json import write_json
+
+
+def write_package_properties(path: Path, properties: PackageProperties) -> Path:
+    """Writes the specified package properties to the `datapackage.json` file.
+
+    Writes a set of package properties (with or without resource properties) back to the
```
martonvago, line 12: As we're rewriting the entire file/object, it made sense for me to allow resource properties to be present but to not require them necessarily. Maybe this (and all our recent changes) will allow us to simplify `write_resource_properties`.

tests/core/test_write_package_properties.py
```diff
@@ -0,0 +1,113 @@
...
+    write_json(old_properties, path)
+
+    assert write_package_properties(path, package_properties) == path
+    assert_file_contains(path, package_properties)
+
+
+def test_rewrites_resource_properties_when_file_exists(path, package_properties):
+    """Should write properties to file when the file already exists, updating existing
+    resource properties."""
+    old_properties = replace(
```
martonvago, line 62: Just found this function, pretty useful!! :auto_rickshaw:

tests/core/test_write_package_properties.py
```diff
@@ -0,0 +1,113 @@
...
+
+    assert write_package_properties(path, new_properties) == path
+    assert_file_contains(path, new_properties)
+
+
+def test_throws_error_if_error_in_package_properties(path, package_properties):
+    """Should throw `CheckError`s if there are errors in the package properties."""
+    package_properties.name = "invalid name with spaces"
+    package_properties.id = None
+    write_json(resource_properties.compact_dict, path)
```
martonvago, line 85: Just to test that the file is not rewritten when the check fails.

seedcase_sprout/core/write_package_properties.py
```diff
@@ -0,0 +1,34 @@
...
+
+    Raises:
+        ExceptionGroup: If there is an error in the properties. A group of
+            `CheckError`s, one error for each failed check.
+    """
+    properties = properties.compact_dict
+    check_properties(
+        properties,
+        ignore=[CheckErrorMatcher(validator="required", json_path="resources")],
+    )
```
lwjohnst86, line 33: Very neat!

seedcase_sprout/core/write_package_properties.py
lwjohnst86: Yea, I'm currently going through the design interface data package section. Does it make sense that this function overwrites the file? It keeps it simple. But how might we keep it from overwriting the resources if there are any? In those situations where they read -> edit -> write.

seedcase_sprout/core/write_package_properties.py
```diff
@@ -0,0 +1,34 @@
...
+
+
+def write_package_properties(path: Path, properties: PackageProperties) -> Path:
+    """Writes the specified package properties to the `datapackage.json` file.
+
+    Writes a set of package properties (with or without resource properties) back to the
+    `datapackage.json` file. The `path` argument is the location of the
+    `datapackage.json` file. Use the `path_properties()` function to provide this path
+    to the correct location. Returns the same path object as given in the `path`
+    argument.
```
lwjohnst86, line 16:
```suggestion
    `datapackage.json` file. Returns the same path object as given in the `path`
    argument.
```

Small fix, given the "local-first" approach.

tests/core/test_write_package_properties.py
```diff
@@ -0,0 +1,113 @@
...
+
+def test_writes_properties_when_file_does_not_exist(path, package_properties):
+    """Should write properties to file when the file doesn't yet exist."""
+    assert write_package_properties(path, package_properties) == path
+    assert_file_contains(path, package_properties)
+
+
+def test_rewrites_package_properties_when_file_exists(path, package_properties):
+    """Should write properties to file when the file already exists, updating existing
+    package properties."""
```
lwjohnst86, line 51: Is it "updating" existing properties or replacing the whole file?

lwjohnst86 (commented):
Very nice! Some small comments

seedcase_sprout/core/write_package_properties.py
martonvago: I think that with our current setup, you are not expected to use the `write_package_properties` function to edit AND write, as you could, as you say, accidentally delete your resource properties.
But if you edit THEN write, then it's all good, because `edit_package_properties` doesn't modify existing resource properties.

More generally, maybe it would make more sense to have a single function for writing to the `datapackage.json` file. Then it would be very obvious that whatever you give to that function will be written to file as is, no merging, no clever updates. This function pretty much does this already. 
I don't think we're likely to implement a write function that touches only the bits of the file we want to change?

tests/core/test_write_package_properties.py
martonvago: Replacing the whole file. I'll rephrase it! But I meant that "package properties are affected by the function". And then the next test tests that "resource properties are affected".

tests/core/test_write_package_properties.py
```diff
@@ -0,0 +1,113 @@
...
+
+
+def test_writes_properties_when_file_does_not_exist(path, package_properties):
+    """Should write properties to file when the file doesn't yet exist."""
+    assert write_package_properties(path, package_properties) == path
+    assert_file_contains(path, package_properties)
+
+
+def test_rewrites_package_properties_when_file_exists(path, package_properties):
+    """Should write properties to file when the file already exists, updating existing
```
martonvago, line 50:
```suggestion
    """Should write properties to file when the file already exists, rewriting existing
```

tests/core/test_write_package_properties.py
```diff
@@ -0,0 +1,113 @@
...
+    package properties."""
+    old_properties = package_properties.compact_dict | {"name": "old-name"}
+    write_json(old_properties, path)
+
+    assert write_package_properties(path, package_properties) == path
+    assert_file_contains(path, package_properties)
+
+
+def test_rewrites_resource_properties_when_file_exists(path, package_properties):
+    """Should write properties to file when the file already exists, updating existing
```
martonvago, line 60:
```suggestion
    """Should write properties to file when the file already exists, rewriting existing
```

seedcase_sprout/core/write_package_properties.py
martonvago: Yeah!! I put this [random comment](https://github.com/seedcase-project/seedcase-sprout/pull/990#issuecomment-2607564805) on another PR that pretty much says the same thing :grin:

lwjohnst86 (approved):
:tada:

PR #988, opened 2025-01-22
refactor: :recycle: use new checks in `create_resource_properties()`

martonvago:
## Description

This PR updates `create_resource_properties` to use the new check functions. It also streamlines the logic a bit.

## Checklist

- [x] Added or updated tests
- [x] Updated documentation
- [x] Ran `just run-all`


seedcase_sprout/core/create_resource_properties.py
```diff
@@ -27,23 +24,19 @@ def create_resource_properties(
...
+            `CheckError`s, one error per failed check.
     """
-    properties = properties.compact_dict
     check_is_dir(path)
-    verify_properties_are_well_formed(properties, ResourceError.type)
-    data_path = create_relative_resource_data_path(path)
-    edited_properties = edit_property_field(properties, "path", str(data_path))
-
-    return ResourceProperties.from_dict(edited_properties)
+    properties.path = str(create_relative_resource_data_path(path))
```
martonvago, line 40: This comes out way nicer now!
And if we do this before the check, the path gets checked as well :tada:

lwjohnst86 (approved):
:tada:

PR #990, opened 2025-01-22
refactor: :recycle: use checks in `write_resource_properties()`

martonvago:
## Description

This PR switches to the new check functions in `write_resource_properties`.

There is probably an argument for reorganising this function a bit, but in this PR I am just adding the check functions.

## Checklist

- [x] Added or updated tests
- [x] Updated documentation
- [x] Ran `just run-all`


tests/core/test_write_resource_properties.py
martonvago, line 1: This could do with a couple more tests, but I expect the function to change a bit, so I'll add them a bit later.

martonvago:
Having looked at these functions a bit, maybe a natural organisation would be:
- `edit_package_properties` (leave as is):
  - takes a partial `PackageProperties` 
  - returns a complete `PackageProperties`, ready to write to file
- `edit_resource_properties` (= most of the current `write_resource_properties`): 
  - takes a partial `ResourceProperties`
  - returns a complete **`PackageProperties`**, ready to write to file 
- `write_properties` (= the current `write_package_properties`)
  - takes a complete `PackageProperties` 

The big change here would be that `edit_resource_properties` would return a `PackageProperties` (with the `ResourceProperties` updated or added). This gives a nice flow, but I'm not sure if it's misleading?

lwjohnst86 (approved):
:+1:

PR #992, opened 2025-01-22
ci: :fire: remove left over workflows

lwjohnst86:
## Description

Small cleanup of the workflows.

This PR needs a quick review.


.github/workflows/puml-to-svg.yml
signekb, line 1: I feel like I have asked this before: but will be stop using puml entirely since this workflow is removed? What if we want to update an old puml?

signekb (commented):
🔥 🔥 
Just one q.

.github/workflows/puml-to-svg.yml
lwjohnst86: You did ask already haha this workflow never actually worked and it was such a low priority I didn't spend time debugging it (plus workflows are such a pain to debug).

PR #996, opened 2025-01-23
refactor: :recycle: `path_package_properties()` -> `path_properties()`

signekb:
## Description

We don't have other properties files (like e.g., resource properties files), so we might as well just call it `path_properties`

This PR needs a quick review.

## Checklist

- [X] Added or updated tests
- [ ] Ran `just run-all`


martonvago (approved):
Good point!

lwjohnst86 (approved):

PR #997, opened 2025-01-24
feat: ✨ output everything from code cells in Reference docs

signekb:
## Description

This PR adds the setting `ipynb-shell-interactivity: all` in a yml header in all Reference docs. 
This ensures that all output from the cell is printed below it. 

It also makes the `output: true` setting in all example cells redundant.

This PR needs a quick review.

## Checklist

- [X] Updated documentation
- [X] Ran `just run-all`


lwjohnst86 (approved):
I don't fully understand that render code, but if it works then :+1:

lwjohnst86:
Jumped the gun here. This shouldn't be a `feat`, this is `docs` or `chore`

PR #998, opened 2025-01-27
refactor: :recycle: commonise Sprout checks

martonvago:
## Description

This PR shares logic across Sprout checks.
As all top-level Sprout checks call a check function from the `checks` package and then add Sprout-specific logic, there was lots of overlap between them. Here, I left that logic in `check_properties`, and made `check_resource_properties` and `check_package_properties` rely on it. The underlying idea being "do all checks and ignore irrelevant errors". This works with the existing error-ignoring strategy.

In our codebase, as it is currently, this made more sense than to do it the other way and build `check_properties` by calling `check_resource_properties` and `check_package_properties`. But feel free to argue otherwise!

Doing it this way, we're only using one check function from the `checks` package (the simplest one), so that's nice. 

## Checklist

- [x] Added or updated tests
- [x] Updated documentation
- [x] Ran `just run-all`


seedcase_sprout/core/checks/check_error_matcher.py
martonvago, line 1: Here, I made the matching of `json_path`s use regular expressions instead of substrings. This is more versatile and accurate. E.g., you can match the `description` field on resources only.

seedcase_sprout/core/sprout_checks/check_package_properties.py
```diff
@@ -28,16 +27,11 @@ def check_package_properties(
...
-    errors = checks.check_package_properties(properties) + get_sprout_package_errors(
-        properties
+    return check_properties(
+        properties,
+        ignore=[
+            *ignore,
+            CheckErrorMatcher(json_path=RESOURCE_FIELD_PATTERN),
+            CheckErrorMatcher(json_path=r"resources$", validator="required"),
+        ],
     )
```
martonvago, line 37: Do all checks, but ignore any resource errors and ignore a missing `resources` field

seedcase_sprout/core/sprout_checks/check_properties.py
```diff
@@ -28,16 +37,21 @@ def check_properties(properties: dict, ignore: list[CheckErrorMatcher] = []) ->
     errors = checks.check_properties(properties)
     errors += get_sprout_package_errors(properties)
 
-    for index, resource in enumerate(properties.get("resources", [])):
-        errors += get_sprout_resource_errors(resource, index)
+    if isinstance(properties.get("resources"), list):
```
martonvago, line 40: Added this check to account for situations where `resources` is the wrong type

seedcase_sprout/core/sprout_checks/check_properties.py
```diff
@@ -28,16 +37,21 @@ def check_properties(properties: dict, ignore: list[CheckErrorMatcher] = []) ->
...
-                validator="type", json_path="path", message="not of type 'array'"
+                validator="required", json_path=rf"{RESOURCE_FIELD_PATTERN}\.data$"
+            ),
+            CheckErrorMatcher(
+                validator="type",
+                json_path=rf"{RESOURCE_FIELD_PATTERN}\.path$",
+                message="not of type 'array'",
             ),
         ],
     )
```
martonvago, line 57: This doesn't change the behaviour, but it's more accurate, because it only matches these fields on resource properties

seedcase_sprout/core/sprout_checks/check_resource_properties.py
```diff
@@ -31,25 +29,14 @@ def check_resource_properties(
...
-    errors = sorted(set(errors))
-
-    if errors:
-        raise ExceptionGroup(
-            f"The following checks failed for the resource properties:\n{properties}",
-            errors,
+    try:
+        check_properties(
+            {"resources": [properties]},
+            ignore=[*ignore, CheckErrorMatcher(json_path=PACKAGE_FIELD_PATTERN)],
```
martonvago, line 35: Do all checks first, then ignore any package errors

seedcase_sprout/core/sprout_checks/check_resource_properties.py
```diff
@@ -31,25 +29,14 @@ def check_resource_properties(
...
         )
+    except ExceptionGroup as error_info:
+        for error in error_info.exceptions:
+            error.json_path = error.json_path.replace("resources[0].", "")
+        raise error_info
```
martonvago, line 40: I'm changing the error `json_path` from e.g. `$.resources[0].name` to `$.name`.
This is not strictly necessary, depends on what `json_path` we want when only a single resource is being checked.

tests/core/checks/test_check_error_matcher.py
```diff
@@ -31,18 +31,23 @@ def test_matches_message(message, matcher, expected):
     "json_path,matcher,expected",
     [
         ("$.match", None, True),
-        ("$.no.match", "", False),
+        ("$.match", "", True),
```
lwjohnst86, line 34: Wait, was `no.match` removed?

seedcase_sprout/core/sprout_checks/check_resource_properties.py
lwjohnst86: Nice to clean up all of this!

lwjohnst86 (approved):
Very nice! Only a small question!

tests/core/checks/test_check_error_matcher.py
martonvago: The earlier version matched if the matcher's `json_path` was either equal to the error's full `json_path` or equal to the very last field in the error's `json_path`. `""` can be neither, so I made it not match. Now, it's a match if the matcher's `json_path` matches the error's `json_path` as a regex. `""` as a regex matches everything (at least in Python's implementation), so I let it match.

But this is completely up to how we want it to behave, so I can flip it back.

PR #999, opened 2025-01-27
refactor :recycle: allow properties input in checks

martonvago:
## Description

This PR makes the suggestion of having check functions in Sprout accept both `dict`s and `Properties` objects.
I thought of this when I was writing up the guide for check functions and kept having to transform each `Properties` object into a `dict` before using it in an example.

Not strictly necessary and maybe we don't want it, but here it is :1st_place_medal: 

## Checklist

- [x] Added or updated tests
- [x] Updated documentation
- [x] Ran `just run-all`


lwjohnst86:
@martonvago there's a bunch of conflicts, could you resolve them?

lwjohnst86 (approved):
:tada:

PR #1000, opened 2025-01-27
refactor: :recycle: make schema path resolve when working directory different

martonvago:
## Description

This PR makes it so that the path to the Data Package schema can be resolved even when code is executed from an "unexpected" working directory, e.g. in the guide folder in a Jupyter notebook.

This PR needs a quick review.

## Checklist

- [x] Added or updated tests
- [x] Updated documentation
- [x] Ran `just run-all`


lwjohnst86 (approved):
oooo good catch! I was wondering how to reference internal files, good to know now!

PR #1016, opened 2025-01-28
docs: :fire: move text into docstrings, remove leftover resource `.puml`

lwjohnst86:
## Description

This PR cleans up the mostly resource section with the things that have been implemented already. Removes old PlantUML files.

This PR needs a quick review.

## Checklist

- [x] Updated documentation
- [x] Ran `just run-all`


seedcase_sprout/core/write_package_properties.py
```diff
@@ -15,8 +15,13 @@ def write_package_properties(path: Path, properties: PackageProperties) -> Path:
     argument.
 
     Args:
-        path: The path to the `datapackage.json` file.
-        properties: The package properties to write.
+        path: The `path` argument is to an existing `datapackage.json` file, to a
+            folder that exists but is empty, or to a folder that doesn't exist yet.
+            Either way, the `datapackage.json` file will be created or overwritten
+            in the path given. Use `path_properties()` as a helper to get the
+            correct path.
```
martonvago, line 22: And if no path is given, could/should it default to the working directory? Or is this something to add later e.g. in lib?
(Same for any function that needs a path pointing to the `datapackage.json`)

martonvago (approved):
Nice!!

seedcase_sprout/core/write_package_properties.py
lwjohnst86: I think it makes most sense in lib, but I'm sure we will come back to this later, I was considering this as well.

Issue #995, opened 2025-01-23
Have one `write_properties()` function


lwjohnst86:
> Having looked at these functions a bit, maybe a natural organisation would be:
> - `edit_package_properties` (leave as is):
>   - takes a partial `PackageProperties` 
>   - returns a complete `PackageProperties`, ready to write to file
> - `edit_resource_properties` (= most of the current `write_resource_properties`): 
>   - takes a partial `ResourceProperties`
>   - returns a complete **`PackageProperties`**, ready to write to file 
> - `write_properties` (= the current `write_package_properties`)
>   - takes a complete `PackageProperties` 
>
> The big change here would be that `edit_resource_properties` would return a `PackageProperties` (with the `ResourceProperties` updated or added). This gives a nice flow, but I'm not sure if it's misleading?

_Originally posted by @martonvago in https://github.com/seedcase-project/seedcase-sprout/issues/990#issuecomment-2607564805_

PR #1019, opened 2025-01-28
docs: :recycle: convert from PlantUML into pseudocode

lwjohnst86:
## Description

This PR converts the `.puml` file into Python pseudocode.

## Checklist

- [x] Updated documentation


docs/design/interface/pseudocode/extract_resource_properties.py
```diff
@@ -0,0 +1,23 @@
...
+    function is often followed by `edit_resource_properties()` to fill in any
+    remaining missing fields, like the `path` property field.  Usually, you use
+    either this function or the `create_resource_properties()` function to
+    create the initial resource properties for a specific (new) data resource.
+
+    Args:
+        data_path: The path to a raw data file of a supported format.
+
+    Returns:
+        Outputs a `ResourceProperties` object. Use `write_resource_properties()`
```
martonvago, line 16: We'll just have to make sure the flow still works if we do https://github.com/seedcase-project/seedcase-sprout/issues/995

docs/design/interface/pseudocode/extract_resource_properties.py
```diff
@@ -0,0 +1,23 @@
...
+            to save the object to the `datapackage.json` file.
+    """
+    check_is_file(data_path)
+    check_is_supported_format(data_path)
+    # Make use of frictionless here?
+    properties = extract_properties_from_file(data_path)
+    return check_resource_properties(properties)
```
martonvago, line 23: And probably ignore some expected errors like a missing `path`.
Or maybe there is also an argument for allowing bad resource properties to be returned by this function in general? If, for whatever reason, the extraction doesn't yield a correct set of properties, it is still useful to me as a user to have something I can fix/edit and not have to type it all out by hand.

docs/design/interface/pseudocode/extract_resource_properties.py
```diff
@@ -0,0 +1,23 @@
...
+    Args:
+        data_path: The path to a raw data file of a supported format.
+
+    Returns:
+        Outputs a `ResourceProperties` object. Use `write_resource_properties()`
+            to save the object to the `datapackage.json` file.
+    """
+    check_is_file(data_path)
+    check_is_supported_format(data_path)
+    # Make use of frictionless here?
```
martonvago, line 21: Yeah this is a good question! 

If we're okay with using frictionless, we could get around the task of writing a converter from their classes to our properties classes by having them output a `dict` and feeding that `dict` to our `ResourceProperties.from_dict`. On paper it sounds great!

The other, unquestionably more arduous, road would be to turn the data into a pandas dataframe (polars not yet supported for this), use pandera to extract the metadata, and convert that output to `ResourceProperties`.

martonvago (dismissed):

docs/design/interface/pseudocode/extract_resource_properties.py
```diff
@@ -0,0 +1,23 @@
+# ruff: noqa
+def extract_resource_properties(data_path: Path) -> ResourceProperties:
+    """Extracts resource properties from a raw data file.
+
+    This function takes the data file found at the `data_path` location and
+    extracts properties from the file into a `ResourceProperties` object. This
+    function is often followed by `edit_resource_properties()` to fill in any
+    remaining missing fields, like the `path` property field.  Usually, you use
```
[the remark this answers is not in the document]
lwjohnst86: Yea, agreed. I think once we use it in the guides/examples, we'll have a better idea for it.

docs/design/interface/pseudocode/extract_resource_properties.py
lwjohnst86: Yes, for sure, I already have it in my mind to refactor those once we've discussed/decided! :grinning:

PR #1021, opened 2025-01-29
docs: :building_construction: move `.puml` into pseudocode

lwjohnst86:
## Description

This PR moves the PlantUML diagram over into pseudocode and also adds a basic Mermaid diagram of the input and output flow.

## Checklist

- [x] Updated documentation
- [x] Ran `just run-all`


docs/design/interface/pseudocode/write_resource_data_to_raw.py
```diff
@@ -0,0 +1,120 @@
+# ruff: noqa
+def write_resource_data_to_raw(data_path, path, resource_properties):
```
lwjohnst86, line 2:
```suggestion
def write_resource_data_to_raw(data_path, path, resource_properties) -> Path:
```

Forgot this.

lwjohnst86: This was the cleanest I could think of for arguments. Would love thoughts!

docs/design/interface/pseudocode/write_resource_data_to_raw.py
```diff
@@ -0,0 +1,120 @@
...
+    path object of the created file.
+
+        Args:
+            data_path: Path to a raw data file that you want Sprout to store
+                into the data package.
+            path: Path to the raw resources folder. Use `path_resources_raw()`
+                to help provide the correct path.
+            resource_properties: The properties object for the specific resource.
+                Use `read_properties()` to read the properties for the resource
+                and `get_properties()` to get the correct resource properties.
```
lwjohnst86, line 20: I am not sure a clean/documented way we could indicate to give the function the relevant resource properties. We don't have this `get_properties()` yet, I imagine it could be something that could be used to say "get the resource properties for this resource ID". Unless there is a very specific and easy Python way of doing this.

docs/design/interface/pseudocode/write_resource_data_to_raw.py
```diff
@@ -0,0 +1,120 @@
...
+    data_path: Path, resource_properties: ResourceProperties
+) -> Error | None:
+    """Basic checks on the data against the resource properties.
+
+    Checks the data for basic things, including against the specific resource's
+    properties for things like:
+
+    - Can it be at a minimal read without problems or warnings?
+    - Do the columns in the data file match those in the properties?
+    - Do the data types in the data file match those in the properties?
```
lwjohnst86, line 90: These were the initial ones I was thinking about, but I guess as we use it in examples and real-world data, we could add more. Could even eventually move this function out into the `checks` package.

docs/design/interface/python-functions.qmd
```diff
@@ -145,19 +145,23 @@ fig-alt="A Plant UML schematic of the detailed code flow within the `create_reso
...
+    in_data_path[/data_path/]
+    in_path[/path/]
+    in_properties[/resource_properties/]
+    function("write_resource_data_to_raw()")
+    out[("{path}/{timestamp}-{uuid}.{ext}.csv")]
+    in_data_path --> function
+    in_path --> function
+    in_properties --> function
+    function --> out
+```
```
lwjohnst86, line 164: This looks like: https://mermaid.ink/svg/pako:eNp1UMuKwzAM_JWiUwtNc8-hp_7B3jYuQThKY4jtoMibXYz_fd2WmH36pHmYGSmC9j1BA8PkVz0ii3K7_IzrehTsZpSxrctYX4v8VH6S7GdiMbS0NdPiA2v6Qm7OITgtxru9gpWNUFe8jyTxHeO6Pyg4PP0-SJu98Z6W6ijG0iJo51TFEEyfTpHeJZ308pa_XH9vsKuqcwn9tsE_Smn8h76hh5SbwREssUXT5zPGu0eBjGRJQZPHngYMkyhQLmUrBvEvH05DIxzoCOzDbYRmwGnJKMy5M10M3hhtYWd0r95vOH0C1aWX3g

docs/design/interface/pseudocode/write_resource_data_to_raw.py
```diff
@@ -0,0 +1,120 @@
...
+def write_resource_data_to_raw(data_path, path, resource_properties):
+    """Writes the raw data file into the resource's raw data folder.
+
+    Copy the file from `data_path` over into the resource location given by
+    `path`. This will compress the file and use a timestamped, unique file
+    name to store it as a backup. See the
+    [design](https://sprout.seedcase-project.org/docs/design/) docs for an
+    explanation of this file. Use `path_resource_raw()` to provide the
+    correct `path` location. Copies and compresses the file, and outputs the
+    path object of the created file.
```
martonvago, line 11: Maybe include that the data is checked against the metadata?

docs/design/interface/pseudocode/write_resource_data_to_raw.py
```diff
@@ -0,0 +1,120 @@
...
+    check_is_file(data_path)
+    check_is_supported_format(data_path)
+    check_data_basics(data_path, resource_properties)
+    check_data_constraints(data_path, resource_properties)
+
+    raw_resource_path = Path(path / create_raw_file_name(data_path))
+    return write_compressed_file(data_path, raw_resource_path)
+
+
+def create_raw_file_name(data_path: Path) -> Error | None:
```
martonvago, line 40:
```suggestion
def create_raw_file_name(data_path: Path) -> str:
```
This could just return a string, right? No need to make it a `Path`

docs/design/interface/pseudocode/write_resource_data_to_raw.py
```diff
@@ -0,0 +1,120 @@
...
+    """
+    # Untested code.
+    timestamp = datetime.datetime.now().isoformat()
+    uuid = str(uuid.uuid4())
+    extension = data_path.suffix
+    return Path("{timestamp}-{uuid}.{extension}.gz")
+
+
+def write_compressed_file(data_path: Path, path: Path) -> Path:
+    """Write and compress the raw data file into the resource's raw data folder.
```
martonvago, line 63:
```suggestion
    """Compress and write the raw data file into the resource's raw data folder.
```

docs/design/interface/pseudocode/write_resource_data_to_raw.py
```diff
@@ -0,0 +1,120 @@
...
+
+        Returns:
+            The path to the written compressed raw resource file.
+    """
+    return path
+
+
+def check_data_basics(
+    data_path: Path, resource_properties: ResourceProperties
+) -> Error | None:
```
martonvago, line 82:
```suggestion
) -> Path:
```
To make it like the other check functions, it could return its input if the checks passed and throw an error if any failed.

martonvago (changes requested):
Makes sense in general, just added some comments!

docs/design/interface/pseudocode/write_resource_data_to_raw.py
lwjohnst86: Hmm, good point. Since the `resource_properties` object will have the `path` already, we can add `raw` to the end of it and fix that problem. And we won't even need to ask for another argument.

Just a clarifying point, with the local-first, there is no package ID, only resource ID.

docs/design/interface/pseudocode/write_resource_data_to_raw.py
```diff
@@ -0,0 +1,116 @@
+# ruff: noqa
+def write_resource_data_to_raw(data_path, resource_properties) -> Path:
```
lwjohnst86, line 2: I removed the `path` so we can use the `path` properties instead. One thing we need to consider is the location of where this function will run. We either need to figure a way to give an absolute path, or restrict this function to only running in a directory that has a `datapackage.json` (so it know's where root is). Or, we have a function to seek out what the root of the package is, if this is run from a subfolder.

docs/design/interface/pseudocode/write_resource_data_to_raw.py
martonvago: That sounds good! A similar idea then would be to just give the resource ID. That would prevent people inputting `ResourceProperties` that haven't been linked/added to a package yet (and so don't have a `path`). But we will see how it makes sense when it's actually used.

Yeah, I have some haziness about how we will go from "global" to local-first. I kind of assumed we would keep some version of the current setup with server environments in mind, and add some wrappers (or default values) around that to offer a local-first interface. But I'll find out soon! :grin:

docs/design/interface/pseudocode/write_resource_data_to_raw.py
```diff
@@ -0,0 +1,116 @@
...
+        Examples:
+            ```{python}
+            #| eval: false
+            ```
+    """
+    check_is_file(data_path)
+    check_is_supported_format(data_path)
+    check_data_basics(data_path, resource_properties)
+    check_data_constraints(data_path, resource_properties)
+    raw_dir = Path(resource_properties.path / "raw")
```
martonvago, line 32: I think `path` is `resources/id/data.parquet`, but this a minor point.

martonvago (dismissed):

docs/design/interface/pseudocode/write_resource_data_to_raw.py
```diff
@@ -0,0 +1,116 @@
+# ruff: noqa
+def write_resource_data_to_raw(data_path, resource_properties) -> Path:
+    """Writes the raw data file into the resource's raw data folder.
+
+    Copy the file from `data_path` over into the resource location given by
+    `path`. This will compress the file and use a timestamped, unique file
+    name to store it as a backup. See the
+    [design](https://sprout.seedcase-project.org/docs/design/) docs for an
+    explanation of this file. Data is always checked against the metadata
```
signekb, line 9:
```suggestion
    explanation of this file. Data is always checked against the properties
```

docs/design/interface/pseudocode/write_resource_data_to_raw.py
signekb: That would make sense to me. And maybe also `check_data_constraints()` below?

docs/design/interface/python-functions.qmd
```diff
@@ -145,19 +145,21 @@ fig-alt="A Plant UML schematic of the detailed code flow within the `create_reso
...
-correct `path` location. Copies and compresses the file, and outputs the
-path object of the created file.
+See the help documentation with `help(write_resource_data_to_raw)` for
+more details.
 
-![Diagram showing the internal function flow of the
-`write_resource_data_to_raw()`
-function.](images/core/write-resource-data-to-raw.svg){#fig-write-resource-data-to-raw
-fig-alt="A Plant UML schematic of the detailed code flow within the `write_resource_data_to_raw()` function."}
+```{mermaid}
```
signekb, line 153: When I attempt to preview this locally, I can't see it. 

The callout block looks like this (with only a glimpse of the diagram's corner): 
image

signekb (changes requested):
Nice 👍  Looks good to me, except that I can't see the mermaid diagram using `quarto preview`.

docs/design/interface/pseudocode/write_resource_data_to_raw.py
lwjohnst86, line 32:
```suggestion
		# Since `path` is to the `data.parquet` file.
    raw_dir = Path(resource_properties.path.parent / "raw")
```

docs/design/interface/python-functions.qmd
lwjohnst86: Yea, there seems to be a bug in Mermaid that makes it not show up on some browsers. We can't really do anything about it yet and I don't know the solution (yet)

PR #1039, opened 2025-02-10
feat: :sparkles: add `get_nested_attr()`

martonvago:
## Description

This PR adds a utility function for making it easier to access nested properties on objects.
Our Properties class structure has lots of nesting and all attributes default to None, so when I want to access a nested attribute (e.g. `resource_properties.dialect.header`) I always have to make sure it doesn't throw an error.

With this function, we can instead say `get_nested_attr(resource_properties, "dialect.header", default=True)`, optionally providing a default value. Similar to the built-in `getattr`.

This PR needs a quick review.

## Checklist

- [x] Added or updated tests
- [x] Updated documentation
- [x] Ran `just run-all`

### DIFF
--- a/seedcase_sprout/core/get_nested_attr.py
+++ b/seedcase_sprout/core/get_nested_attr.py
@@ -1,0 +1,51 @@
+from typing import Any, TypeVar
+
+T = TypeVar("T")
+
+
+def get_nested_attr(
+    base_object: Any, attributes: str, default: T | None = None
+) -> T | None:
+    """Returns the attribute specified by `attributes`.
+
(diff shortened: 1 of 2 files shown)


seedcase_sprout/core/get_nested_attr.py
```diff
@@ -0,0 +1,51 @@
+from typing import Any, TypeVar
+
+T = TypeVar("T")
```
lwjohnst86, line 3: Ah, that's neat!

lwjohnst86 (approved):
Cool! Looks good :grin:

signekb (approved):
Looks good to me ✨

PR #1041, opened 2025-02-12
style: :lipstick: refactor so WIP status of interface are icons

lwjohnst86:
## Description

The main reason for this stylistic change is that Mermaid diagrams don't seem to like being in a callout-block. So I removed the callout blocks and used icons instead.

This PR needs a quick review.

## Checklist

- [x] Updated documentation
- [x] Ran `just run-all`


signekb (approved):
Nice ✨  
(I did really like the callout blocks though - unfortunate with the callout block vs mermaid bug)

lwjohnst86:
@signekb I knooowwww, I liked them too :cry:

PR #1046, opened 2025-02-12
refactor: :recycle: convert `example_package_properties` to function

lwjohnst86:
## Description

As I was updating the `create_package_properties()`, I found out an issue with the example object I made, so I converted it into a function to fix that issue and put it on the website too.

This PR needs a quick review.

## Checklist

- [x] Updated documentation
- [x] Ran `just run-all`


lwjohnst86:
This actually depends on my other work, so it will fail the website build. Don't merge until the other PR has been approved and merged.

seedcase_sprout/core/example_package_properties.py
```diff
@@ -3,23 +3,46 @@
...
+    Returns:
+        Outputs a correctly formatted, `PackageProperties` object.
+
+    Examples:
+        ```{python}
+        import seedcase_sprout.core as sp
+        sp.example_package_properties()
+        ```
+    """
+    default_properties = PackageProperties.default()
```
martonvago, line 23: What would you think about just inlining those 3 attributes set by the `default` method? Then you wouldn't need to juggle between `dict`s and Properties classes, and you could decide if you wanted the generated values for `id` and `created` to be different for each example object.

Or if you want to use the default, you could set other attributes on it directly (`default_properties.name = ...`), and that would also avoid the update and the conversions.

seedcase_sprout/core/example_package_properties.py
```diff
@@ -3,23 +3,46 @@
...
+from seedcase_sprout.core.sprout_checks.check_package_properties import (
+    check_package_properties,
 )
+
+
+def example_package_properties() -> PackageProperties:
+    """Generate an example package properties object.
+
+    Returns:
+        Outputs a correctly formatted, `PackageProperties` object.
```
signekb, line 15:
```suggestion
        Outputs a correctly formatted example `PackageProperties` object.
```

seedcase_sprout/core/example_package_properties.py
```diff
@@ -3,23 +3,46 @@
...
+
+    Examples:
+        ```{python}
+        import seedcase_sprout.core as sp
+        sp.example_package_properties()
+        ```
+    """
+    default_properties = PackageProperties.default()
+    update_properties = PackageProperties(
```
signekb, line 24:
```suggestion
    updated_properties = PackageProperties(
```
Maybe? If this isn't redundant with Marton's suggestion above (L23).

signekb (changes requested):
Nice 👍  Just minor things - in extension to Marton's suggestion.

seedcase_sprout/core/example_package_properties.py
lwjohnst86: Good point...

tests/core/test_example_package_properties.py
martonvago, line 1: This will be terribly nitpicky, but instead of testing the return type, maybe a more substantial test would be removing the `check_package_properties` check in the implementation and adding it here as a test. I think this would make sense because the example properties are fixed, so they don't need to be checked every time the code runs, but it is for sure useful to check them at least once (but this can happen separately from running the code)!

tests/core/test_example_package_properties.py
```diff
@@ -0,0 +1,6 @@
+from seedcase_sprout.core.example_package_properties import example_package_properties
+from seedcase_sprout.core.properties import PackageProperties
+
+
+def test_creates_correct_properties_object():
+    assert type(example_package_properties()) is PackageProperties
```
lwjohnst86, line 6:
```suggestion
    assert check_package_properties(example_package_properties())
```

tests/core/test_example_package_properties.py
```diff
@@ -0,0 +1,6 @@
+from seedcase_sprout.core.example_package_properties import example_package_properties
+from seedcase_sprout.core.properties import PackageProperties
```
lwjohnst86, line 2:
```suggestion
from seedcase_sprout.core.example_package_properties import example_package_properties
from seedcase_sprout.core.sprout_checks.check_package_properties import check_package_properties
```

tests/core/test_example_package_properties.py
lwjohnst86: Something like the below?

seedcase_sprout/core/example_package_properties.py
```diff
@@ -1,25 +1,51 @@
...
+        licenses=[
+            LicenseProperties(
+                name="ODC-BY-1.0",
+                path="https://opendatacommons.org/licenses/by",
+                title="Open Data Commons Attribution License 1.0",
+            )
+        ],
+    )
+
+    check_package_properties(properties)
```
lwjohnst86, line 49:
```suggestion
```

tests/core/test_example_package_properties.py
martonvago: Yeah!

PR #1050, opened 2025-02-14
feat: :sparkles: add Pandera checks on the data

martonvago:
## Description

This PR adds Pandera checks for all Frictionless field types (except `geojson`, which we can add later). These are just the individual checks for checking individual values, I will put up another PR where I use them to check the whole data file against a `ResourceProperties`.

For `string` fields, I added checks for the different formats, the others currently support only the default format.

There are some other customisation options missing as well (e.g. customisable truth values for booleans, custom date/time patterns). Not all of these are supported by `frictionless-py` either. I will add more of these as I build on this.

## Checklist

- [x] Added or updated tests
- [x] Updated documentation
- [x] Ran `just run-all`


pyproject.toml
```diff
@@ -33,6 +33,8 @@ requests = "^2.32.3"
 platformdirs = "^4.3.2"
 jsonschema = "^4.23.0"
 dacite = "^1.8.1"
+xmlschema = "^3.4.3"
```
martonvago, line 36: I added this because the Data Package standard uses the [XML definitions](https://www.w3.org/TR/xmlschema-2/) for date and time data types, and this was the only library I found that had validators for these. I also had a go at writing these checks myself, but they are always going to be less comprehensive than the library ones and introduce lots of complexity...

seedcase_sprout/core/sprout_checks/data/data_type_checks.py
```diff
@@ -0,0 +1,110 @@
...
+        return abs(float(lat.strip())) <= 90 and abs(float(long.strip())) <= 180
+    except ValueError:
+        return False
+
+
+EMAIL_PATTERN = r"^[^@\s]+@[^@\s]+\.[^@\s]+$"
+
+
+def check_email(value: str) -> bool:
+    """Checks if `value` meets the main format constraints of email addresses.
```
martonvago, line 70: The [best regex](https://stackoverflow.com/a/201378) for an email is super long, and we don't actually want to send emails to these addresses anyway, so I put in a simpler, more permissive one.

tests/core/sprout_checks/data/test_data_type_checks.py
```diff
@@ -0,0 +1,146 @@
...
+    check_binary,
+    check_email,
+    check_geopoint,
+    check_json,
+    check_uuid,
+    check_xml_type,
+)
+
+
+@mark.parametrize(
```
martonvago, line 21: This is obviously not a full test suite for checking the various formats, that is outsourced to `xmlschema`.

seedcase_sprout/core/sprout_checks/data/get_pandera_checks.py
```diff
@@ -0,0 +1,152 @@
...
+    check_geopoint,
+    check_json,
+    check_uuid,
+    check_xml_type,
+)
+
+# https://datapackage.org/standard/table-schema/#boolean
+BOOLEAN_VALUES = {"false", "False", "FALSE", "0", "true", "True", "TRUE", "1"}
+
+STRING_FORMAT_CHECKS = {
```
martonvago, line 24: I didn't add a specific check for the `uri` format because it's pretty difficult to define concisely and correctly what a valid URI looks like. `xmlschema` has a URI data type but doesn't do any validation for the same reason.

seedcase_sprout/core/sprout_checks/data/get_pandera_checks.py
```diff
@@ -0,0 +1,152 @@
...
+    "email": pa.Check(
+        check_email,
+        element_wise=True,
+        error="Invalid email address.",
+    ),
+    "binary": pa.Check(
+        lambda value: check_xml_type(value, XSD_BASE64_BINARY),
```
martonvago, line 31: `xmlschema` has this validator, so we might as well use it

seedcase_sprout/core/sprout_checks/data/data_type_checks.py
lwjohnst86: Yea, definitely isn't a high priority to have a correct email string!

seedcase_sprout/core/sprout_checks/data/data_type_checks.py
lwjohnst86, line 1: I'd rather not have such nested folder structure, just makes it harder to keep track of things in a project.

Also, the name should be `check_data_types.py` to follow the `action_object` naming scheme.

seedcase_sprout/core/sprout_checks/data/data_type_checks.py
```diff
@@ -0,0 +1,94 @@
...
+def check_xml_type(value: str, xml_type: str) -> bool:
+    """Checks if the given `value` is formatted as a valid XML data type.
+
+    The Frictionless Data Package standard follows the definitions at
+    https://www.w3.org/TR/xmlschema-2/ for time and date-related data types.
+    This function is for checking values against these XML data type definitions.
+
+    Args:
+        value: The value to check.
+        xml_type: The XML data type to check against.
```
lwjohnst86, line 22:
```suggestion
        type: The XML data type to check against.
```

Since `xml_type` is already in the name of the function.

seedcase_sprout/core/sprout_checks/data/data_type_checks.py
```diff
@@ -0,0 +1,94 @@
...
+
+    Returns:
+        True if the value is valid as the given type, False otherwise.
+    """
+    try:
+        XML_SCHEMA_TYPES[xml_type].decode(value)
```
lwjohnst86, line 28:
```suggestion
        XML_SCHEMA_TYPES[type].decode(value)
```

seedcase_sprout/core/sprout_checks/data/data_type_checks.py
```diff
@@ -0,0 +1,94 @@
...
+    Returns:
+        True if the value is valid as the given JSON type, False otherwise.
+    """
+    try:
+        return isinstance(json.loads(value), expected_type)
+    except json.JSONDecodeError:
+        return False
+
+
+def check_geopoint(value: str) -> bool:
```
lwjohnst86, line 50:
```suggestion
def check_is_geopoint(value: str) -> bool:
```

Just to be clearer that it is checking that it is a geopoint value.

seedcase_sprout/core/sprout_checks/data/data_type_checks.py
```diff
@@ -0,0 +1,94 @@
...
+    Args:
+        value: The value to check.
+
+    Returns:
+        True if the value meets the main format constraints, False otherwise.
+    """
+    return bool(re.match(EMAIL_PATTERN, value)) and len(value) <= 254
+
+
+def check_uuid(value: str) -> bool:
```
lwjohnst86, line 81:
```suggestion
def check_is_uuid(value: str) -> bool:
```

seedcase_sprout/core/sprout_checks/data/data_type_checks.py
```diff
@@ -0,0 +1,94 @@
...
+        True if the value is valid as the given type, False otherwise.
+    """
+    try:
+        XML_SCHEMA_TYPES[xml_type].decode(value)
+        return True
+    except (xmlschema.XMLSchemaDecodeError, KeyError):
+        return False
+
+
+def check_json(value: str, expected_type: type[list | dict]) -> bool:
```
lwjohnst86, line 34:
```suggestion
def check_is_json(value: str, expected_type: type[list | dict]) -> bool:
```

seedcase_sprout/core/sprout_checks/data/get_pandera_checks.py
```diff
@@ -0,0 +1,152 @@
...
+    XSD_TIME,
+)
+
+from seedcase_sprout.core.properties import FieldProperties
+from seedcase_sprout.core.sprout_checks.data.data_type_checks import (
+    check_email,
+    check_geopoint,
+    check_json,
+    check_uuid,
+    check_xml_type,
```
lwjohnst86, line 18:
```suggestion
    check_is_email,
    check_is_geopoint,
    check_is_json,
    check_is_uuid,
    check_is_xml_type,
```

seedcase_sprout/core/sprout_checks/data/get_pandera_checks.py
lwjohnst86: Fine with me!

seedcase_sprout/core/sprout_checks/data/get_pandera_checks.py
```diff
@@ -0,0 +1,152 @@
...
+)
+
+# https://datapackage.org/standard/table-schema/#boolean
+BOOLEAN_VALUES = {"false", "False", "FALSE", "0", "true", "True", "TRUE", "1"}
+
+STRING_FORMAT_CHECKS = {
+    "email": pa.Check(
+        check_email,
```
lwjohnst86, line 26:
```suggestion
        check_is_email,
```

seedcase_sprout/core/sprout_checks/data/get_pandera_checks.py
lwjohnst86, line 31:
```suggestion
        lambda value: check_is_xml_type(value, XSD_BASE64_BINARY),
```

seedcase_sprout/core/sprout_checks/data/get_pandera_checks.py
```diff
@@ -0,0 +1,152 @@
...
+        element_wise=True,
+        error="Invalid email address.",
+    ),
+    "binary": pa.Check(
+        lambda value: check_xml_type(value, XSD_BASE64_BINARY),
+        element_wise=True,
+        error="Invalid data format. Expected a Base64-encoded string.",
+    ),
+    "uuid": pa.Check(
+        check_uuid,
```
lwjohnst86, line 36:
```suggestion
        check_is_uuid,
```

seedcase_sprout/core/sprout_checks/data/get_pandera_checks.py
```diff
@@ -0,0 +1,152 @@
...
+)
+
+# https://datapackage.org/standard/table-schema/#boolean
+BOOLEAN_VALUES = {"false", "False", "FALSE", "0", "true", "True", "TRUE", "1"}
+
+STRING_FORMAT_CHECKS = {
+    "email": pa.Check(
+        check_email,
+        element_wise=True,
+        error="Invalid email address.",
```
lwjohnst86, line 28:
```suggestion
        error="The value {value} given for the email field doesn't seem to be a correct email string.",
```

Is it possible to add the value so the message is more helpful?

seedcase_sprout/core/sprout_checks/data/get_pandera_checks.py
```diff
@@ -0,0 +1,152 @@
...
+    ),
+    "binary": pa.Check(
+        lambda value: check_xml_type(value, XSD_BASE64_BINARY),
+        element_wise=True,
+        error="Invalid data format. Expected a Base64-encoded string.",
```
lwjohnst86, line 33:
```suggestion
        error="The value {value} given isn't in a correct data format, we expect that it be Base64-encoded.",
```

I think we could make that language easier, I don't understand what the Base64 means and why it is important.

seedcase_sprout/core/sprout_checks/data/get_pandera_checks.py
```diff
@@ -0,0 +1,152 @@
...
+    ),
+    "uuid": pa.Check(
+        check_uuid,
+        element_wise=True,
+        error="Invalid UUID format. Expected a valid UUID.",
```
lwjohnst86, line 38:
```suggestion
        error="The value {value} given for the UUID isn't in a correct format for a UUID.",
```

seedcase_sprout/core/sprout_checks/data/get_pandera_checks.py
```diff
@@ -0,0 +1,152 @@
...
+    ):
+        return [format_check]
+
+    match field.type:
+        case "boolean":
+            return [
+                pa.Check(
+                    lambda value: value in BOOLEAN_VALUES,
+                    element_wise=True,
+                    error=f"Invalid boolean value. Must be one of {BOOLEAN_VALUES}.",
```
lwjohnst86, line 63:
```suggestion
                    error=f"The logic/boolean value given is {value}, but needs to be one of {BOOLEAN_VALUES}.",
```

Is it possible to add the value into the error message?

seedcase_sprout/core/sprout_checks/data/get_pandera_checks.py
```diff
@@ -0,0 +1,152 @@
...
+                )
+            ]
+
+        case "time":
+            return [
+                pa.Check(
+                    lambda value: check_xml_type(value, XSD_TIME),
+                    element_wise=True,
+                    error="Invalid time format. Expected HH:MM:SS.",
```
lwjohnst86, line 72: Can you revise the rest to match what I've done above?

tests/core/sprout_checks/data/test_data_type_checks.py
lwjohnst86, line 1: I'll review this after the changes above :relaxed:

lwjohnst86 (changes requested):
Nice start!! Some suggestions :relaxed:

seedcase_sprout/core/sprout_checks/data/data_type_checks.py
martonvago: If we don't mind metadata checks mixing with data checks in `sprout_checks`, I don't mind!

seedcase_sprout/core/sprout_checks/data/get_pandera_checks.py
martonvago: This will be used by Pandera to produce error messages like:
```log
E               pandera.errors.SchemaErrors: {
E                   "DATA": {
E                       "DATAFRAME_CHECK": [
E                           {
E                               "schema": null,
E                               "column": "my_string",
E                               "check": "Invalid email address.",
E                               "error": "Column 'my_string' failed validator number 0: <Check check_email: Invalid email address.> failure case examples: [{'my_string': 'jane_55@doe.c o.uk'}]"
E                           }
E                       ]
E                   }
```
So it already has this `failure case examples` field.
Should I still update the error messages to the longer ones?

seedcase_sprout/core/sprout_checks/data/get_pandera_checks.py
martonvago: You'd only see this error if you set the [format](https://datapackage.org/standard/table-schema/#string) of one of your fields as `binary`, so it would make more sense in that context, because you would expect the format to be validated.

seedcase_sprout/core/sprout_checks/data/get_pandera_checks.py
lwjohnst86: Ooo, gotcha! Ok, then forget about the `{value}` bit. But yes, we should try to make nicer error messages as much as is possible. I've updated my suggestions.

seedcase_sprout/core/sprout_checks/data/get_pandera_checks.py
martonvago: Same as above, Pandera will add failure cases

seedcase_sprout/core/sprout_checks/data/get_pandera_checks.py
martonvago: Okay, thanks! I'll update them! 
But just for context, these messages will appear after all the data has been checked, so, for a large dataset, there will probably be several errors? As above, there will be one error message and then a list of failure cases. So I'm just wondering if saying "the given value" is not a bit confusing? But maybe it's obvious that it refers to each of the listed values :sweat_smile: .

But I think we shouldn't name the field as "the email field" because it's just the format, they can call it whatever they want (e.g. `contact_address`).

martonvago: How about something like: 
The given value doesn't seem to be formatted correctly as an email address.
?

lwjohnst86: Yea, that wording works!

lwjohnst86: The important thing here is that we try to make friendly and helpful error messages, not vague and overly concise ones. We want to prevent the proliferation of these types of error messages! :stuck_out_tongue:

seedcase_sprout/core/sprout_checks/check_data_types.py
```diff
@@ -0,0 +1,94 @@
...
+    The Frictionless Data Package standard follows the definitions at
+    https://www.w3.org/TR/xmlschema-2/ for time and date-related data types.
+    This function is for checking values against these XML data type definitions.
+
+    Args:
+        value: The value to check.
+        type: The XML data type to check against.
+
+    Returns:
+        True if the value is valid as the given type, False otherwise.
```
lwjohnst86, line 25:
```suggestion
        True if the value is the correct type, False otherwise.
```

seedcase_sprout/core/sprout_checks/check_data_types.py
```diff
@@ -0,0 +1,94 @@
...
+
+def check_is_json(value: str, expected_type: type[list | dict]) -> bool:
+    """Checks if the given `value` is formatted as a valid JSON object or array.
+
+    Args:
+        value: The value to check.
+        expected_type: The expected JSON type: an object or an array.
+
+    Returns:
+        True if the value is valid as the given JSON type, False otherwise.
```
lwjohnst86, line 42:
```suggestion
        True if the value is a correct JSON type, False otherwise.
```

seedcase_sprout/core/sprout_checks/check_data_types.py
```diff
@@ -0,0 +1,94 @@
...
+import xmlschema
+
+# Data types for validating time and date values
+XML_SCHEMA_TYPES = xmlschema.XMLSchema11(
+    '<xsd:schema xmlns:xsd="http://www.w3.org/2001/XMLSchema"/>'
+).types
+
+
+def check_is_xml_type(value: str, type: str) -> bool:
+    """Checks if the given `value` is formatted as a valid XML data type.
```
lwjohnst86, line 14:
```suggestion
    """Checks if `value` is correctly formatted as an XML data type.
```

seedcase_sprout/core/sprout_checks/check_data_types.py
```diff
@@ -0,0 +1,94 @@
...
+    """
+    try:
+        XML_SCHEMA_TYPES[type].decode(value)
+        return True
+    except (xmlschema.XMLSchemaDecodeError, KeyError):
+        return False
+
+
+def check_is_json(value: str, expected_type: type[list | dict]) -> bool:
+    """Checks if the given `value` is formatted as a valid JSON object or array.
```
lwjohnst86, line 35:
```suggestion
    """Checks if the `value` is correctly formatted as a JSON object or array.
```

seedcase_sprout/core/sprout_checks/check_data_types.py
```diff
@@ -0,0 +1,94 @@
...
+        True if the value is valid as the given JSON type, False otherwise.
+    """
+    try:
+        return isinstance(json.loads(value), expected_type)
+    except json.JSONDecodeError:
+        return False
+
+
+def check_is_geopoint(value: str) -> bool:
+    """Checks if the given `value` is a valid geographic point.
```
lwjohnst86, line 51:
```suggestion
    """Checks if the `value` is correctly formatted as a geographic point.
```

seedcase_sprout/core/sprout_checks/check_data_types.py
```diff
@@ -0,0 +1,94 @@
...
+
+    Args:
+        value: The value to check.
+
+    Returns:
+        True if the value is a valid geographic point, False otherwise.
```
lwjohnst86, line 57:
```suggestion
        True if the value is a geographic point, False otherwise.
```

seedcase_sprout/core/sprout_checks/check_data_types.py
```diff
@@ -0,0 +1,94 @@
...
+        value: The value to check.
+
+    Returns:
+        True if the value meets the main format constraints, False otherwise.
+    """
+    return bool(re.match(EMAIL_PATTERN, value)) and len(value) <= 254
+
+
+def check_is_uuid(value: str) -> bool:
+    """Checks if `value` can be parsed as a valid UUID.
```
lwjohnst86, line 82:
```suggestion
    """Checks if `value` can be parsed as an UUID.
```

tests/core/sprout_checks/test_check_data_types.py
```diff
@@ -0,0 +1,134 @@
...
+        ("2014-13", XSD_GYEAR_MONTH, False),
+        ("P1Y2M3DT10H30M45.343S", XSD_DURATION, True),
+        ("0Y1347M0D", XSD_DURATION, False),
+        ("c29tZSB0ZXh0IDEyMw==", XSD_BASE64_BINARY, True),
+        ("some text 123", XSD_BASE64_BINARY, False),
+        ("2022", "unknown_type", False),
+    ],
+)
+def test_checks_xml_type(value, xml_type, expected):
+    """Should determine if values are valid XML data types."""
```
lwjohnst86, line 42:
```suggestion
    """Should determine if values is a XML data types."""
```

tests/core/sprout_checks/test_check_data_types.py
```diff
@@ -0,0 +1,134 @@
...
+            ),
+            True,
+        ),
+        ("[]", False),
+        ("not,json,,", False),
+        ('"[{"prop1": "value"}, {"prop2": 123}]"', False),
+    ],
+)
+def test_checks_json_object(json_object, expected):
+    """Should determine if the input is a valid JSON object."""
```
lwjohnst86, line 63:
```suggestion
    """Should determine if the input is a JSON object."""
```

tests/core/sprout_checks/test_check_data_types.py
```diff
@@ -0,0 +1,134 @@
...
+    [
+        ("[]", True),
+        ('[{"prop1": "value"}, {"prop2": 123}]', True),
+        ("{}", False),
+        ("not,json,,", False),
+        ('{"name": "value"}', False),
+    ],
+)
+def test_checks_json_array(json_array, expected):
+    """Should determine if the input is a valid JSON array."""
```
lwjohnst86, line 78:
```suggestion
    """Should determine if the input is a JSON array."""
```

tests/core/sprout_checks/test_check_data_types.py
```diff
@@ -0,0 +1,134 @@
...
+        ("45", False),
+        ("", False),
+        ("180, 90", False),
+        ("91, 181", False),
+        ("-91, -181", False),
+        ("A, B", False),
+    ],
+)
+def test_checks_geopoint(geopoint, expected):
+    """Should determine if the input is a valid geopoint."""
```
lwjohnst86, line 102:
```suggestion
    """Should determine if the input is a geopoint."""
```

tests/core/sprout_checks/test_check_data_types.py
```diff
@@ -0,0 +1,134 @@
...
+        ("@email.co.uk", False),
+        ("jane@@email.co.uk", False),
+        ("jane@", False),
+        ("jane@email", False),
+        ("jane@email.", False),
+        (f"jane.doe{'x' * 256}@email.co.uk", False),
+    ],
+)
+def test_checks_email(email, expected):
+    """Should determine if the input is (likely to be) a valid email address."""
```
lwjohnst86, line 120:
```suggestion
    """Should determine if the input is (likely to be) an email address."""
```

lwjohnst86 (changes requested):
Looks good! Just some doc suggestions that I will commit and then will approve! :dancers:

lwjohnst86 (dismissed):
Btw, love the error messages you've updated, very nice :grin:

PR #1051, opened 2025-02-14
feat: :sparkles: add conversion from resource properties to Pandera schema

martonvago:
## Description

This PR adds the ability to convert a set of resource properties into a Pandera schema. This schema will be able to validate a polars dataframe against itself.

## Checklist

- [x] Added or updated tests
- [x] Updated documentation
- [x] Ran `just run-all`


seedcase_sprout/core/sprout_checks/data/get_polars_data_type.py
```diff
@@ -0,0 +1,45 @@
...
+
+from seedcase_sprout.core.properties import FieldType
+
+
+def get_polars_data_type(field_type: FieldType | None) -> pl.DataType:
+    """Maps Frictionless field types to polars data types.
+
+    If the Frictionless field type has formatting constraints that are not included
+    in any specialised polars data type, the mapping is to string. The formatting
+    constraints are then checked without polars.
```
martonvago, line 11: The other option would have been to define custom data types for each of these, but pandera with polars doesn't support this (yet).

seedcase_sprout/core/sprout_checks/data/get_polars_data_type.py
```diff
@@ -0,0 +1,45 @@
...
+
+    Args:
+        field_type: The Frictionless field type to map.
+
+    Returns:
+        The polars data type the field is mapped to.
+
+    Raises:
+        NotImplementedError: If Sprout doesn't yet support the Frictionless field type.
```
martonvago, line 20: I.e. I haven't written any check functions for it yet :P

seedcase_sprout/core/sprout_checks/data/get_polars_data_type.py
```diff
@@ -0,0 +1,45 @@
...
+            | "object"
+            | "geopoint"
+        ):
+            return pl.String
+        case "number":
+            return pl.Float64
+        case "integer":
+            return pl.Int64
+        case _:
+            return pl.Object
```
martonvago, line 45: This catches the `any` field type, which is also the default field type. `pl.Object` can be anything and no validation rules are imposed.

seedcase_sprout/core/sprout_checks/data/resource_properties_to_pandera_schema.py
```diff
@@ -0,0 +1,40 @@
...
+        "schema.fields",
+        default=[],
+    )
+
+    columns = {
+        field.name: pap.Column(
+            dtype=get_polars_data_type(field.type),
+            checks=get_pandera_checks(field),
+            nullable=not get_nested_attr(field, "constraints.required", default=False),
+            coerce=True,
```
martonvago, line 35: First coerce the value to `dtype`, then try to apply any validation checks.

seedcase_sprout/core/sprout_checks/data/resource_properties_to_pandera_schema.py
```diff
@@ -0,0 +1,40 @@
...
+        )
+        for field in fields
+    }
+
+    return pap.DataFrameSchema(columns, strict=True)
```
martonvago, line 40: `strict=True`: the data will need to have exactly the same columns as defined in the schema. This is the Frictionless default, but there are customisation options in the Data Package Standard.

seedcase_sprout/core/sprout_checks/data/get_polars_data_type.py
```diff
@@ -0,0 +1,45 @@
...
+            | "object"
+            | "geopoint"
+        ):
+            return pl.String
+        case "number":
+            return pl.Float64
+        case "integer":
+            return pl.Int64
+        case _:
+            return pl.String
```
martonvago, line 45: This catches the `any` field type, which is also the default field type. I chose `pl.String` for it because it has no validation constraints and can contain anything that polars can read. 
There's also a `pl.Unknown` type, which looks great, but Pandera gives a warning if I use it, saying it's not supported (not that it needs to do anything in particular...).

seedcase_sprout/core/sprout_checks/data/get_polars_data_type.py
```diff
@@ -0,0 +1,45 @@
...
+            | "date"
+            | "time"
+            | "year"
+            | "yearmonth"
+            | "duration"
+            | "list"
+            | "array"
+            | "object"
+            | "geopoint"
+        ):
```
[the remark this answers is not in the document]
martonvago: That's true, but unfortunately they are not perfect matches, so we cannot default on their validation mechanisms. 
- Frictionless `object` and `array` are explicitly JSON, but Polars `Object` and `Array` are any Python object and any Python array. 
- For the date/time related ones, they describe the same things but have different formatting constraints: Frictionless wants to validate against the XML format, Polars against something different. So we cannot safely cast e.g. a datetime value to a Polars datetime because some valid XML datetimes are invalid in Polars and some Polars datetimes are invalid in XML.
- Even booleans are problematic, because Polars will cast a number like 123 to True, but Frictionless wants to flag that as invalid.

As all we want to do with the data frame is to check it against the resource properties, it's easiest to keep these as strings and have a check function for each. If we wanted to do operations on the values, it would be better to read them in as more specific data types.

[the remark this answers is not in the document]
lwjohnst86: Gotcha. That's odd. I'll add a comment to the code then to explain this.

seedcase_sprout/core/sprout_checks/get_polars_data_type.py
```diff
@@ -0,0 +1,45 @@
...
+    Returns:
+        The Polars data type the field is mapped to.
+
+    Raises:
+        NotImplementedError: If Sprout doesn't yet support the Frictionless field type.
+    """
+    match field_type:
+        case "geojson":
+            raise NotImplementedError()
+        case (
```
lwjohnst86, line 25:
```suggestion
        # While Polars does have all these data types, there isn't a
        # perfect overlap between them and what Frictionless has, even
        # if they have similar/same names for the types. For example,
        # checks against date/datetimes/times types are different between
        # Polars and Frictionless. Or the way booleans get treated. Polars
        # may cast `123` to True, but Frictionless will indicate it is not
        # a boolean. We'll slowly improve on this as we use Sprout.
        case (
```

lwjohnst86 (dismissed):
One small comment that I'll commit, otherwise, very nice :stars:

seedcase_sprout/core/sprout_checks/get_polars_data_type.py
```diff
@@ -0,0 +1,52 @@
...
+    Returns:
+        The Polars data type the field is mapped to.
+
+    Raises:
+        NotImplementedError: If Sprout doesn't yet support the Frictionless field type.
+    """
+    match field_type:
+        case "geojson":
+            raise NotImplementedError()
+        # While Polars does have all these data types, there isn't a
```
martonvago, line 25:
```suggestion
        # While Polars does have most of these data types, there isn't a
```
Because year, yearmonth, geopoint don't have matches

martonvago:
Hmm, this is not under `src` now

lwjohnst86 (dismissed):

tests/core/sprout_checks/test_resource_properties_to_pandera_schema.py
```diff
@@ -0,0 +1,117 @@
...
+            ]
+        )
+    )
+
+    schema = resource_properties_to_pandera_schema(resource_properties)
+
+    assert [column.name for column in schema.columns.values()] == [
+        "my_date",
+        "my_boolean",
+    ]
```
signekb, line 96: Would it be relevant to check type for multiple columns here as well?

signekb (changes requested):
Looks good 👍 🌟 Just a couple of questions

src/seedcase_sprout/core/sprout_checks/get_polars_data_type.py
```diff
@@ -0,0 +1,52 @@
...
+        # a boolean. We'll slowly improve on this as we use Sprout.
+        case (
+            "string"
+            | "boolean"
+            | "datetime"
+            | "date"
+            | "time"
+            | "year"
+            | "yearmonth"
+            | "duration"
```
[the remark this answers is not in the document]
martonvago: Good question!
[Original answer](https://github.com/seedcase-project/seedcase-sprout/pull/1051#discussion_r1958523968)

Since then, for me, the question has become more about mapping between Frictionless data/field types and parquet data types, as that is the final form of the data. More details [here](https://github.com/seedcase-project/seedcase-sprout/issues/1078#issuecomment-2678326052)

PR #1060, opened 2025-02-18
feat: :sparkles: add `check_data()`

### LINKED ISSUE #801: Create and test `validate_raw_data` function (part of `write_resource_data_to_raw`)

### LINKED ISSUE #1032: Function for running basic checks on data
Implement [this function](https://github.com/seedcase-project/seedcase-sprout/blob/414606a5574087d27cc2bfe0cd481981d4c27512/docs/design/interface/pseudocode/write_resource_data_to_raw.py#L78).

When using Pandera, more general structural checks and constraint checks are done by the same function, so it might make sense to have one function for running all checks on data. If we want to do it this way, we can just update/extend the basic checks function when it comes to https://github.com/seedcase-project/seedcase-sprout/issues/886.

martonvago:
## Description

This PR adds the `check_data` function for checking data against the resource properties.

It now does structural, data type, and nullability checks, and constraint checks will be added here as well. They will be added to the Pandera schema when it is generated from the resource properties.

This is obviously only a partial implementation of the Data Package standard. I will make an issue with a list of things to implement next.

Closes #801 
Closes #1032

This PR needs an in-depth review. Focus on whether this is the behaviour we want.

## Checklist

- [x] Added or updated tests
- [x] Updated documentation
- [x] Ran `just run-all`


src/seedcase_sprout/core/sprout_checks/check_data.py
```diff
@@ -0,0 +1,90 @@
...
+)
+
+
+def check_data(data_path: Path, resource_properties: ResourceProperties) -> Path:
+    """Checks that the CSV data file at `data_path` matches the resource properties.
+
+    Runs the following checks:
+    - Can the data file be read as a CSV file?
+    - Does the data file have a header row that matches column names in the properties?
+    - Do the columns in the data file match those in the properties?
```
martonvago, line 24: This implements the default Frictionless setting that wants an exact match between data and metadata columns.

src/seedcase_sprout/core/sprout_checks/check_data.py
```diff
@@ -0,0 +1,90 @@
...
+        The `data_path` if all checks are successful.
+
+    Raises:
+        ExceptionGroup[CheckError]: If the resource properties are invalid.
+        polars.exceptions.ComputeError: If the data file cannot be parsed as CSV.
+        polars.exceptions.ShapeError: If the columns in the data file cannot
+            be matched with the columns in the resource properties.
+        pandera.errors.SchemaErrors: If data types in the data file do not match the
+            data types in the resource properties or if the data violates any
+            constraints in the resource properties.
```
martonvago, line 42: There is an argument for streamlining this a bit. We could remap all errors to a custom error (like we do for metadata checks) and expose only that. Probably makes sense to do that once the error cases are more or less stable.

src/seedcase_sprout/core/sprout_checks/check_data.py
```diff
@@ -0,0 +1,90 @@
...
+    """
+    check_resource_properties(resource_properties)
+
+    # Prepare for reading the data file
+    lazy_frame = pl.scan_csv(
+        data_path,
+        has_header=True,
+        infer_schema=False,
+        missing_utf8_is_empty_string=True,
+    )
```
martonvago, line 52: This defines how the data will be loaded. 
- As per our [discussion](https://github.com/seedcase-project/seedcase-sprout/issues/1032), we assume that the data always has a header and validate accordingly.
- I'm loading in everything as a string without inferring the data types (`infer_schema=False`) because I don't want any casting or validation at this point. That will be done by Pandera later.
- Empty strings are not converted to `null` (`missing_utf8_is_empty_string=True`) because Frictionless supports custom missing values, so we want to be in charge of null conversion.

src/seedcase_sprout/core/sprout_checks/check_data.py
```diff
@@ -0,0 +1,90 @@
...
+    lazy_frame = lazy_frame.with_columns(
+        pl.col(field.name).replace(
+            schema_missing_values
+            if field.missing_values is None
+            else field.missing_values,
+            None,
+        )
+        for field in fields
+    )
```
martonvago, line 73: Here, we determine what values count as [missing values](https://datapackage.org/standard/table-schema/#missingValues) and then convert them to null. Pandera will later check if they are allowed to be null.

For now, this only supports a string array for missing values.

src/seedcase_sprout/core/sprout_checks/check_data.py
```diff
@@ -0,0 +1,90 @@
...
+    # Check the number of columns in each data row
+    try:
+        data_frame = lazy_frame.collect()
+    except pl.exceptions.ComputeError as error:
+        if "found more fields than defined" in str(error):
+            raise pl.exceptions.ShapeError(
+                "At least one data row contains more items than the expected number of"
+                f"columns. Expected columns: {expected_columns}."
+            ) from error
+        raise
```
martonvago, line 84: This entire ugly bit is about remapping that particular `ComputeError` to a `ShapeError`. We can do without it, but the `ComputeError` says `found more fields than defined in 'Schema'\n\nConsider setting 'truncate_ragged_lines=True'.`, which is even worse than my version.

(See also the note about errors at the top of the function.)

src/seedcase_sprout/core/sprout_checks/check_data.py
```diff
@@ -0,0 +1,90 @@
...
+
+    Runs the following checks:
+    - Can the data file be read as a CSV file?
+    - Does the data file have a header row that matches column names in the properties?
+    - Do the columns in the data file match those in the properties?
+    - Do the data types in the data file match those in the properties?
+    TODO: - Does the data in the data file meet the constraints in the properties?
+
+    Args:
+        data_path: The path to the raw data file.
```
martonvago, line 29: One extension here could be to allow either a path or a list of paths as input.
Then this function could also deal with the situation where there are multiple data files.
E.g., in `build_resource_parquet` we want to check data coming from multiple data files.

src/seedcase_sprout/core/sprout_checks/check_data.py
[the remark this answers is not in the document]
martonvago: One extension here could be to allow either a path or a list of paths as input.
Then this function could also deal with the situation where there are multiple data files.
E.g., in `build_resource_parquet` we want to check data coming from multiple data files.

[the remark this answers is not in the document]
martonvago: Thinking further ahead, my feeling is that an important factor in deciding how to reorganise this is what data types we want in the final parquet format. Especially how we want to represent fields with special format constraints like `duration` or `datetime`. If we leave them as strings, it's more difficult to query them. If we convert them to a parquet datatype, we might encounter issues with how closely the parquet data type matches the frictionless data type.

src/seedcase_sprout/core/sprout_checks/check_data.py
```diff
@@ -0,0 +1,90 @@
...
+)
+from seedcase_sprout.core.sprout_checks.resource_properties_to_pandera_schema import (
+    resource_properties_to_pandera_schema,
+)
+
+
+def check_data(data_path: Path, resource_properties: ResourceProperties) -> Path:
+    """Checks that the CSV data file at `data_path` matches the resource properties.
+
+    Runs the following checks:
```
lwjohnst86, line 21:
```suggestion
    Runs the following checks:
    
```

To render the Markdown list correctly.

src/seedcase_sprout/core/sprout_checks/check_data.py
```diff
@@ -0,0 +1,90 @@
...
+)
+from seedcase_sprout.core.sprout_checks.check_resource_properties import (
+    check_resource_properties,
+)
+from seedcase_sprout.core.sprout_checks.resource_properties_to_pandera_schema import (
+    resource_properties_to_pandera_schema,
+)
+
+
+def check_data(data_path: Path, resource_properties: ResourceProperties) -> Path:
```
lwjohnst86, line 18: Hmmm, I'm not sure, why load data here? Rather than an argument as DataFrame? And why output a Path? The way I see this function being used is within a pipeline of working with the data *already loaded in*. So it feels like it would be going backward/not fitting well in that workflow.

lwjohnst86 (changes requested):
There's a key understanding bit here that I've commented. Once we've resolved that, than I can continue reviewing :relaxed:

src/seedcase_sprout/core/sprout_checks/check_data.py
lwjohnst86: Based on your comments in other PRs, I better understand where your questions/comments were coming from :smile: :stuck_out_tongue_closed_eyes:

martonvago: Ah because that was the [brief](https://github.com/seedcase-project/seedcase-sprout/blob/414606a5574087d27cc2bfe0cd481981d4c27512/docs/design/interface/pseudocode/write_resource_data_to_raw.py#L78)

martonvago: But this needs to be restructured in light of all the discussion today anyway

martonvago:
I will close this because it will be restructured completely based on the discussion about data types.
But I will keep the branch around for a while for reference becase I'll use the same building blocks.

PR #1062, opened 2025-02-19
chore: :twisted_rightwards_arrows: small build fixes

lwjohnst86:
## Description

Small fixes to the build of the website and to updates to the justfile. The `sync` command comes from Poetry >2.0.0, so everyone will need to update if they haven't already.

This PR needs a quick review.


lwjohnst86:
Since the build is failing because I put the `rm` in the wrong location, I'll just merge this since it is a small fix.

Issue #1027, opened 2025-01-31
Create and test `extract_resource_properties()` function


lwjohnst86:

signekb:
@lwjohnst86 With #1067, #1081 and #1085, we can now extract properties from csv, tsv, xlsx, xls, json, ndjson, and jsonl. Is that sufficient for this function for now or should I add extraction from geojson, topojson, and ods files as well (the remaining file formats in [`SUPPORTED_FORMATS`](https://github.com/seedcase-project/seedcase-sprout/blob/e01bbaad48f301e988ebb032fe4421d4c8c4d287/src/seedcase_sprout/core/check_is_supported_format.py#L3-L16))? The two json formats might be covered by `frictionless[json]` already.

lwjohnst86:
I think if you could add Parquet (parq, parquet), then it would be done! I doubt we will have many who store data in Parquet, but as we push for it more, maybe there will be a change.

signekb:
Parquet has been added! Sorry, I forgot to inclue them in the list above 👆

lwjohnst86:
@signekb sounds good!

signekb:
Don here #1097

PR #1067, opened 2025-02-20
feat: ✨ add initial `extract_resource_properties()` (csv, tsv, and parquet files)

signekb:
## Description

This PR adds an initial `extract_resource_properties()` function that is able to extract resource properties from csv, tsv, and parquet files. 

I will add stacked PRs to this branch to extract properties from other file types.

Related to #1027 

## Checklist

- [X] Added or updated tests
- [X] Ran `just run-all`


pyproject.toml
```diff
@@ -27,7 +27,7 @@ license = "MIT"
 [tool.poetry.dependencies]
 python = "^3.12"
 polars = "^1.5.0"
-frictionless = {extras = ["sql"], version = "^5.17.0"}
+frictionless = {extras = ["parquet"], version = "^5.18.0"}
```
signekb, line 30: I thought that I might as well remove sql, since we won’t use it.

src/seedcase_sprout/core/extract_properties_from_file.py
signekb, line 1: I thought it would be simpler to remove this function (which was only one line anyways) and then move it to the `extract_resource_properties_function()`. Let me know if you disagree.

tests/core/test_extract_resource_properties.py
```diff
@@ -0,0 +1,124 @@
+import polars as pl
+from pytest import mark
+
+from seedcase_sprout.core.extract_resource_properties import extract_resource_properties
+
+non_empty_data = pl.DataFrame(
```
signekb, line 6: Mental note: Maybe I should some add tests for non-tidy data as well (e.g., mixed data types, missing values)

tests/core/test_extract_resource_properties.py
```diff
@@ -0,0 +1,158 @@
...
+schema_non_tidy_data = {
+    "fields": [
+        {"name": "id", "type": "string"},
+        {"name": "name", "type": "string"},
+        {"name": "dob", "type": "string"},
+        {"name": "height", "type": "number"},
+        {"name": "survey_datetime", "type": "string"},
+        {"name": "completed", "type": "boolean"},
+    ]
+}
```
signekb, line 58: With differing data types, `frictionless` uses "string"

src/seedcase_sprout/core/extract_resource_properties.py
```diff
@@ -0,0 +1,53 @@
...
+        Outputs a `ResourceProperties` object. Use `write_resource_properties()`
+            to save the object to the `datapackage.json` file.
+    """
+    check_is_file(data_path)
+    check_is_supported_format(data_path)
+
+    properties = describe(data_path).to_dict()
+
+    if properties["format"] == "tsv":
+        _clean_tsv_properties(properties)
```
martonvago, line 34:
```suggestion
    properties.pop("dialect", None)
```
Just to explain: we recently [decided](https://github.com/seedcase-project/seedcase-sprout/issues/1032#issuecomment-2668266424) with Luke to remove the `dialect` property, so it makes sense to delete it here.

But if frictionless could determine that the file has another dialect than the default, we could alert the user?

src/seedcase_sprout/core/extract_resource_properties.py
```diff
@@ -0,0 +1,53 @@
...
+
+    Args:
+        data_path: The path to a raw data file of a supported format.
+
+    Returns:
+        Outputs a `ResourceProperties` object. Use `write_resource_properties()`
+            to save the object to the `datapackage.json` file.
+    """
+    check_is_file(data_path)
+    check_is_supported_format(data_path)
```
martonvago, line 29: In the context of my second comment below:
If we only support the default dialect and only for CSV when we check data, then is there a point in supporting all these formats for metadata extraction? Maybe so, I just can't see it fully quite yet.

tests/core/test_extract_resource_properties.py
```diff
@@ -0,0 +1,158 @@
...
+        "schema": expected_schema,
+    }
+    # When
+    properties = extract_resource_properties(file_path)
+
+    # Then
+    assert properties.compact_dict == expected_properties_compact_dict
+
+
+@mark.parametrize(
```
martonvago, line 126: You can put multiple `@mark.parametrize(...)`s on top of each other and then you don't have to write out the combinations yourself. So you could add `@mark.parametrize("extension", [...])` on top.

tests/core/test_extract_resource_properties.py
martonvago, line 1: Very nice tests :clap: :clap: :clap:

martonvago, line 1: Maybe it would be worth it to test all the frictionless data types? E.g. I wonder if it can recognise a `duration` or a `yearmonth`...

martonvago (changes requested):
Pretty elegant :tada: :partying_face:

src/seedcase_sprout/core/extract_resource_properties.py
lwjohnst86: Yes! We definitely do. This initial extraction is when people first make a data resource, so the original file may be in multiple formats. BUT when we save the file to the raw folder, we convert it to CSV and compress.

martonvago: Ah, so it's:
extract resource properties > convert to CSV > check CSV against resource properties > compress CSV and save to raw folder
?

lwjohnst86: Yea! Exactly :+1: At least the first time a data resource is created.

lwjohnst86: @signekb this first step would be to save the raw file. Well, partly raw, but still checked and cleaned file. Then we convert to parquet later. We want some easy level of backup/recordkeeping for data that's stored via Sprout. But we don't want to rely on CSV, there's a lot of advantages to Parquet (e.g. ensured data types via the internal schema).

martonvago: And what would be wrong with:
extract resource properties > convert to parquet > check parquet against resource properties > save parquet
?

tests/core/test_extract_resource_properties.py
signekb: Oh yeah, that's right! Thanks :star2:

src/seedcase_sprout/core/extract_resource_properties.py
lwjohnst86: I've thought about that, but how do we handle a situation when someone makes a mistake in the data, and needs to correct it?

With the CSVs, they can save another version with the corrected data, while leaving the incorrect data still in the previous CSV file, but it will be removed when converting to Parquet (we keep the most recent row if there is a duplicate in observational unit).

lwjohnst86: I've been thinking of and meaning to make a bigger diagram showing the flow of things, so maybe this would help give light to things.

Basically: extract properties > save properties > check data against properties > store to CSV as backup/raw > read them all in and join them > check against properties > remove duplicates > store as Parquet. That might be one of the simpler workflows.

signekb: My question is about the decision to save each file in the raw folder as a csv and not for example a parquet? We could still save each file as a parquet in `raw/` before merging them into the collected `data.parquet`, right?

lwjohnst86: Hmmm, I see. I think, at least for me, the reason to use CSV is that almost everyone understands what that is and how to open it. With Parquet, not so much. It's a binary format, so you can't just easily "look at it". And because it is plain text, which has other advantages. But that's definitely something to think about. Do we just save to Parquet right away in the raw folder? :thinking:

src/seedcase_sprout/core/extract_resource_properties.py
signekb: If we find that it would be nice to alert the user, we can add it later on 👍

src/seedcase_sprout/core/extract_resource_properties.py
martonvago: Re making a mistake in the data and CSVs being easy to understand:
If the flow is 
```
extract resource properties > convert to parquet > check parquet against resource properties > save parquet
```
when an error is uncovered at the check stage or they just want to change a some data, they could update the original data file and feed it again to 
```
convert to parquet > check parquet against resource properties > save parquet
```

Maybe one drawback here is that the user is left entirely in charge of their original data files: where they save them, if they save them, how they version them, etc. But there's still a safe and tidy parquet version in Sprout. And it's super easy to export that as a CSV.

tests/core/test_extract_resource_properties.py
signekb: Good point. I've added all data types except nested ones like `array`, `object`, and `json`

martonvago (dismissed):

tests/core/test_extract_resource_properties.py
signekb: Ah, will add that 👍  Ah, I wrote that a bit too fast. There seem to be a [`geojson`](https://framework.frictionlessdata.io/docs/fields/geojson.html) field type.

src/seedcase_sprout/core/extract_resource_properties.py
signekb: This seems to be a discussion that maybe should be moved outside of this PR actually? I'll create an issue and resolve this.

tests/core/test_extract_resource_properties.py
```diff
@@ -0,0 +1,214 @@
+import polars as pl
+from pytest import mark
+
+from seedcase_sprout.core.extract_resource_properties import extract_resource_properties
+
+empty_data = pl.DataFrame([])
+schema_empty_data = {"fields": []}
+
+tidy_data = pl.DataFrame(
```
signekb, line 9: this test data is getting a little out of hand, but I've now included all datatypes except `object` and `any`

lwjohnst86: I think that's fine :+1: Keep as is :relaxed:

lwjohnst86 (approved):
Looks great!! Nice job :tada:

PR #810, opened 2024-10-24
docs: :memo: add start of the creating resources guide

signekb:
## Description

This adds the guide on creating and managing data resources. It is not complete as there are some things I don't know how they may work from an implementation perspective, but this is a good starting place for us.

Closes #759

This PR needs an in-depth review.

PR #1070, opened 2025-02-20
docs: :building_construction: expand and refine naming scheme

### LINKED ISSUE #1059: Idea: More accurate names for `create_` functions
`create_` is used in different ways and for different purposes. e.g. `create_resource_properties()` makes a properties object for the first time ready to write to file. `create_package_properties()` writes it directly to file...

Another word might be `assemble_`. Either that or `create_` always also writes to. 🤔

lwjohnst86:
## Description

I've been thinking a bit more about the naming scheme. As we've developed more and used it and added the guide, some things feel a bit rough/has tension. Especially with how we've named things. Comments about each change is added to the PR.

Closes #1059

### DIFF
--- a/docs/design/architecture/naming.qmd
+++ b/docs/design/architecture/naming.qmd
@@ -39,20 +39,23 @@ We may also occasionally use "properties" to refer to the file itself.
 |----------------------------|--------------------------------------------|
 | package(s) | A data package that contains a collection of related data resources and properties. |
 | resource(s) | A single data file within a package, along with its properties and associated *raw* data files. |
-| properties or property | Metadata about a package or resource. |
-| file structure or structure | The file and folder structure of a package. |
+| properties | Metadata about a package or resource. |
 | path | A file path listing the location of folders or files that Sprout interacts with. |
 | data | The data file within a resource. Within Sprout, this includes the Parquet file and raw files. |
-| identifier | A unique identifier for a package or resource, denoted as `id`. |
+| identifier | A unique identifier for a package (*only for server environments*) or resource, denoted as `id`. |
+| changes | The changes made to the data package that will be stored in the version control repository of the data package. |
+| version | The version of the data package, denoted as `version`. |
+| observational unit | The unit of observation, meaning the data associated with a specific entity at a specific time, e.g. a person who had their weight measured on their 30th birthday. |
 
 : General objects used throughout Sprout.
 
 | File or folder | Description |
 |----------------------------|--------------------------------------------|
-| packages | The folder that contains all packages. |
+| dirs | A shortened name for folder or directory. |
+| packages | The folder that contains all packages. *Only for server environments*. |
 | resources | The folder that contains all resource files in a package. |
 | raw | The folder that contains raw data files for a data resource. |
-| raw data file | A compressed raw data file format, such as a `.zip`, `.gzip`, or `.tar.gz` file. |
+| raw data file | A compressed raw data file format, such as a `.gzip` file. |
 | JSON | The main way we save and pass information around on the package's or resource's properties. |
 | Parquet | A columnar, efficient storage file format that we use to save data within a resource. |
 
@@ -62,14 +65,13 @@ We may also occasionally use "properties" to refer to the file itself.
 
 | Action | Description |
 |----------------------------|--------------------------------------------|
-| create | Create a new object. |
-| build | Build implies either creating a new object or recreating an existing one, e.g. (re-)build a file like the README or Parquet file. |
-| list | List basic details about many objects. |
-| edit | Edit an object, specifically the properties object. |
-| add | Add another object to an existing object, specifically adding data to a resource. |
+| commit | Saves changes made in a data package to the version control system. |
+| create | Creates a new object or sets up a new object and writes it to the file system. |
+| build | Build means generating or re-generating an object from other objects, e.g. (re-)build the text for the README. |
+| update | Update an object, mainly of the properties or version objects. |
 | write | Write an object to a file. |
-| copy | Copy a file from one location to another. |
-| delete | Permanently delete an object. |
+| read | Reads an object from a file. |
+| delete | Permanently delete a file(s) associated with an object (e.g. delete a whole package or observational unit). |
 | extract | Extract specific information from an object, specifically the properties from a data file. |
 
 : General actions that Sprout can take on objects or files/folders.


docs/design/architecture/naming.qmd
```diff
@@ -39,20 +39,23 @@ We may also occasionally use "properties" to refer to the file itself.
...
 | resource(s) | A single data file within a package, along with its properties and associated *raw* data files. |
-| properties or property | Metadata about a package or resource. |
-| file structure or structure | The file and folder structure of a package. |
+| properties | Metadata about a package or resource. |
 | path | A file path listing the location of folders or files that Sprout interacts with. |
 | data | The data file within a resource. Within Sprout, this includes the Parquet file and raw files. |
-| identifier | A unique identifier for a package or resource, denoted as `id`. |
+| identifier | A unique identifier for a package (*only for server environments*) or resource, denoted as `id`. |
+| changes | The changes made to the data package that will be stored in the version control repository of the data package. |
+| version | The version of the data package, denoted as `version`. |
```
lwjohnst86, line 47: I added this, not to develop them in our current planning period/MVP, but because at some point it should be a very good idea to determine how we incorporate these. These are mainly around setting up a changelog/a formal way of setting versions and tagging those versions in the history.

docs/design/architecture/naming.qmd
```diff
@@ -62,14 +65,13 @@ We may also occasionally use "properties" to refer to the file itself.
...
 | Action | Description |
 |----------------------------|--------------------------------------------|
-| create | Create a new object. |
-| build | Build implies either creating a new object or recreating an existing one, e.g. (re-)build a file like the README or Parquet file. |
-| list | List basic details about many objects. |
-| edit | Edit an object, specifically the properties object. |
-| add | Add another object to an existing object, specifically adding data to a resource. |
+| commit | Saves changes made in a data package to the version control system. |
+| create | Creates a new object or sets up a new object and writes it to the file system. |
+| build | Build means generating or re-generating an object from other objects, e.g. (re-)build the text for the README. |
```
lwjohnst86, line 70: I decided that `create` as a word feels like it makes more sense to create an object and save it as a file (or folder). `build` on the other hand feels more like assembling something from other things, so I settled on *not* creating a file after building. This has some impacts on functions we've already made (not many) so will require a refactor. But looking forward to some thoughts!

docs/design/architecture/naming.qmd
```diff
@@ -62,14 +65,13 @@ We may also occasionally use "properties" to refer to the file itself.
...
+| update | Update an object, mainly of the properties or version objects. |
```
lwjohnst86, line 71: With the addition of `version`, it makes sense to say `update_version()`. And since `edit` and `update` are so similar in meaning, it makes more sense to rename `edit` to `update` so we only have one action.

docs/design/architecture/naming.qmd
```diff
@@ -39,20 +39,23 @@ We may also occasionally use "properties" to refer to the file itself.
 |----------------------------|--------------------------------------------|
 | package(s) | A data package that contains a collection of related data resources and properties. |
 | resource(s) | A single data file within a package, along with its properties and associated *raw* data files. |
-| properties or property | Metadata about a package or resource. |
-| file structure or structure | The file and folder structure of a package. |
+| properties | Metadata about a package or resource. |
 | path | A file path listing the location of folders or files that Sprout interacts with. |
 | data | The data file within a resource. Within Sprout, this includes the Parquet file and raw files. |
-| identifier | A unique identifier for a package or resource, denoted as `id`. |
+| identifier | A unique identifier for a package (*only for server environments*) or resource, denoted as `id`. |
```
martonvago, line 45: I thiiink we don't have resource IDs. But maybe resource names count as identifiers

docs/design/architecture/naming.qmd
martonvago: I think it mostly makes sense!
But yeah, presumably there will be some edge cases.

martonvago (approved):

docs/design/architecture/naming.qmd
martonvago: Ah the folder name, I always forget that

docs/design/architecture/naming.qmd
```diff
@@ -39,20 +39,23 @@ We may also occasionally use "properties" to refer to the file itself.
...
+| changes | The changes made to the data package that will be stored in the version control repository of the data package. |
```
signekb, line 46: Not sure what is meant by "version control repository". Is it a dir at the package root or a completely separate repository somehow?

docs/design/architecture/naming.qmd
```diff
@@ -39,20 +39,23 @@ We may also occasionally use "properties" to refer to the file itself.
...
 
 | File or folder | Description |
 |----------------------------|--------------------------------------------|
-| packages | The folder that contains all packages. |
+| dirs | A shortened name for folder or directory. |
+| packages | The folder that contains all packages. *Only for server environments*. |
 | resources | The folder that contains all resource files in a package. |
 | raw | The folder that contains raw data files for a data resource. |
-| raw data file | A compressed raw data file format, such as a `.zip`, `.gzip`, or `.tar.gz` file. |
+| raw data file | A compressed raw data file format, such as a `.gzip` file. |
```
signekb, line 58: I have lost the overview of how we store raw data files. The current plan is to store them as csvs *and* as gzips?

docs/design/architecture/naming.qmd
```diff
@@ -62,14 +65,13 @@ We may also occasionally use "properties" to refer to the file itself.
 
 | Action | Description |
 |----------------------------|--------------------------------------------|
-| create | Create a new object. |
-| build | Build implies either creating a new object or recreating an existing one, e.g. (re-)build a file like the README or Parquet file. |
-| list | List basic details about many objects. |
-| edit | Edit an object, specifically the properties object. |
-| add | Add another object to an existing object, specifically adding data to a resource. |
+| commit | Saves changes made in a data package to the version control system. |
+| create | Creates a new object or sets up a new object and writes it to the file system. |
```
signekb, line 69:
```suggestion
| create | Creates a new object or sets up a new object. |
```
Does create also write though? At least it doesn't for e.g., `create_resource_properties()` as it is written in the [resources guide PR](https://github.com/seedcase-project/seedcase-sprout/pull/810/files#) currently. There you have to use `write_resource_properties()` after creating.

docs/design/architecture/naming.qmd
signekb: Ooo, naming is hard. My intuition was that maybe we could avoid `build` altogether` and just use `create` for everything to keep the naming simpler. But I also see how that can be misleading. 
I think this makes sense, at least for now! 🌟

docs/design/architecture/naming.qmd
signekb: I like it!

docs/design/architecture/naming.qmd
signekb: I also (vaguely) remember something about this

signekb (changes requested):
Very nice 👌 I have some minor comments and questions.

docs/design/architecture/naming.qmd
signekb: And like `create`, `update` doesn't write to files, right?

docs/design/architecture/naming.qmd
lwjohnst86: Since we go by "one folder is one data package" and since "one Git repo is one folder", then we can have a Git repository (the `.git/` folder) in the data package, where they share the same root.

docs/design/architecture/naming.qmd
lwjohnst86: Both! A gzip is a compression algorithm. You can compress anything. So it will be `.csv.gz`. Originally it was going to be the file format they uploaded in, but with some changes we've made, it seems we've settled on `csv` for all raw files.

lwjohnst86, line 58:
```suggestion
| raw data file | A compressed raw data file format, such as a `.csv.gzip` file. |
```

docs/design/architecture/naming.qmd
lwjohnst86: Yea, after we've decided on this scheme, we can rename what we've already created to match the intent. But we want it to write to file to match the intent and how it is used with `create_` and many other `create_` functions.

docs/design/architecture/naming.qmd
lwjohnst86: Hmmm, good question... :thinking: will have to think how that works with `update_version()`. Still early days for this though, so maybe for now, we say yes, it doesn't write to files.

signekb (approved):

PR #1071, opened 2025-02-20
docs: :memo: rename functions to `update_` rather than `edit_`

lwjohnst86:
## Description

Related to adding the `version` object in PR #1070 which could have an `update` action, I thought that renaming `edit_` to `update_` so we don't have two different action names for the same thing.

### DIFF
--- a/docs/design/interface/functions.qmd
+++ b/docs/design/interface/functions.qmd
@@ -27,23 +27,6 @@ are the core external-facing functions in Sprout. See the
 [Outputs](outputs.qmd) section for an overview and explanation of the
 different outputs provided by Sprout.
 
-There are some small differences between the naming scheme and the
-functions described here:
-
-1.  `edit` only ever edits the `datapackage.json` which contains the
-    properties/metadata of the package or resource(s) but never edits
-    the data itself. If there is a need for updates to the data itself,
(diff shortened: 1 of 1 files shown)


martonvago (approved):

docs/design/interface/functions.qmd
```diff
@@ -81,11 +64,22 @@ Using a template, this will build a README file with the contents of the
...
+### {{< var done >}} `update_package_properties(path, properties)`
 
-See the help documentation with `help(edit_package_properties())` for
+See the help documentation with `help(update_package_properties())` for
 more details.
 
+```{mermaid}
+flowchart
+    in_path[/path/]
+    in_properties[/properties/]
```
signekb, line 75:
```suggestion
    in_properties[/package_properties/]
```
To fit the resource_properties below. Alternatively, they could be written as their classes `PackageProperties` and `ResourceProperties`. Then, the path could also be `Path`

signekb (changes requested):
Nice 👍  Just a small comment.

docs/design/interface/functions.qmd
lwjohnst86: Good idea, I revise them in another PR and not commit these suggestions, since I'd also have to update the arguments above.

PR #1074, opened 2025-02-20
docs: :building_construction: revise to `build_resource_data()` to output a DataFrame

lwjohnst86:
## Description

After seeing @K-Beicher working with the data and making the guide, it felt off to immediately save to Parquet. Given this is `core` and developer-focused/highly flexible, I can imagine situations where there may be some post-processing before generating to the Parquet file. So I've modified this function to output to a DataFrame rather than save directly to Parquet.


docs/design/interface/pseudocode/build_resource_data.py
martonvago, line 1: Broadly, this makes sense, but this seems like a good place to ask:
At which point will the data frame be ready for conversion to parquet?
Now ignoring any use-case specific custom transformations, for me this comes down to parquet data types. 
How do we want to represent fields with special format constraints like `duration` or `datetime` in parquet?
Will we ever want to query them (e.g. sort by `duration`, select data with `duration` in the top 25% etc.) or summarise them (e.g. show average `duration`)? If so, leaving them as strings makes this more difficult. But if we convert them to a parquet datatype (e.g. integer value for the number of seconds in a `duration`), it's not guaranteed we won't have to cut corners.

Depending on the answer, we'll just have to make sure this function and `check_data` don't repeat the same work. `check_data` already does the bulk of creating a Frictionless-compliant data frame from a CSV, so we could pull that logic out into a separate function.
One consideration is that we don't want to iterate through all rows of the data multiple times if we can avoid it, as presumably that will be our performance bottleneck? So it would be nice if data was checked only once within one workflow.

lwjohnst86: I'm not completely sure of your question here.

To address some of the comments though:

- We won't ever need to query the Parquet file. That would be the user at the end when they want to analyze things.
- We want to store as specific datatypes as possible into the Parquet file. It makes no sense (to me) that we store things that are dates, etc as strings when Parquet has a rich feature set for data types. Part of the struggle in the analysis stage is converting columns to the correct format without messing things up.
- Performance isn't really something we should be concerned with right now. These checks or any other intense work that happens without Sprout to store/structure the data is a very very small part of the whole lifecycle of data use and happens only occasionally. Maybe at some point in the future we think about performance, but not at all right now. Getting it right and getting it high quality is our biggest priority. So if we end up repeating work a bit, that is really not a big problem right now. We can do profiling later on to identify the actual bottlenecks in the code, we may be surprised where those bottlenecks may actually be! :stuck_out_tongue:

martonvago: You pretty much answered my question! :grin: 

- Re querying: I had some vague idea that we will eventually want to support generating data extracts. Not do the data extraction ourselves, but generate some sort of file that specifies the subset of the data to be extracted. 
- Re people analysing their own data: ah, so they would be analyzing the parquet data! I guess that was obvious, sorry :sweat_smile: ! 

Okay, so we want specific data types in parquet.
Then a decent piece of work will be mapping Frictionless data types to parquet data types (without caring about Polars data types).

Would we want the mapping to go both ways? 
E.g., if `duration` is mapped [formatted string] >> [integer for number of seconds], will we ever want to undo this conversion and go [integer for number of seconds] >> [formatted string]?

lwjohnst86: Hmm, true, we did talk about generating some data extracts. Yea, I think that idea may be on hold for a long time, mainly to keep some things easy and not get into privacy/legal issues when we work on Flower.

I can't yet envision why we might map both ways, so maybe we'll have to see as we go along!

martonvago: I guess we might want to map both ways if we wanted to export the parquet file as something Frictionless-compliant. Whether CSV or parquet, if we export with integers in a `duration` field, that will not match the Data Package standard.

lwjohnst86: Hmm good point. Maybe we deal with that when we get to it. There's a lot of data types and most of the are not very commonly used, so it might be a non-issue.

docs/design/interface/pseudocode/build_resource_data.py
```diff
@@ -49,11 +51,7 @@ def build_resource_parquet(
     # This function could be several functions or the one full function.
     check_data(data, resource_properties)
```
lwjohnst86, line 52: @martonvago this is how I see `check_data()` being used, based on my comment in #1060

docs/design/interface/pseudocode/build_resource_data.py
```diff
@@ -1,20 +1,22 @@
...
-    that the files stored in the `resources/raw/` folder are already correctly
-    structured and tidy, it still runs checks to ensure the data are correct
-    by comparing to the properties. All data in the
-    `resources/raw/` folder will be merged into one single data object and then
-    written back to the Parquet file. The Parquet file will be overwritten.
-
-    If there are any duplicate observation units in the data, only the most recent
-    observation unit will be kept. This way, if there are any errors or mistakes
+) -> DataFrame:
+    """Merge all raw resource file(s) into a single DataFrame file.
```
signekb, line 5:
```suggestion
    """Merge all raw resource file(s) into a single DataFrame.
```

docs/design/interface/pseudocode/build_resource_data.py
```diff
@@ -1,20 +1,22 @@
...
+
+    This function takes the file(s) given by `raw_files_path`, merges them
+    together, does some checks and minor processing before outputting them as a
+    single DataFrame. The `resource_properties` object is used to check the data
+    and ensure it is correct. This function can be used to apply any additional
```
signekb, line 10: This sounds like it's this function that does some additional processing to the data before saving it to Parquet? Is that intentional?

docs/design/interface/pseudocode/build_resource_data.py
```diff
@@ -1,20 +1,22 @@
...
+    processing to the data before saving it as a Parquet file.  While Sprout
+    generally assumes that the files stored in the `resources/<id>/raw/` folder
+    are already correctly structured and tidy, it still runs checks to ensure
```
signekb, line 13:
```suggestion
    are already correctly structured and tidy, this function still runs checks to ensure
```

signekb (changes requested):
Nice work 👍  Just some minor docstring suggestions and a question about the `DataFrame`.

docs/design/interface/functions.qmd
```diff
@@ -127,17 +127,17 @@ flowchart
...
 details.
 
 ```{mermaid}
 flowchart
     in_raw_files_path[/raw_files_path/]
     in_properties[/resource_properties/]
-    function("build_resource_parquet()")
-    out[("./resources/{id}/data.parquet")]
+    function("build_resource_data()")
+    out[("DataFrame")]
```
lwjohnst86, line 140:
```suggestion
    out[("Polars<br>DataFrame")]
```

docs/design/interface/pseudocode/build_resource_data.py
lwjohnst86, line 10:
```suggestion
    and ensure it is correct. This function doesn't write to Parquet, so that you can apply any additional
```

lwjohnst86: I clarified the sentence.

docs/design/interface/pseudocode/build_resource_data.py
```diff
@@ -1,20 +1,22 @@
...
-    written back to the Parquet file. The Parquet file will be overwritten.
-
-    If there are any duplicate observation units in the data, only the most recent
-    observation unit will be kept. This way, if there are any errors or mistakes
+) -> DataFrame:
+    """Merge all raw resource file(s) into a single DataFrame file.
+
+    This function takes the file(s) given by `raw_files_path`, merges them
+    together, does some checks and minor processing before outputting them as a
+    single DataFrame. The `resource_properties` object is used to check the data
```
lwjohnst86, line 9:
```suggestion
    single (Polars) DataFrame. The `resource_properties` object is used to check the data
```

docs/design/interface/pseudocode/build_resource_data.py
lwjohnst86: Yea, since again, we can't be 100% certain the data won't be modified outside of Python/Sprout.

martonvago (commented):
I'm very happy to approve, just one last question:
In light of the earlier discussion on this PR, does this function convert the data to non-string datatypes? I don't mind if it happens here or in another function, but just so it is clear when we come back to implement this.

lwjohnst86:
@martonvago hmmm, I don't know what that workflow looks like regarding this data type issue, so I don't really know. So I guess, we see as we go along? :smile: :stuck_out_tongue:

PR #1075, opened 2025-02-20
docs: :building_construction: rename to `_dirs` rather than `_structure`

lwjohnst86:
## Description

Since removing `create_package_structure()`, it makes more sense to rename this function to simply `create_resource_dirs()`, since that's all it is doing.

This PR needs a quick review.


martonvago (approved):

signekb (approved):
I like it 👍

PR #1077, opened 2025-02-20
docs: :building_construction: pseudocode for `write_resource_parquet()`

lwjohnst86:
## Description

Building off of #1074, this is the function for writing to the resource's Parquet file.


docs/design/interface/pseudocode/write_resource_parquet.py
martonvago, line 1: Makes sense, just the same [question](https://github.com/seedcase-project/seedcase-sprout/pull/1074/files#r1965201353) as in the other PR.

We'll want to avoid calling `check_data` too often, and when we can call it will depend, in any case, on the final shape of the data frame that is ready for conversion to parquet.

docs/design/interface/pseudocode/write_resource_parquet.py
```diff
@@ -0,0 +1,42 @@
+# ruff: noqa
+def write_resource_parquet(
+    data: DataFrame, resource_properties: ResourceProperties
+) -> Path:
+    """Check and write the resource data into a Parquet file.
+
+    This function takes the `data` built from `build_resource_data()`, checks it
+    against the `resource_properties`, and then writes the data to the resources
+    `data.parquet` file .  The Parquet file is saved based on the path found in
+    `ResourceProperties.path` and it always overwritten.  Before writing, this
```
signekb, line 10:
```suggestion
    `ResourceProperties.path` and is always overwritten.  Before writing, this
```

docs/design/interface/pseudocode/write_resource_parquet.py
```diff
@@ -0,0 +1,42 @@
...
+
+    Examples:
+
+        ``` python
+        import seedcase_sprout.core as sp
+
+        data = sp.build_resource_data(
+            raw_files_path=sp.path_resources_raw_files(1),
+            resource_properties=sp.example_resource_properties,
+        )
```
signekb, line 22: Wouldn't we expect users to read the resource properties from `datapackage.json`? We could have one example `datapackage.json` to read the example resource properties from instead?

docs/design/interface/pseudocode/write_resource_parquet.py
```diff
@@ -0,0 +1,42 @@
...
+        ```
+
+    Args:
+        data: A DataFrame object with the resources data from the files in its
+            `raw/` folder.
+        resource_properties: The `ResourceProperties` object that contains the
+            properties of the resource you want to create the Parquet file for.
+
+    Returns:
+        Outputs the path object of the created Parquet file.
```
signekb, line 34:
```suggestion
        Outputs the path of the created Parquet file.
```

signekb (changes requested):
This overall makes sense to me :) Just some text suggestions and a question.

docs/design/interface/pseudocode/write_resource_parquet.py
lwjohnst86: Ooo I think I misunderstood what you were asking. You are referring to the `example_resource_properties`? Yea, we can totally change that later.

PR #1100, opened 2025-02-28
feat: :sparkles: add `path_readme()`

signekb:
## Description

This PR adds a `path_readme()` which was referred to in a TODO item in #810 
This function will probably have to be rewritten a bit when we fully switch to the “local-first” approach, but I thought it made sense to add it now, so we don’t forget about it. 

This PR needs a quick review.

## Checklist

- [X] Added or updated tests
- [X] Ran `just run-all`


src/seedcase_sprout/core/__init__.py
```diff
@@ -88,9 +89,10 @@
     # "delete_resource_data",
     # "delete_resource_properties",
     # Path -----
-    "path_package",
-    "path_properties",
-    "path_packages",
+    path_package,
+    path_packages,
+    path_properties,
+    path_readme,
```
lwjohnst86, line 95:
```suggestion
    "path_package",
    "path_packages",
    "path_properties",
    "path_readme",
```

They need to be in quotes.

lwjohnst86 (changes requested):
:+1: just one fix

lwjohnst86 (approved):

PR #1111, opened 2025-03-06
refactor: :recycle: check column data types

martonvago:
## Description
Stacked PR [1/2]

This PR adds functions for checking the data type of columns in Polars data frames. More precisely, these functions check if all string values in a column can be converted to their target (Polars and Parquet) data type. They replace Pandera checks, so I deleted those.

The functions make more sense as part of the [full check pipeline](https://github.com/seedcase-project/seedcase-sprout/blob/feat/check-df-data-types/src/seedcase_sprout/core/sprout_checks/check_data_types.py) -- in the [next PR](#1125 ).

## Checklist

- [x] Added or updated tests
- [x] Updated documentation
- [x] Ran `just run-all`


src/seedcase_sprout/core/sprout_checks/check_column_data_types.py
martonvago, line 1: The idea behind all of these is the following:
Many Polars functions for casting and parsing dates/times have a `strict` parameter. When this is set to `False`, any values that fail during the operation will be set to null. This gives us a way of running a check on all values in a column (without terminating early due to an error) and marking incorrect values with null. The [context](https://github.com/seedcase-project/seedcase-sprout/blob/4e12cf3ef67f8c497943018736c51f1704ed0de1/src/seedcase_sprout/core/sprout_checks/check_data_types.py#L56) for these functions will show that "marking with null" happens in another column.

So all check expressions were written to return null for incorrect values and return non-null for correct values.

src/seedcase_sprout/core/sprout_checks/check_column_data_types.py
```diff
@@ -0,0 +1,222 @@
...
+        field_name: The name of the column to check.
+
+    Returns:
+        A Polars expression for checking the column.
+    """
+    return (
+        pl.when(pl.col(field_name).is_null() | pl.col(field_name).is_in(BOOLEAN_VALUES))
+        .then(pl.lit(""))
+        .otherwise(pl.lit(None))
+    )
```
martonvago, line 27: Like doing:
```python
if value is None or value in BOOLEAN_VALUES:
    return ""
return None
```
What I return as a non-null value doesn't matter in this setup, I won't use it for anything.

src/seedcase_sprout/core/sprout_checks/check_column_data_types.py
```diff
@@ -0,0 +1,216 @@
...
+    Args:
+        field_name: The name of the column to check.
+        field_type: The type of the field.
+
+    Returns:
+        A Polars expression for checking the column.
+    """
+    return (
+        pl.col(field_name)
+        .cast(FRICTIONLESS_TO_POLARS[field_type], strict=False)
```
martonvago, line 41: Here and below `strict=False` means: map to `None` if the operation fails instead of throwing an error.

src/seedcase_sprout/core/sprout_checks/check_column_data_types.py
```diff
@@ -0,0 +1,207 @@
...
+        A Polars expression for checking the column.
+    """
+    first_datetime = (
+        data_frame.get_column(field_name)
+        .drop_nulls()
+        .str.to_datetime(strict=False)
+        .first()
+    )
+    has_timezone = bool(first_datetime.tzinfo) if first_datetime else False
+    datetime_format = "%Y-%m-%dT%H:%M:%S%.f" + ("%z" if has_timezone else "")
```
martonvago, line 88: I'm explicitly setting the format because otherwise it passes datetimes that are not in the format specified by Frictionless as well.

src/seedcase_sprout/core/sprout_checks/check_column_data_types.py
```diff
@@ -0,0 +1,207 @@
...
+        .is_not_null()
+    )
+
+
+def check_is_datetime(data_frame: pl.DataFrame, field_name: str) -> pl.Expr:
+    """Checks if the column contains only datetime values.
+
+    Mixing values with and without timezone information is not allowed.
+    Mixing values with different timezones is allowed, as they will be standardised
+    before saving to Parquet.
```
martonvago, line 70: How we handle these cases is very much up for discussion.
Right now:
- Check passes if all datetimes have timezones but the timezones are not all the same. We have to convert them to the same timezone before saving to Parquet.
- Check fails if timezone-aware and timezone-naive datetimes are mixed. The rationale is that we just don't know what timezone timezone-naive datetimes are in. Alternatively, we could pass in this case as well, and force all datetimes to the same timezone regardless before saving to Parquet.

src/seedcase_sprout/core/sprout_checks/check_column_data_types.py
```diff
@@ -0,0 +1,207 @@
...
+    return pl.col(field_name).str.contains(GEOPOINT_PATTERN)
+
+
+def check_is_json(field_name: str, expected_type: type[list | dict]) -> pl.Expr:
+    """Checks if the column contains only JSON values.
+
+    Failed values are marked with False.
+
+    Warning: uses `map_elements` to check the formatting of each value and may run
+    slowly on large datasets.
```
martonvago, line 160: I couldn't find a way of checking JSON using only Polars expressions. There is a `json_decode` expression, but it has no `strict=False` mode, so if it encounters an error it will exit the expression chain.

src/seedcase_sprout/core/sprout_checks/check_column_data_types.py
```diff
@@ -0,0 +1,207 @@
...
+    Returns:
+        True if the value is a correct JSON type, False otherwise.
+    """
+    try:
+        return isinstance(json.loads(value), expected_type)
+    except json.JSONDecodeError:
+        return False
+
+
+FRICTIONLESS_TO_COLUMN_CHECK = {
```
martonvago, line 191: Used [here](https://github.com/seedcase-project/seedcase-sprout/blob/dbd096ca7a8a8bfdb219ca1d18bcb9b0df119589/src/seedcase_sprout/core/sprout_checks/check_data_types.py#L112)

src/seedcase_sprout/core/sprout_checks/check_column_data_types.py
```diff
@@ -0,0 +1,207 @@
...
+    "datetime": check_is_datetime,
+    "date": check_is_date,
+    "time": check_is_time,
+    "geopoint": check_is_geopoint,
+    "array": lambda value: check_is_json(value, list),
+    "object": lambda value: check_is_json(value, dict),
+    "geojson": lambda value: check_is_json(value, dict),
+    "string": lambda _: pl.lit(True),
+    "any": lambda _: pl.lit(True),
+    "duration": lambda _: pl.lit(True),
```
martonvago, line 206: Should I spend time writing a Polars-compatible regex for Frictionless durations?

src/seedcase_sprout/core/sprout_checks/check_column_data_types.py
```diff
@@ -0,0 +1,207 @@
...
+BOOLEAN_VALUES = {"false", "False", "FALSE", "0", "true", "True", "TRUE", "1"}
+
+
+def check_is_boolean(field_name: str) -> pl.Expr:
+    """Checks if the column contains only boolean values.
+
+    Failed values are marked with False.
+
+    Args:
+        field_name: The name of the column to check.
```
lwjohnst86, line 18: Might be good to be consistent with the language (and/or clarify what each means). There's both field and column being mentioned. Maybe you can say "check if field (also called column)..."? Either way, we should be consistent here and also explicit what `field` means in these cases. I prefer `column` because that's clearer, but that's only applicable to tabular data. Which is fine because the code uses Polars which needs a data frame. So maybe rename everything to `column` rather than `field`?

src/seedcase_sprout/core/sprout_checks/check_column_data_types.py
```diff
@@ -0,0 +1,207 @@
...
+
+    Returns:
+        A Polars expression for checking the column.
+    """
+    return pl.col(field_name).is_in(BOOLEAN_VALUES)
+
+
+def check_is_castable_type(field_name: str, field_type: FieldType) -> pl.Expr:
```
lwjohnst86, line 26:
```suggestion
def check_is_given_type(field_name: str, field_type: FieldType) -> pl.Expr:
```

castable isn't super clear what it means. Maybe "given"? But I don't love that either ... :thinking: Or just `check_is_field_type()`

src/seedcase_sprout/core/sprout_checks/check_column_data_types.py
```diff
@@ -0,0 +1,207 @@
...
+    """Checks if the column contains only values of the given type.
+
+    The check is done by attempting to cast to the appropriate Polars data type.
+    Failed values are marked with False.
```
lwjohnst86, line 30:
```suggestion
    The check is done by attempting to convert (cast) the column to the
    appropriate Polars data type. If it fails, the values are marked with 
    False.
```

src/seedcase_sprout/core/sprout_checks/check_column_data_types.py
```diff
@@ -0,0 +1,207 @@
...
+
+    Args:
+        field_name: The name of the column to check.
+        field_type: The type of the field.
```
lwjohnst86, line 34:
```suggestion
        field_name: The name of the field to check.
        field_type: The type of the field.
```

src/seedcase_sprout/core/sprout_checks/check_column_data_types.py
lwjohnst86: Yea, this is tricky. I think failing if there is mixing is a good idea. What about if there is no timezone information at all?

src/seedcase_sprout/core/sprout_checks/check_column_data_types.py
```diff
@@ -0,0 +1,207 @@
...
+
+    Returns:
+        A Polars expression for checking the column.
+    """
+    return (
+        pl.col(field_name)
+        .cast(FRICTIONLESS_TO_POLARS[field_type], strict=False)
```
lwjohnst86, line 41:
```suggestion
        # Strict is false means map to None if it fails rather than give an error.
        .cast(FRICTIONLESS_TO_POLARS[field_type], strict=False)
```

Just to add implementation comments :relaxed:

src/seedcase_sprout/core/sprout_checks/check_column_data_types.py
```diff
@@ -0,0 +1,207 @@
...
+    Args:
+        field_name: The name of the column to check.
+
+    Returns:
+        A Polars expression for checking the column.
+    """
+    return (
+        pl.col(field_name).str.starts_with("-").not_()
+        & (pl.col(field_name) + "-01")
+        .str.to_date(format="%Y-%m-%d", strict=False)
```
lwjohnst86, line 60:
```suggestion
        # Strict is false means map to None if it fails rather than give an error.
        .str.to_date(format="%Y-%m-%d", strict=False)
```

src/seedcase_sprout/core/sprout_checks/check_column_data_types.py
```diff
@@ -0,0 +1,207 @@
...
+        data_frame: The data frame being operated on.
+        field_name: The name of the column to check.
+
+    Returns:
+        A Polars expression for checking the column.
+    """
+    first_datetime = (
+        data_frame.get_column(field_name)
+        .drop_nulls()
+        .str.to_datetime(strict=False)
```
lwjohnst86, line 84:
```suggestion
        # Strict is false means map to None if it fails rather than give an error.
        .str.to_datetime(strict=False)
```

src/seedcase_sprout/core/sprout_checks/check_column_data_types.py
```diff
@@ -0,0 +1,207 @@
...
+    Args:
+        data_frame: The data frame being operated on.
+        field_name: The name of the column to check.
+
+    Returns:
+        A Polars expression for checking the column.
+    """
+    first_datetime = (
+        data_frame.get_column(field_name)
+        .drop_nulls()
```
lwjohnst86, line 83: Why drop nulls at this point?

src/seedcase_sprout/core/sprout_checks/check_column_data_types.py
lwjohnst86: Does this line not add timezone if it isn't found? via the `else ""`?

src/seedcase_sprout/core/sprout_checks/check_column_data_types.py
```diff
@@ -0,0 +1,207 @@
...
+        .str.to_datetime(strict=False)
+        .first()
```
lwjohnst86, line 85: first is taking the first value? Is this to just get to first datetime in the column?

src/seedcase_sprout/core/sprout_checks/check_column_data_types.py
```diff
@@ -0,0 +1,207 @@
...
+    )
+    has_timezone = bool(first_datetime.tzinfo) if first_datetime else False
+    datetime_format = "%Y-%m-%dT%H:%M:%S%.f" + ("%z" if has_timezone else "")
+
+    return (
+        pl.col(field_name).str.starts_with("-").not_()
```
lwjohnst86, line 91:
```suggestion
        # Drop field values that start with `-` negative.
        pl.col(field_name).str.starts_with("-").not_()
```

This is right?

src/seedcase_sprout/core/sprout_checks/check_column_data_types.py
lwjohnst86, line 1: Why did you rename it to `column_data_types`?

Might be good to rename all `field` into `column` below (See some of my other comments about it below). To be consistent.

src/seedcase_sprout/core/sprout_checks/check_column_data_types.py
```diff
@@ -0,0 +1,207 @@
...
+    return (
+        pl.col(field_name).str.to_time(format="%H:%M:%S%.f", strict=False).is_not_null()
+    )
+
+
+# https://stackoverflow.com/a/18690202
+GEOPOINT_PATTERN = (
+    r"^(?:[-+]?(?:[1-8]?\d(?:\.\d+)?|90(?:\.0+)?)),\s*"
+    r"(?:[-+]?(?:180(?:\.0+)?|(?:1[0-7]\d|[1-9]?\d)(?:\.\d+)?))$"
+)
```
lwjohnst86, line 137: Nice! haha

src/seedcase_sprout/core/sprout_checks/check_column_data_types.py
lwjohnst86: That's fine, we don't really need to support that for now.

src/seedcase_sprout/core/sprout_checks/check_column_data_types.py
lwjohnst86: No hahah

tests/core/sprout_checks/test_check_column_data_types.py
```diff
@@ -0,0 +1,262 @@
...
+        .get_column("index")
+        .to_list()
+    )
+
+    assert fails == expected_fails
+
+
+# Datetime
+DATETIME_BAD_VALUES_WHEN_TIMEZONE = [
+    "",
```
lwjohnst86, line 174: What about when the first value is a correct timezone value? But the rest aren't?

tests/core/sprout_checks/test_check_column_data_types.py
```diff
@@ -0,0 +1,262 @@
...
+    "2002-10-10T12:00:00+01:00",
+    "2002-10-10T12:00:40-05:00",
+    "2002-10-10T12:00:00.34Z",
+    "2002-10-10T12:40:00.34-04:30",
+    "2002-10-10T17:00:55Z",
+    "0000-01-01T00:00:00Z",
+]
+
+DATETIME_BAD_VALUES_WHEN_NO_TIMEZONE = [
+    "",
```
lwjohnst86, line 200: Same comment as above. Also, what if the first value has a timezone, but the rest don't?

lwjohnst86 (changes requested):
:tada: just some comments :star:

tests/core/sprout_checks/test_check_column_data_types.py
martonvago: If the first value has no timezone information (= timezone-naive), then we will expect the rest of the values not to have a timezone either and check accordingly. So we will fail any value that is timezone-aware or just a badly formatted datetime.

As above, if the first value does have timezone information (= timezone-aware), then we will expect the rest of the values to have a timezone as well. So we will fail any value that is timezone-naive or just a badly formatted datetime.

lwjohnst86 (dismissed):
Looks good! :mechanical_arm:

signekb (dismissed):

PR #1121, opened 2025-03-06
docs: :memo: update example in `create_resource_structure()`

lwjohnst86:
## Description

Updated the example code to run in the `create_resource_structure()` function.

This PR needs a quick review.

## Checklist

- [x] Updated documentation
- [x] Ran `just run-all`

### DIFF
--- a/src/seedcase_sprout/core/create_resource_structure.py
+++ b/src/seedcase_sprout/core/create_resource_structure.py
@@ -23,6 +23,7 @@ def create_resource_structure(path: Path) -> list[Path]:
 
     Returns:
        A list of the two created directories:
+
           - A path to the resource directory and
           - A path to the raw data directory.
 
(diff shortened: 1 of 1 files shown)


src/seedcase_sprout/core/create_resource_structure.py
```diff
@@ -23,33 +23,35 @@ def create_resource_structure(path: Path) -> list[Path]:
...
+            # Create a package structure first
+            sp.create_package_properties(
+                properties=sp.example_package_properties(),
+                path=temp_path
+            )
+
+            resource_path = Path(temp_path / "resources")
+            resource_path.mkdir()
+            # Create a resource structure
+            sp.create_resource_structure(path=resource_path)
```
martonvago, line 53:
```suggestion
            resources_path = Path(temp_path / "resources")
            resources_path.mkdir()
            # Create a resource structure
            sp.create_resource_structure(path=resources_path)
```
To avoid confusion between the path to the "resources" folder and the path to a resource

martonvago (changes requested):
Nice!! :smile:

src/seedcase_sprout/core/create_resource_structure.py
lwjohnst86: Nice catch!

martonvago (approved):

PR #1126, opened 2025-03-10
docs: :memo: add supported formats doc to design interface

### LINKED ISSUE #1018: Include a doc with the list of supported input data formats in interface design
What raw data formats we are currently able to extract properties from.

signekb:
## Description

This PR adds a doc on which data formats we support property extraction for. 

Closes #1018 

This PR needs a quick review.

## Checklist

- [X] Updated documentation
- [X] Ran `just run-all`

### DIFF
--- a/docs/design/interface/supported-formats.qmd
+++ b/docs/design/interface/supported-formats.qmd
@@ -1,0 +1,18 @@
+---
+title: "Supported data formats"
+---
+
+Currently, Sprout supports the extraction of properties from the
+following data formats:
+
+```{python}
+#| echo: false
+from seedcase_sprout.core.check_is_supported_format import SUPPORTED_FORMATS
+
+for format in sorted(SUPPORTED_FORMATS):
+    print(format)
+```
+
+For more information on how to extract properties from these formats,
+see the [Creating and managing data
+resources](/docs/guides/resources.qmd) guide.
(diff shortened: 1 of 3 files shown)


docs/design/interface/supported-formats.qmd
lwjohnst86, line 1: This could be renamed to "inputs", which will match the similar one in architecture as well as the other interface `outputs.qmd` one. And it will allow us to add more things to this file than "just" the supported data formats.

docs/design/interface/supported-formats.qmd
```diff
@@ -0,0 +1,18 @@
...
+
+Currently, Sprout supports the extraction of properties from the
+following data formats:
+
+```{python}
+#| echo: false
+from seedcase_sprout.core.check_is_supported_format import SUPPORTED_FORMATS
+
+for format in sorted(SUPPORTED_FORMATS):
+    print(format)
```
lwjohnst86, line 13:
```suggestion
#| output: asis
from seedcase_sprout.core.check_is_supported_format import SUPPORTED_FORMATS

print("`." + "`, `.".join(sorted(SUPPORTED_FORMATS)) + "`")
```

Outputs this as Markdown. Uses the vectorised `join()` rather than use a for loop, so that the output is a comma separated sequence of file extensions. Outputs to look like this (but then rendered to HTML via the `output: asis` option.

``` markdown
`.csv`, `.parquet`, ...
```

lwjohnst86 (changes requested):

lwjohnst86:
It seems that the supported formats file wasn't renamed, so there's two files that were merged in

PR #1129, opened 2025-03-11
feat: :sparkles: add `read_csv()` to load CSV as DataFrames

martonvago:
## Description

This PR adds a function to load a CSV file into a Polars data frame.

I expect this to be used used when checking data and also when converting to Parquet possibly.
In the check flow, it will look something like:
```python
def check_data(data_path: Path, resource_properties: ResourceProperties) -> Path:
    df = read_csv(data_path)  # <---
    check_data_header(df, resource_properties)
    df = set_missing_values_to_null(df, resource_properties)
    check_data_types(df, resource_properties)
    return data_path
```

## Checklist

- [x] Added or updated tests
- [x] Updated documentation
- [x] Ran `just run-all`

### DIFF
--- a/src/seedcase_sprout/core/read_csv.py
+++ b/src/seedcase_sprout/core/read_csv.py
@@ -30,7 +30,9 @@ def read_csv(data_path: Path) -> pl.DataFrame:
         return pl.read_csv(
             data_path,
             has_header=True,
+            # Doesn't work all the time, and we want to use properties anyway.
             infer_schema=False,
+            # Set as empty string since it will be easier to match with Frictionless.
             missing_utf8_is_empty_string=True,
         )
     except pl.exceptions.ComputeError as error:
--- a/src/seedcase_sprout/core/read_csv.py
+++ b/src/seedcase_sprout/core/read_csv.py
@@ -38,6 +38,6 @@ def read_csv(data_path: Path) -> pl.DataFrame:
             raise ValueError(
                 f"At least one row in the data file at {data_path} contains more items "
                 "than the expected number of columns. Please make sure that all rows "
-                "are the same length."
+                “have the same number of items."
             ) from error
         raise
--- a/src/seedcase_sprout/core/read_csv.py
+++ b/src/seedcase_sprout/core/read_csv.py
@@ -38,6 +38,6 @@ def read_csv(data_path: Path) -> pl.DataFrame:
             raise ValueError(
                 f"At least one row in the data file at {data_path} contains more items "
                 "than the expected number of columns. Please make sure that all rows "
-                “have the same number of items."
+                "have the same number of items."
             ) from error
         raise
--- a/src/seedcase_sprout/core/read_csv.py
+++ b/src/seedcase_sprout/core/read_csv.py
@@ -1,0 +1,43 @@
+from pathlib import Path
+
+import polars as pl
+
+from seedcase_sprout.core.check_is_file import check_is_file
+
+
+def read_csv(data_path: Path) -> pl.DataFrame:
+    """Loads data in the given CSV file into a Polars data frame.
+
+    Args:
+        data_path: The path to the CSV file.
+
+    Returns:
+        The Polars data frame.
+
+    Raises:
+        FileNotFoundError: If the file cannot be found or the path doesn't point to
+            a file.
+        ValueError: If the file is not a CSV file.
+        pl.exceptions.NoDataError: If the file is empty.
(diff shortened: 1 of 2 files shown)


src/seedcase_sprout/core/read_csv.py
```diff
@@ -0,0 +1,43 @@
...
+    if data_path.suffix != ".csv":
+        raise ValueError(f"{data_path} is not a CSV file.")
+
+    try:
+        return pl.read_csv(
+            data_path,
+            has_header=True,
+            infer_schema=False,
+            missing_utf8_is_empty_string=True,
+        )
```
martonvago, line 35: - We expect there to be a header because that is the default table dialect setting in Frictionless. If there is no header or the header is wrong, that will be caught be the `check_data_header` function.
- We don't infer the schema because it cannot infer it 100% correctly, and we want to construct it based on the properties anyway.
- Missing values are represented by the empty string instead of null. Frictionless allows custom missing values -- we have to read these from the properties and set any value listed as missing to null ourselves. If we let Polars set values it considers missing to null when loading the data, we wouldn't be able to do this.

src/seedcase_sprout/core/read_csv.py
```diff
@@ -0,0 +1,43 @@
...
+    except pl.exceptions.ComputeError as error:
+        if "found more fields than defined" in str(error):
+            raise ValueError(
+                f"At least one row in the data file at {data_path} contains more items "
+                "than the expected number of columns. Please make sure that all rows "
+                "are the same length."
+            ) from error
+        raise
```
martonvago, line 43: This entire ugly bit is about remapping that particular `ComputeError` to a `ValueError`. We can do without it, but the `ComputeError` says `found more fields than defined in 'Schema'\n\nConsider setting 'truncate_ragged_lines=True'.`, which is even less informative than my version.

src/seedcase_sprout/core/read_csv.py
```diff
@@ -0,0 +1,43 @@
...
+            has_header=True,
+            infer_schema=False,
+            missing_utf8_is_empty_string=True,
+        )
+    except pl.exceptions.ComputeError as error:
+        if "found more fields than defined" in str(error):
+            raise ValueError(
+                f"At least one row in the data file at {data_path} contains more items "
+                "than the expected number of columns. Please make sure that all rows "
+                "are the same length."
```
signekb, line 41:
```suggestion
                “have the same number of items."
```
Maybe?

src/seedcase_sprout/core/read_csv.py
signekb: I like it!

tests/core/test_read_csv.py
```diff
@@ -0,0 +1,92 @@
...
+def test_throws_error_if_file_not_found(data_path):
+    """Should throw an error if the file cannot be found."""
+
+    with raises(FileNotFoundError):
+        read_csv(data_path)
+
+
+def test_throws_value_error_if_file_not_csv(tmp_path):
+    """Should throw a ValueError if the file is not a CSV file."""
+    path = tmp_path / "not-data.txt"
```
signekb, line 24:
```suggestion
    path = tmp_path / "data.txt"
```
Why is this called no-data? :)

signekb (dismissed):

signekb (dismissed):
🚀

src/seedcase_sprout/core/read_csv.py
```diff
@@ -0,0 +1,43 @@
...
+    """
+    check_is_file(data_path)
+    if data_path.suffix != ".csv":
+        raise ValueError(f"{data_path} is not a CSV file.")
+
+    try:
+        return pl.read_csv(
+            data_path,
+            has_header=True,
+            infer_schema=False,
```
lwjohnst86, line 33:
```suggestion
            # Doesn't work all the time, and we want to use properties anyway.
            infer_schema=False,
```

src/seedcase_sprout/core/read_csv.py
```diff
@@ -0,0 +1,43 @@
...
+            missing_utf8_is_empty_string=True,
```
lwjohnst86, line 34:
```suggestion
            # Set as empty string since it will be easier to match with Frictionless.
            missing_utf8_is_empty_string=True,
```

lwjohnst86 (changes requested):
Very minor additions of some comments.

lwjohnst86 (approved):

PR #1130, opened 2025-03-11
feat: :sparkles: add `check_data_header()`

martonvago:
## Description

This PR adds a function for checking the header of a data frame against the resource properties.
In the check flow, it will look something like:
```python
data = read_csv(data_path) 
def check_data(data: pl.DataFrame, resource_properties: ResourceProperties) -> pl.DataFrame:
    check_data_header(data, resource_properties) # <---
    data = set_missing_values_to_null(data, resource_properties)
    check_data_types(data, resource_properties)
    return data
```

## Checklist

- [x] Added or updated tests
- [x] Updated documentation
- [x] Ran `just run-all`

### DIFF
--- a/src/seedcase_sprout/core/sprout_checks/check_data_header.py
+++ b/src/seedcase_sprout/core/sprout_checks/check_data_header.py
@@ -5,12 +5,12 @@ from seedcase_sprout.core.properties import ResourceProperties
 
 
 def check_data_header(
-    data_frame: pl.DataFrame, resource_properties: ResourceProperties
+    data: pl.DataFrame, resource_properties: ResourceProperties
 ) -> pl.DataFrame:
     """Checks that column names in the data frame match column names in the properties.
 
     Args:
-        data_frame: The data frame to check.
+        data: The data frame to check.
         resource_properties: The resource properties to check against.
 
     Returns:
@@ -19,7 +19,7 @@ def check_data_header(
     Raises:
         ValueError: If the header doesn't match the expected column names.
     """
-    data_columns = data_frame.schema.names()
+    data_columns = data.schema.names()
     expected_columns = [
         field.name
         for field in get_nested_attr(resource_properties, "schema.fields", default=[])
@@ -29,4 +29,4 @@ def check_data_header(
             "Column names in the data frame do not match column names in the resource "
             f"properties. Expected {expected_columns}, but found {data_columns}."
         )
-    return data_frame
+    return data
--- a/src/seedcase_sprout/core/sprout_checks/check_data_header.py
+++ b/src/seedcase_sprout/core/sprout_checks/check_data_header.py
@@ -1,0 +1,32 @@
+import polars as pl
+
+from seedcase_sprout.core.get_nested_attr import get_nested_attr
+from seedcase_sprout.core.properties import ResourceProperties
+
+
+def check_data_header(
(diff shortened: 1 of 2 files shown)


tests/core/sprout_checks/test_check_data_header.py
```diff
@@ -0,0 +1,74 @@
...
+    return ResourceProperties(
+        name="data",
+        title="data",
+        path=str(Path("resources", "1", "data.csv")),
+        description="My data...",
+        schema=TableSchemaProperties(),
+    )
+
+
+def test_no_error_when_data_frame_and_properties_both_empty():
```
martonvago, line 29: In practice, this situation will not happen in the normal check flow because an empty data frame or resource properties will not make it this far.

lwjohnst86:
@martonvago small comment, the `check_data()` function won't have the first arg as data. I noticed that in my pseudocode that was an mistake, since it was based on the assumption of using Frictionless `validate()` which seemed to require a file. But we don't use that anymore.

src/seedcase_sprout/core/sprout_checks/check_data_header.py
```diff
@@ -0,0 +1,32 @@
+import polars as pl
+
+from seedcase_sprout.core.get_nested_attr import get_nested_attr
+from seedcase_sprout.core.properties import ResourceProperties
+
+
+def check_data_header(
+    data_frame: pl.DataFrame, resource_properties: ResourceProperties
```
lwjohnst86, line 8:
```suggestion
    data: pl.DataFrame, resource_properties: ResourceProperties
```

To match what the argument is called in other functions.

src/seedcase_sprout/core/sprout_checks/check_data_header.py
```diff
@@ -0,0 +1,32 @@
...
+) -> pl.DataFrame:
+    """Checks that column names in the data frame match column names in the properties.
+
+    Args:
+        data_frame: The data frame to check.
```
lwjohnst86, line 13:
```suggestion
        data: The data frame to check.
```

src/seedcase_sprout/core/sprout_checks/check_data_header.py
```diff
@@ -0,0 +1,32 @@
...
+        resource_properties: The resource properties to check against.
+
+    Returns:
+        The data frame, if the column names match.
+
+    Raises:
+        ValueError: If the header doesn't match the expected column names.
+    """
+    data_columns = data_frame.schema.names()
```
lwjohnst86, line 22:
```suggestion
    data_columns = data.schema.names()
```

src/seedcase_sprout/core/sprout_checks/check_data_header.py
```diff
@@ -0,0 +1,32 @@
...
+    expected_columns = [
+        field.name
+        for field in get_nested_attr(resource_properties, "schema.fields", default=[])
+    ]
+    if data_columns != expected_columns:
+        raise ValueError(
+            "Column names in the data frame do not match column names in the resource "
+            f"properties. Expected {expected_columns}, but found {data_columns}."
+        )
+    return data_frame
```
lwjohnst86, line 32:
```suggestion
    return data
```

lwjohnst86 (changes requested):
Very minor changes! Otherwise good!

lwjohnst86 (approved):

PR #1138, opened 2025-03-17
fix: :bug: update nested objects correctly

### LINKED ISSUE #1137: [bug]: The dictionary `update` method doesn't work (well) with nested structures
Turns out that `update` works like:
```python
old = {"level1": {"level2": {"a": 1, "b": 1}}}
changes = {"level1": {"level2": {"a": 99}}}
old.update(changes)
# >> {'level1': {'level2': {'a': 99}}}
# and not {'level1': {'level2': {'a': 99, 'b': 1}}}
```
This is not ideal because we want to keep unchanged values around. (I thought we tested this, but apparently not.)

Instead, we could use something like [deepmerge](https://github.com/toumorokoshi/deepmerge)? This package is not as widely used as some of the others, but it does one thing: it updates nested dictionaries. You can even customise how it should handle different types -- would be useful for our Properties objects.

martonvago:
## Description

This PR adds a dependency and a function to handle updating nested objects (see issue for details).
I wrote it so that it can handle both dictionaries and properties classes.

Closes #1137

## Checklist

- [x] Added or updated tests
- [x] Updated documentation
- [x] Ran `just run-all`


src/seedcase_sprout/core/deep_update.py
```diff
@@ -0,0 +1,64 @@
...
+    Returns:
+        A new properties object with the updates.
+    """
+    return old.from_dict(merger.merge(old.compact_dict, updates.compact_dict))
+
+
+DEEP_MERGER = Merger(
+    # Merge strategies for types
+    [
+        (list, ["override"]),
```
lwjohnst86, line 33: Hmmmm, so if someone wanted to update the author list or resource list, they would have to give all the same authors/resources, plus the updated ones? Not sure that's a desirable behaviour. What was your reason for doing "override"?

tests/core/test_deep_update.py
```diff
@@ -0,0 +1,127 @@
...
+
+    assert deep_update(old, updates) == {
+        "level1": {"level2": {"prop1": "new", "prop2": "b"}}
+    }
+
+
+def test_overrides_lists():
+    """Lists should be updated using the 'override' strategy."""
+    old = {"level1": {"level2": [{"prop1": "a", "prop2": "b"}]}}
+    updates = {"level1": {"level2": [{"prop1": "new"}]}}
```
lwjohnst86, line 34: See comment above about this.

tests/core/test_deep_update.py
lwjohnst86, line 1: Can you include a test with a Package that has two Resources, and where one Resource is updated but the other isn't?

Or, where there is a Package, with at least two Contributors, and only one Contributor is modified?

lwjohnst86 (changes requested):
very nice, I have a few comments.

src/seedcase_sprout/core/deep_update.py
martonvago: The reason is that when you have 2 lists, it's not obvious which item corresponds to which item. To update the current version with the updates, you need to be able to find the current version. 

We could come up with a strategy for this, e.g.:
- Items in the current version and items in the updates are matched based on their index in the list.
  - Problems: what if you give the update list the same items as in the current list but in a different order? What if new items are not added to the end of the update list?
- Items in the current version and items in the updates are matched based on a unique attribute, e.g. `name`.
  - Problems: not all properties have a `name` field or a similar ID-like field; this ID-like field would always need to be included in the updates

These are just the most naive ones I could think of though! Once we have a strategy we're happy with, we can add it to the `Merger`.

src/seedcase_sprout/core/deep_update.py
```diff
@@ -0,0 +1,64 @@
...
+    # Fallback strategy for other types
+    ["override"],
+    # Strategy when types conflict
+    ["override"],
+)
+
+PropertiesOrDict = TypeVar("PropertiesOrDict", Properties, dict[str, Any])
+
+
+def deep_update(old: PropertiesOrDict, updates: PropertiesOrDict) -> PropertiesOrDict:
```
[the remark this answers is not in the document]
martonvago: Then let's have "nested"! :tada: 
For me, the key thing about this function is that it deals with nesting (these functions often have "deep" in the name, like in `deepcopy`, to say that they process data below the first layer too). I find "full" and "complete" a bit vague.

src/seedcase_sprout/core/deep_update.py
```diff
@@ -0,0 +1,64 @@
...
+    """Updates an object with values from another object.
```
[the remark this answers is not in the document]
martonvago: For me this is a bit vague/confusing, but I'll have a think about what to add.

tests/core/test_deep_update.py
martonvago: For sure, they will behave like the `FieldProperties` below, because they are lists.
(Re the dicussion above about list strategies: a great unique field for `ResourceProperties` is `name`! This could simplify updating resource properties a lot, potentially!, For `ContributorProperties`, I'm not too sure about a unique field...)

martonvago:
Hmm, I was hoping the file rename would be recognised -- it wasn't 😮‍💨, sorry about that!

tests/core/test_deep_update.py
martonvago: ^^ I'll add these once we've decided how we want it to behave

lwjohnst86 (dismissed):

tests/core/test_nested_update.py
```diff
@@ -0,0 +1,127 @@
...
+    updates = ResourceProperties(
+        schema=TableSchemaProperties(fields=[FieldProperties(name="my_new_field")]),
+    )
+    expected = ResourceProperties(
+        name="old-name",
+        title="Old Title",
+        schema=TableSchemaProperties(
+            fields=[FieldProperties(name="my_new_field")],
+            fields_match="equal",
+        ),
```
signekb, line 99: What if I want to add a field then? Then, I would have to add the old field *and* the new field to the updated properties? *editing* vs *adding*, I guess.

signekb (changes requested):
Nice 👍  Just some questions about the tests.

tests/core/test_nested_update.py
signekb: Oh, I just saw that @lwjohnst86 has similar [questions](https://github.com/seedcase-project/seedcase-sprout/pull/1138#discussion_r1999258575) last week. But, not sure what the conclusion was?

lwjohnst86:
@martonvago can you resolve the conflicts?

PR #1139, opened 2025-03-17
fix: :bug: fix typing of check functions

martonvago:
## Description

This PR fixes the typing of check functions that accept either a dict or a `Properties`.

The problem with the original version was that a return type of `PackageProperties | dict` is too broad. For example, VS code thinks that all methods on dict and all methods on `PackageProperties` are available on the output. But that is not true.
bad

But we can restrict the type of the output because we know that it is the same as the type of the input. Now VS code can infer the methods available on the output correctly:
good

I guess this would have been flagged by mypy?

This PR needs a quick review.

## Checklist

- [x] Added or updated tests
- [x] Updated documentation
- [x] Ran `just run-all`

### DIFF
--- a/src/seedcase_sprout/core/sprout_checks/check_package_properties.py
+++ b/src/seedcase_sprout/core/sprout_checks/check_package_properties.py
@@ -1,14 +1,14 @@
 from seedcase_sprout.core.checks.check_error_matcher import CheckErrorMatcher
-from seedcase_sprout.core.properties import PackageProperties
 from seedcase_sprout.core.sprout_checks.check_properties import (
     RESOURCE_FIELD_PATTERN,
     check_properties,
 )
+from seedcase_sprout.core.sprout_checks.common_types import PackagePropertiesOrDict
 
 
 def check_package_properties(
-    properties: PackageProperties | dict, ignore: list[CheckErrorMatcher] = []
-) -> PackageProperties | dict:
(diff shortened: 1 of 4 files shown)


lwjohnst86 (approved):
Neat solution, though I'm still not super comfortable with the "lets make a new type to solve this issue" and "having multiple types in one argument" approach. But I do think this will be mostly resolved when we split the code out into the `check-datapackage` package.

martonvago:
This particular bit would stay even after splitting out the check package, because it's specific to how the check functions are used in Sprout. I guess an alternative would be to define separate functions for `dict`s and `Properties` classes, so 2x the number of functions. And then use [overloading](https://adamj.eu/tech/2021/05/29/python-type-hints-how-to-use-overload/) to make it more user-friendly.

martonvago:
But then how would you check a dictionary that was just read into Sprout from datapackage.json?

signekb:
Could we align it so we either use only `dict` or `Properties` as input and output in both the `checks` package and `sprout`?

signekb:
Ah, yeah, I see that… Hmmm, and I didn’t really like the using `.compact_dict()` all the time

lwjohnst86:
I've been refactoring the Sprout check functions to match what the design docs say, so this won't be necessary anymore. Will close.

PR #1147, opened 2025-03-18
refactor: :recycle: remove original extension and `.gz` from batch file names

signekb:
## Description

Since we won’t compress these files after all, now that we’ll store batch files as Parquet.

This PR needs a quick review.

## Checklist

- [X] Added or updated tests
- [X] Ran `just run-all`

### DIFF
--- a/src/seedcase_sprout/core/create_batch_file_name.py
+++ b/src/seedcase_sprout/core/create_batch_file_name.py
@@ -11,6 +11,6 @@ def create_batch_file_name(path: Path) -> str:
         path: The path to the batch file to extract the original extension from.
 
     Returns:
-        The created batch file name in the format {timestamp}-{uuid}{ext}.gz
+        The created batch file name in the format {timestamp}-{uuid}{ext}
     """
-    return f"{get_compact_iso_timestamp()}-{uuid4()}{path.suffix}.gz"
+    return f"{get_compact_iso_timestamp()}-{uuid4()}{path.suffix}"
--- a/src/seedcase_sprout/core/create_batch_file_name.py
+++ b/src/seedcase_sprout/core/create_batch_file_name.py
@@ -1,16 +1,15 @@
-from pathlib import Path
 from uuid import uuid4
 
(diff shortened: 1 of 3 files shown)


src/seedcase_sprout/core/create_batch_file_name.py
```diff
@@ -11,6 +11,6 @@ def create_batch_file_name(path: Path) -> str:
         path: The path to the batch file to extract the original extension from.
 
     Returns:
-        The created batch file name in the format {timestamp}-{uuid}{ext}.gz
+        The created batch file name in the format {timestamp}-{uuid}{ext}
```
lwjohnst86, line 14: Hm, I think we can drop the `{ext}` bit, I don't know what value that will provide. Thoughts?

lwjohnst86 (changes requested):

src/seedcase_sprout/core/create_batch_file_name.py
martonvago: Will this file name be passed to the Polars parquet writer function? Then it could have ".parquet" hardcoded in it? Or, yeah, drop the extension here and add it when writing to parquet.

tests/core/test_create_batch_file_name.py
martonvago, line 1: Should we have parquet files in the tests (eventually), just to make it consistent with the design?

src/seedcase_sprout/core/create_batch_file_name.py
signekb: Ah, you’re right! If we’ll always store as parquet, then the original extension doesn’t have to be a part of the batch file name.

signekb: @martonvago I like your second option above; add the extension when writing to parquet. Then, if we’ll change which format we’ll save to (for currently unknown reasons), the we don’t have to change it here.

martonvago (approved):
:tada:

lwjohnst86 (approved):

Issue #1143, opened 2025-03-17
Discussion: Object type for properties to use throughout package


lwjohnst86:
> ... how would you check a dictionary that was just read into Sprout from datapackage.json?

_Originally posted by @martonvago in https://github.com/seedcase-project/seedcase-sprout/issues/1139#issuecomment-2730295408_

A discussion around which object type properties should be and how to go about checking them. Moving the discussion over to here to keep track of it and eventually resolve.

lwjohnst86:
I think the easiest thing to do here is when we read in the properties, convert to a PackageProperties class. So we are always internally consistent that it is a dataclass and not in some cases a dict and others a Properties. I'll have to look into how this might work, but it would make it easier to be consistent and limit the number of objects being used.

martonvago:
Generally, I agree 100% with always using Properties classes to represent properties inside Sprout.
But when properties are read in from `datapackage.json`, the tricky thing is that we don't know anything about them, like if they have the right attributes or the right types. That's why we check them as dictionaries before converting them into Properties objects.

If we convert the contents of `datapackage.json` to `PackageProperties` without checking them
- If an attribute is the wrong type in `datapackage.json`, `from_dict` will throw an error when trying to construct a `PackageProperties`.
  - This can happen when there is a genuine mistake, say, someone edited the file by hand.
  - Or when `datapackage.json` meets the general Frictionless standard, but not Sprout-specific restrictions. E.g., Sprout wants a string for the data `path` of a resource, but Frictionless is happy with a list or a string.
- Any attributes in `datapackage.json` but not in `PackageProperties` are stripped. Maybe good, maybe bad, but it means that Sprout will have no oversight over these.

I'm not sure if we want to expose Sprout-specific check functions from Sprout, but if so, we could expose ones that take a Properties class as argument, and handle running the same checks on dictionary representations internally when reading. Maybe?

lwjohnst86:
Hm, but when we read from `datapackage.json` we use `check-datapackage` functions, which will be `dict` as input. It's not like we lose access to the `check-datapackage` functionality.

martonvago:
That's true, but we have some Sprout-specific checks that we do in addition to the ones in `check-datapackage`.
For example, we check that the data path on resources is a string. If someone puts a list for a data path in `datapackage.json`, this will pass `check-datapackage` checks, but we will get a type error when we try to create a `PackageProperties`.

[...]

PR #1156, opened 2025-03-19
feat: :sparkles: add `read_properties()` function

lwjohnst86:
## Description

When I started refactoring `edit_package_properties()` to `update_package_properties()`, I realized we hadn't yet had a `read_properties()` function. So this adds it.

## Checklist

- [x] Added or updated tests
- [x] Updated documentation
- [x] Ran `just run-all`


docs/design/interface/functions.qmd
```diff
@@ -324,7 +324,18 @@ details on the properties.
...
+```{mermaid}
+flowchart
+    in_path[/path/]
+    function("read_properties()")
+    out[("PackageProperties")]
+    in_path --> function
+    function --> out
+```
 
 Reads the `datapackage.json` file, checks that is correct, and then
```
martonvago, line 340:
```suggestion
Reads the `datapackage.json` file, checks that it is correct, and then
```
Maybe?

src/seedcase_sprout/core/read_properties.py
```diff
@@ -0,0 +1,50 @@
...
+
+from seedcase_sprout.core.check_is_file import check_is_file
+from seedcase_sprout.core.properties import PackageProperties
+from seedcase_sprout.core.read_json import read_json
+
+
+def read_properties(path: Path) -> PackageProperties:
+    """Read in the properties from the `datapackage.json` file.
+
+    Reads the `datapackage.json` file, checks that is correct, and then
```
martonvago, line 11:
```suggestion
    Reads the `datapackage.json` file, checks that it is correct, and then
```

src/seedcase_sprout/core/read_properties.py
```diff
@@ -0,0 +1,50 @@
...
+            temp_path = Path(temp_dir)
+
+            properties_path = Path(temp_path / "datapackage.json")
+            # Create a package structure first
+            sp.write_package_properties(
+                properties=sp.example_package_properties(),
+                path=properties_path
+            )
+
+            # Edit package properties
```
martonvago, line 40:
```suggestion
            # Read package properties
```

src/seedcase_sprout/core/read_properties.py
```diff
@@ -0,0 +1,50 @@
...
+            sp.read_properties(properties_path)
+        ```
+
+    Raises:
+        FileNotFound: If the `datapackage.json` file doesn't exist.
+        JSONDecodeError: If the `datapackage.json` file couldn't be read.
+    """
+    check_is_file(path)
+    properties = read_json(path)
+    return PackageProperties.from_dict(properties)
```
martonvago, line 50: Maybe put `check_properties(properties)` (imported from `sprout_checks`) before building the object, so `from_dict` doesn't break if the properties are incorrect

tests/core/test_read_properties.py
martonvago, line 1: Just a couple more tests for the case when the contents of `datapackage.json` are not correct. [This comment](https://github.com/seedcase-project/seedcase-sprout/issues/1143#issuecomment-2733252935) has some relevant examples, otherwise just based on what `check_properties` checks.

martonvago (changes requested):
Very useful :book: :star:

signekb (changes requested):

src/seedcase_sprout/core/read_properties.py
```diff
@@ -47,4 +51,8 @@ def read_properties(path: Path) -> PackageProperties:
     """
     check_is_file(path)
     properties = read_json(path)
+    if properties.get("resources"):
+        check_properties(properties)
+    else:
+        check_package_properties(properties)
```
martonvago, line 57:
```suggestion
    check_properties(
        properties,
        ignore=[CheckErrorMatcher(validator="required", json_path=r"resources$")],
    )
```
Only because `properties.get` will throw if `properties` is valid JSON but not a `dict`.
So it needs some sort of `isinstance(properties, dict)` check. I will add this check in `check_properties` because it fits nicely in there and then the error thrown will be a nice `CheckError`.

The ignore bit here ignores missing resources (resources = None).
So far, we haven't considered empty resources (resources = []) correct because that's not allowed in Frictionless. But you can ignore that error by adding `CheckErrorMatcher(validator="minItems", json_path=r"resources$")`.

tests/core/test_read_properties.py
```diff
@@ -1,22 +1,43 @@
...
+
+    assert expected_properties == actual_properties
+
+
+def test_reads_when_resource_not_exists(tmp_path):
+    """Should raise an error if a required field is missing among the package
+    properties."""
+
+    expected_properties = example_package_properties()
+    expected_properties.resources = None
```
martonvago, line 35: If you want the same treatment for `resources = []`, could we add that in this test (or a new one)?

martonvago (changes requested):

src/seedcase_sprout/core/read_properties.py
lwjohnst86: I knew you had done a better way, that's much better, thanks!!

tests/core/test_read_properties.py
lwjohnst86: This isn't even allowed from the checks, I can't even write to file if this is `[]`. So this would belong in the `write_` or `check_`.

martonvago (changes requested):

tests/core/test_read_properties.py
```diff
@@ -28,8 +28,7 @@ def test_reads_in_as_package_properties(tmp_path):
 
 
 def test_reads_when_resource_not_exists(tmp_path):
-    """Should raise an error if a required field is missing among the package
-    properties."""
+    """Should give an error the resource doesn't exist."""
```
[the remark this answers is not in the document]
lwjohnst86: Oopsssyyyy :fearful: :stuck_out_tongue:

martonvago (dismissed):
:tada:

PR #1157, opened 2025-03-19
refactor: :recycle: revised to `update_package_properties()`

### LINKED ISSUE #1141: Arguments to `edit_resource_properties()`
> I know that this function doesn't exist yet, I'm just wondering about its arguments.
Here it takes 1) the path to `datapackage.json` and 2) the updates to make to the target resource properties.
I think this is very much what we want, but we also need to be able to tell which resource properties to update. As resource properties are uniquely identified by their ID, we need to know the ID of the target resource properties.
> 
> So either
> 1. Always specify the data path (which has the ID) in the input resource properties representing the updates to make. This could be enforced by raising an error if the path is missing.
> 2. Or have a `resource_id` argument to `edit_resource_properties`.
> 
> The second option seems less confusing, perhaps?

_Originally posted by @martonvago in https://github.com/seedcase-project/seedcase-sprout/pull/810#discussion_r1996870820_

lwjohnst86:
## Description

This PR renames the function `edit_package_properties()` to `update_package_properties()` based on the design changes. It also switches to taking only `PackageProperties`, as `current` and `updates` as arguments.

Closes #1141 

## Checklist

- [x] Added or updated tests
- [x] Updated documentation
- [x] Ran `just run-all`


src/seedcase_sprout/core/update_package_properties.py
```diff
@@ -63,34 +56,30 @@ def edit_package_properties(
...
-    current_properties = read_json(path)
     check_properties(
-        current_properties,
+        current,
         ignore=[CheckErrorMatcher(validator="required")],
     )
 
-    current_properties.update(properties.compact_dict)
+    updated = current.compact_dict
+    updated.update(updates.compact_dict)
```
martonvago, line 78: Hopefully this will gain some elegance with `nested_update` :sweat_smile:

tests/core/test_update_package_properties.py
```diff
@@ -94,43 +84,19 @@ def test_current_resources_not_modified(tmp_path, properties):
             )
         ],
     )
-    properties_path = get_properties_path(tmp_path, current_properties)
```
martonvago, line 97: There's a stray `tmp_path` in the test definition above (I can't comment on it)

martonvago (dismissed):
Looks really nice!

martonvago (dismissed):

src/seedcase_sprout/core/update_package_properties.py
```diff
@@ -1,45 +1,38 @@
...
-
-    Raises:
-        FileNotFound: If the `datapackage.json` file doesn't exist.
-        JSONDecodeError: If the `datapackage.json` file couldn't be read.
-        ExceptionGroup: If there is an error in the current, incoming or resulting
+        The updated package properties as a `PackageProperties` object. Use
+            `write_package_properties()` to save it back to the `datapackage.json`
+            file.
+
+        ExceptionGroup: If there is an error in the current, incoming, or resulting
```
signekb, line 35:
```suggestion
    Raises:
        ExceptionGroup: If there is an error in the current, incoming, or resulting
```

signekb (changes requested):
Nice 👍  Just some minor suggestions.

PR #1159, opened 2025-03-19
docs: :building_construction: add flow diagram for updating the README

lwjohnst86:
## Description

This is the flow for updating the README file whenever there is a change to the `datapackage.json` file. Ignore the intro section, it's copied over from the other PRs.


docs/design/interface/flows.qmd
```diff
@@ -0,0 +1,62 @@
...
+
+```{mermaid}
+%%| label: fig-flow-update-readme
+%%| fig-cap: "Diagram showing the flow of objects and functions to update the README file after changing the package or resource properties."
+%%| fig-alt: "A flow diagram showing the steps to update the README file after changing the package or resource properties."
+flowchart
+    properties[(PackageProperties)]
+    path_properties("path_properties()")
+    read_package_properties("read_package_properties()")
+    as_readme_text("as_readme_text()")
```
lwjohnst86, line 53: I renamed it to `as_` rather than `build_` since it matches other `as_` functions like `asdict()`. And since `build_` to me never really "sang" to me as a good word to describe what was happening. Thoughts?

martonvago: I agree, I like it better this way!

martonvago (dismissed):
Makes sense to me!

docs/design/interface/flows.qmd
signekb: Fine by me :+1: Should `asdict` be `as_dict` then?

signekb, line 53: Maybe we should add some tests on how `as_readme_text()` behaves with custom/edited text?

signekb (dismissed):
✅ ✅ ✅  Just some questions/suggestions not directly related to the approval of this PR.

docs/design/interface/flows.qmd
lwjohnst86: @signekb, `asdict` is a function from another package, so we don't have control on the naming.

lwjohnst86: Hmm, not sure I understand. Since it outputs a string, you can post-process it however you want. The template itself can't be edited/isn't customisable (yet, maybe we can change that later).

PR #1160, opened 2025-03-19
docs: :building_construction: add flow diagram for extracting resource properties

lwjohnst86:
## Description

Another flow diagram for extracting resource properties. Ignore intro section.


docs/design/interface/flows.qmd
```diff
@@ -0,0 +1,73 @@
...
+useful when the raw data is in a format that contains metadata about the
+data, such as a CSV file with a header row that contains the column
+names. The `update_resource_properties()` function will internally call
+`check_resource_properties()`. The `extract_resource_properties()`
+function can not extract all required properties from the raw data, so
+it must be updated by the user. While `TableSchemaProperties` or
+`ResourceProperties` can be updated, the properties in
+`ResourceProperties` must all be filled in and the
+`extract_resource_properties()` function is not able to fill in most of
+the important properties.
```
martonvago, line 47: This bit is a bit confusing to me and seems to repeat the previous sentence (ll. 41-43).

docs/design/interface/flows.qmd
```diff
@@ -0,0 +1,73 @@
...
+    properties_updated[("ResourceProperties<br>(updated)")]
+    update_resource_properties("update_resource_properties()")
+    path_properties("path_properties()")
+    extract_resource_properties("extract_resource_properties()")
+    write_resource_properties("write_resource_properties()")
+
+    raw_data --> extract_resource_properties
+    extract_resource_properties --> properties_extracted
+    properties_extracted --> update_resource_properties
+    schema_properties_updates -. Or .- properties_updates
```
martonvago, line 67: I understand that this means "you can change attributes on the topmost level of the resource properties object or nested attributes in the schema of the resource properties", but the arrow from `TableSchemaProperties` to `update_resource_properties` implies that you can pass that into the function directly, which you can't.

martonvago (changes requested):

docs/design/interface/flows.qmd
```diff
@@ -32,19 +32,68 @@ like Firefox. To see them, try using a different browser like Chrome or
...
+    path_properties --> write_package_properties
+```
+
+## Extract resource properties from data
+
+The flow for extracting resource properties from data. This is useful
+when the data is in a format that contains metadata about the data, such
+as a CSV file with a header row that contains the column names. The
+`extract_resource_properties()` function cannot extract all required
+properties from the data, so it must be updated by the user. So both
```
martonvago, line 92:
```suggestion
properties from the data, so they must be updated by the user. So both
```

docs/design/interface/flows.qmd
```diff
@@ -32,19 +32,68 @@ like Firefox. To see them, try using a different browser like Chrome or
...
+`TableSchemaProperties` and `ResourceProperties` will need many of their
+fields filled in to after using `extract_resource_properties()`. The
```
martonvago, line 94:
```suggestion
fields filled in after using `extract_resource_properties()`. The
```

martonvago (changes requested):
Just the typos, otherwise I approve!!

martonvago (dismissed):

signekb (changes requested):
Based on our discussions yesterday, shouldn’t `extract_resource_properties` take a `pl.DataFrame` as input instead of a path to a file?

PR #1161, opened 2025-03-19
docs: :building_construction: flow diagram to manually create resource properties

lwjohnst86:
## Description

Another flow diagram for creating the resource properties without using extraction.


martonvago (dismissed):
Nice! And if we use `name` as the ID of the resource this becomes so much simpler, no need to worry about where the ID will come from.

signekb (dismissed):

PR #1163, opened 2025-03-19
docs: :building_construction: flow diagram to add data to batch

### LINKED ISSUE #1150: [discussion]: Should `create_resource_structure` create the `resources` folder?
Currently, `create_resource_structure` takes the `resources` folder as input.
All example code that we have has to first create this folder before calling `create_resource_structure`.

A small change would be to have `create_resource_structure` create this folder if it doesn't exist.

This would cut some repetitive code and I think it makes sense logically as well to create the `resources` folder automatically when adding the first resource to a package.

lwjohnst86:
## Description

Flow to add new, added, or modified data to batch folder.

Closes #1150


docs/design/interface/flows.qmd
```diff
@@ -0,0 +1,61 @@
...
+For some reason, the diagrams below don't display well on some browsers
+like Firefox. To see them, try using a different browser like Chrome or
+Edge.
+:::
+
+## Save new, added, or modified raw data to batch
+
+A data resource needs data, not just properties. As long as there is the
+resource properties, you can add data to the resource. Whenever data is
+added in anyway to a data resource, it gets first saved in the `batch/`
```
signekb, line 39:
```suggestion
added to a data resource, it gets first saved in the `batch/`
```
or “any way"

docs/design/interface/flows.qmd
```diff
@@ -0,0 +1,61 @@
...
+folder to keep track of additions or changes. You add the data when:
+
+-   A new data resource is created and raw data is added to it.
+-   Additional raw data is added to an existing data resource.
+-   You need to fix or update or modify existing data in the resource by
+    correcting the data (e.g. fixing a data entry issue).
```
signekb, line 45: With the current design raw data isn’t really being added to the resource, right? It’s always converted to parquet. So, maybe rephrase this? 

```suggestion
-   A new data resource is created and data is added to it.
-   Additional data is added to an existing data resource.
-   You need to fix, update, or modify existing data in the resource by
    correcting the data (e.g. fixing a data entry issue).
```

signekb (changes requested):

docs/design/interface/flows.qmd
lwjohnst86: Very good point!

martonvago (changes requested):

docs/design/interface/flows.qmd
```diff
@@ -0,0 +1,64 @@
...
+added to a data resource, it gets first saved in the `batch/`
+folder to keep track of additions or changes. You add the data when:
+
+-   A new data resource is created and data is added to it.
+-   Additional data is added to an existing data resource.
+-   You need to fix, update, or modify existing data in the resource by
+    correcting the data (e.g. fixing a data entry issue).
+
+The data must be in a tidy format and must have already been
+loaded in as a Polars DataFrame.
```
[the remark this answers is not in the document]
lwjohnst86: Yea, I'm considering having the preparation steps and things be another package called `seedcase-shear` based on what @K-Beicher does in the example repos. Then we can simplify this package to only take tidied DataFrames and putting them into a structured format via the data package. In the design repo, I've been slowly adding some thoughts on how the data is/should be structured/what it means to be tidy.

signekb (changes requested):

signekb (approved):

PR #1164, opened 2025-03-19
docs: :building_construction: flow to create the resource data

lwjohnst86:
## Description

Flow diagram to make or re-make the Parquet resource data file.


docs/design/interface/flows.qmd
```diff
@@ -0,0 +1,69 @@
...
+    resource_properties[(ResourceProperties)]
+    check_data("check_data()")
+
+    path_resource_batch_files --> read_resource_batches
+    read_resource_batches --> list_data
+    list_data --> join_resource_batches
+    join_resource_batches --> joined_data
+    joined_data --> write_resource_data
+    resource_properties --> write_resource_data
+    joined_data -.-> check_data
```
martonvago, line 67: This `check_data` function probably cannot be exactly the same as the existing `check_data`. 

This is because the purpose of the existing `check_data` is to check that user-submitted data matches the Frictionless standard. Looking at CSVs, these are read in from text-based files and all values are treated as string until type-checked and converted to an appropriate Polars/Parquet data type. Null values have to be filled in correctly as well.

Here, data comes in already converted to an appropriate Polars/Parquet data type and with null values filled in correctly.
So it probably needs fewer processing steps:
- check resource properties
- check headers
- check constraints?
- I don't know if it's even possible for data types to be wrong when the Parquet schema is correct

An alternative approach would be to check if the Parquet files are the same ones we saved earlier. If they are, we know they are correct. We could, e.g., compute the hash of each batch Parquet file and save the hash (in a file, e.g. in `user_data_path`, or even directly in the metadata of the Parquet file). When loading each batch Parquet file, we could then check if it hashes to its saved hash.

martonvago (changes requested):

docs/design/interface/flows.qmd
lwjohnst86: This flow to optionally check the data is there just in case there are other additional post-processing steps. Like I had mentioned previously, if we do more checks and it takes some computing time, in the big picture, it doesn't matter since this will be done not very regularly. I'd rather have a simple function that does many things by default, rather than multiple slightly related functions used for specific conditions.

martonvago: Probably! But maybe this context helps a bit:

There are different kinds of `DataFrame` going around:
1. the output of `read_csv`: a data frame with all string columns. This corresponds 1-to-1 to the CSV data source. This data frame can either match the resource properties or not, we don't know until we feed it to `check_data`.
2. the input to Polars' `write_parquet` = the output of Polars' `read_parquet`: a data frame with the correct Polars/Parquet data types. This data frame is very likely to match the resource properties. (The user can still change Parquet files outside of Sprout, but that's more of an edge case.)

How do we go from (1) to (2)? We have to convert all values in (1) to a form that is compatible with our target data types in (2). For many types of values, this can be done automatically (e.g. parsing a string of digits into an integer). These values will look the same in (1) and (2) if we inspect the data frames. But some other types of values will look different: e.g., to save a `yearmonth` as a date we add `-01`. So if we inspect (1) and (2) they will be different: (1) `2012-02` vs (2) `2012-02-01`. (1) matches the Frictionless standard, (2) does not, but it is derived from a value that does.

All this is to say that checking (1) against the resource properties and checking (2) against the resource properties is not exactly the same task. Some values and data types in (2) are different.

Why not have `check_data` take (2) as input then? Because in order to construct (2) we have to check that (1) matches the resource properties (e.g., we cannot create a date column in (2) from strings in (1) that are not correct dates). 

So what should we do?
- I can have a look if it's possible to tweak the column check functions to pass for correct values in both (1) and (2)
- We could have a separate check function for (2)
- We could have a look at the hash approach. The advantage of that one is that we can include anything we want in the hash, even the filename that has the timestamp! Obviously, it would only make sense in a Sprout context. It would (intentionally) prevent you from dropping any old Parquet file into the batch folder. (Of course, it's not at all attack-proof, a user can always create their own hash, but I don't think we're trying to defend against malicious use very much.)

signekb (dismissed):

docs/design/interface/flows.qmd
lwjohnst86: I think that our discussions this week, especially about making a hard requirement for Sprout be a Polars DataFrame and it being Tidy, resolves most of these discussions/comments. So I will resolve them here now :relaxed:

PR #1165, opened 2025-03-19
docs: :building_construction: flow diagram after modifying table schema

lwjohnst86:
## Description

Last of the flow diagrams for now, though others are coming later this week and early next week :P

This one I am less certain about how it will work exactly... I've made a TODO item to highlight it.


docs/design/interface/flows.qmd
```diff
@@ -0,0 +1,66 @@
...
+## Modifying data types or table schema
+
+Sometimes you may need to modify the data types or table schema of the
+data in a resource. So you'd first modify the resource properties before
+using this flow. The `update_resource_batches()` function will
+update each batch DataFrame with the new data types or table schema, then
+you can use `write_resource_batches()` to save the updated data back to the
+batch files.
+
+<!-- TODO: I'm not completely clear yet how this flow might work. Does it overwrite existing ones? Or delete? -->
```
martonvago, line 44: I think it will create new files, so the old ones will be overwritten. And I don't think there's any point keeping the old ones around? Only the metadata will change, the data (even if converted in a different way because a data type changed) will remain the same. And if we want to track metadata changes, it's enough to track just that, without the data?

docs/design/interface/flows.qmd
```diff
@@ -0,0 +1,66 @@
...
+%%| label: fig-flow-modify-data-types
+%%| fig-cap: "Diagram showing the flow of objects and functions to modify data types or table schema."
+%%| fig-alt: "A flow diagram showing the steps to modify data types or table schema."
+flowchart
+    path_resource_batch_files("path_resource_batch_files()")
+    read_resource_batches("read_resource_batches()")
+    properties_updated[("ResourceProperties<br>(updated)")]
+    list_data[("List[DataFrame]")]
+    list_data_updated[("List[DataFrame]<br>(updated)")]
+    update_resource_batches("update_resource_batches()")
```
martonvago, line 56: Do we want to talk about (at this level of detail) how this update will happen exactly?
I just want to flag that it's not enough to update the Polars/Parquet schema. If, for example, a data type changed, we have to make sure all values are converted to the Polars/Parquet data type corresponding to the new Frictionless data type.

One (not particularly inspired) approach would be to undo the Frictionless-style data frame > Parquet-style dataframe conversion and then feed the Frictionless-style data frame to `write_resource_batch`.

docs/design/interface/flows.qmd
lwjohnst86: Yea, that sounds good. So they, write back with the same timestamps? To preserve the sequence of correcting a mistake in the data?

docs/design/interface/flows.qmd
lwjohnst86: Hmmm, I'm not sure I understand this comment. The way I see how it works would be:

1. Parquet (as DataFrame) with old metadata and data types gets compared to the updated resource properties.
2. Any columns in the DataFrame that don't match the updated properties gets converted into type given in the updated properties.
3. After converting to the updated data types, they get written back to the Parquet file.

Is this what you were meaning or?

martonvago: Yes, that's how I see it too!
I was just asking about the details of step 2. There are 10 data types we property distinguish between (not counting the ones we save as string), so that creates quite a few update combinations (string > bool, int > bool, bool > string, bool > int, etc.). Some of these will be trivial (e.g. anything > string), some possible but not trivial (e.g. string > bool, if the column contains valid bool values), and some impossible (e.g. time > date). 
Possible but not trivial updates may include changing the values of the column as well (e.g. map any value in `"false", "False", "FALSE", "0"` to False), not just the data type.
I was just wondering how we wanted to go about doing this.

docs/design/interface/flows.qmd
martonvago: Good point about the timestamp! I guess they must, otherwise we would lose the sequence.

docs/design/interface/flows.qmd
lwjohnst86: I think based on all our discussions elsewhere that these comments have been addressed.

PR #1169, opened 2025-03-19
docs: :building_construction: simplify the deletion functions

lwjohnst86:
## Description

Many of these functions aren't needed, they can be simplified into two instead of so many.

This PR needs a quick review.


signekb (changes requested):
Ah, nice simplification 👌  Just a minor, minor suggestion to remove an extra blank line. Except that, I approve.

signekb (approved):

PR #1170, opened 2025-03-19
docs: :building_construction: add design for deleting obs units

### LINKED ISSUE #1093: Expand on and refine how observational units are managed

lwjohnst86:
## Description

This is a first draft of how we might delete observational units. There's still some things we could improve on, but it's a starting point :D

Closes #1093


docs/design/interface/functions.qmd
```diff
@@ -341,14 +341,61 @@ See the help documentation with `help(write_file)` for more details.
...
+An observational unit is the level of detail on the entity (e.g. human, animal,
+event) that the data was collected on at a given point in time.  An example
+would be: A person in a research study who came to the clinic in May 2024 to
+have their blood collected and to fill out a survey.
+
+### {{< var wip >}} `delete_observational_unit(data_list, observational_unit, confirm)`
+
+As per legal and privacy regulations, a person can request that any personally
+identifiable and sensitive data of theirs be deleted in the "right to be
+forgotten" regulation. This function makes this process easier by searching for
```
signekb, line 353: Add link to this regulation?

docs/design/interface/functions.qmd
```diff
@@ -341,14 +341,61 @@ See the help documentation with `help(write_file)` for more details.
...
+
+# For multiple people with the IDs "1234" and "5678"
+observational_unit = {"person_id": ["1234", "5678"]}
+
+# For one person with the IDs "1234" and date of collections "2024-05-01"
+# and "2024-05-02"
+observational_unit = {
+    "person_id": ["1234"],
+    "date_of_collection": ["2024-05-01", "2024-05-02"]
+}
```
signekb, line 381: If a person requests that they’ll be forgotten, why would we need the dates? Shouldn’t this action delete *all* of their data?
(Keep in mind that I don’t know much about this regulation)

docs/design/interface/functions.qmd
```diff
@@ -341,14 +341,61 @@ See the help documentation with `help(write_file)` for more details.
...
+The data given must be a list of DataFrames, even if it is only a list of one DataFrame.
+The `observational_unit` is a dictionary that contains the information about the
+observational unit that needs to be deleted. The dictionary must contain one or more
+primary keys that represent the observational unit in the data as well as the value
+to delete. For example:
+
+```python
+# For one person with the ID "1234", even if there are multiple time-points
+# of data for that person.
+observational_unit = {"person_id": ["1234"]}
```
signekb, line 371: Would it be nice if users could add the observational unit to the properties? Or does the obs unit depend on the situation and is therefore not *set* for a given resource? 
I might conflate the notion of observational units with [primary keys](https://datapackage.org/standard/table-schema/#primaryKey) a bit

docs/design/interface/functions.qmd
```diff
@@ -341,14 +341,61 @@ See the help documentation with `help(write_file)` for more details.
...
+```
+
+The function will search for all instances of the keys and values in the data and delete them.
 
-### {{< var wip >}} `delete_participant_data()`
+```{mermaid}
+flowchart
+    in_data[/data_list/]
+    in_observational_unit[/observational_unit/]
+    in_confirm[/confirm/]
```
signekb, line 390: What is the `confirm` for? 
It seems that this function isn’t actually overwriting the stored `batch/` `parquet` files, but rather outputting a DataFrame? 
I would expect that the confirm would be for the actual (over)writing of the files themselves, which is the action that effectively deletes the obs unit, right?

signekb (changes requested):
Nice 🪻 I have some questions and suggestions.

docs/design/interface/functions.qmd
lwjohnst86: It's possible that some data doesn't need to be deleted or if they ask to delete after a certain time. I can't actually know what the needs might be, but at least an observational unit is the ID and date. And that will also be the primary key, so might be good to account for situations that need the full primary keys/obs unit, not just part of it.

docs/design/interface/functions.qmd
lwjohnst86: Ah you are right! :+1:

signekb (dismissed):

PR #1173, opened 2025-03-19
refactor: :recycle: convert `path_` functions into class without checks

### LINKED ISSUE #1080: Revise `path_` functions to match changes in design

lwjohnst86:
## Description

I am not done with this PR, but I wanted to first get some thoughts and opinions on this approach. It simplifies the functionality a lot, plus we can use these paths to help write/create non-existent files (which we couldn't do before). I used lowercase for the class, even though we have a style guide of using `PascalCase`. But, it results in conflict with the pathlib `Path`! Thoughts??

Right now, this PR fails, since I didn't update the tests yet.

Closes #1080

## Checklist

- [x] Added or updated tests
- [x] Updated documentation
- [x] Ran `just run-all`


martonvago:
Cool idea!! I think I like it! As you say, it seems like this would simplify things a lot, it works well with how we use path functions -- so why not :grin: 

In terms of what this class is an abstraction for, I guess it is like a navigator / map / layout / signpost / traffic control thing that gives you all the possible places you can go within your Sprout package. So, crucially, one instance of this class is tied to a particular package. So it's a package map.

As for the name, I would prefer not to call it `path` because (as you say) it really should be capitalised, it conflicts with `pathlib.Path`, and it is too broad for what this class actually does. Some options maybe: `PackagePath`, `SproutPath`, `PackageMap`, `PackageStructure`? For me, it would be nice if the fact that it is tied to one package was reflected in the name.

lwjohnst86:
Yea, agreed! :grin: As for the name, I really want something short and easy, preferably one word. What about like `Nav` (as you suggested "navigator"), or maybe `Route` (like for a road)?

martonvago:
Ooh, both of those are great! Maybe I prefer `Route`? `sp.Route().resources()` reads very nicely. `sp.Nav().resources()` is not bad either, but maybe it has more of a "change the cwd to the resources folder" energy. But only maybe :P .

src/seedcase_sprout/core/path_local.py
```diff
@@ -8,286 +8,95 @@
...
-
-            Path(temp_path / "resources").mkdir()
-            # Get the path to the resource folders
-            sp.path_resources(path=temp_path)
-        ```
-    """
-    path = path / "resources"
-    return check_is_dir(path)
+    def __init__(self, path: Path = Path.cwd()) -> Path:
+        """Set the base path."""
```
signekb, line 50:
```suggestion
        """Set the base path of the package."""
```

src/seedcase_sprout/core/path_local.py
```diff
@@ -8,286 +8,95 @@
...
+
+    def resources(self) -> Path:
+        """Path to the `resources` folder."""
+        return self.path / "resources"
+
+    def resource(self, resource_id: str) -> Path:
+        """Gets the absolute path to the specified resource folder.
+
+        Args:
+            resource_id: The ID of the resource. Use `ResourceProperties.name` to get
```
signekb, line 69: Align docstrings here (and below) so they are all either "Path to the […]” or "Gets the absolute path to […]"

signekb (changes requested):
In terms of naming, I like `PackagePath` the best. I don’t think `Route` and `Nav` are informative/descriptive/intuitive enough :)

lwjohnst86:
`PackagePath` is too long though. :thinking: hmmmm. Plus `PackagePath` also confuses things with the already existing `path_package()` in the multi-user setting.

src/seedcase_sprout/core/path_local.py
lwjohnst86: Nice catch, I had copied those over from the previous functions :+1:

signekb:
But maybe we remove the paths related to multiusers, as we talked about after the update meeting Tuesday?

src/seedcase_sprout/core/path_local.py
```diff
@@ -8,286 +8,95 @@
...
-            )
-
-            Path(temp_path / "resources").mkdir()
-            # Get the path to the resource folders
-            sp.path_resources(path=temp_path)
-        ```
-    """
-    path = path / "resources"
-    return check_is_dir(path)
+    def __init__(self, path: Path = Path.cwd()) -> Path:
```
signekb, line 49: Didn’t you mention something about having the root path direct to where the `datapackage.json` file is at some point — not the cwd?

lwjohnst86:
Updated based on everyone's comments and merged in :grin:

PR #1179, opened 2025-03-21
fix: :bug: check that package and resource properties are `dict`s

martonvago:
## Description

This PR adds two missing instance checks in the Sprout-specific `check_properties`. Without these the test cases I've added fail.

This PR needs a quick review.

## Checklist

- [x] Added or updated tests
- [x] Updated documentation
- [x] Ran `just run-all`

### DIFF
--- a/src/seedcase_sprout/core/sprout_checks/check_properties.py
+++ b/src/seedcase_sprout/core/sprout_checks/check_properties.py
@@ -44,11 +44,13 @@ def check_properties(
     )
 
     errors = checks.check_properties(properties_dict)
-    errors += get_sprout_package_errors(properties_dict)
 
-    if isinstance(properties_dict.get("resources"), list):
-        for index, resource in enumerate(properties_dict.get("resources")):
-            errors += get_sprout_resource_errors(resource, index)
+    if isinstance(properties_dict, dict):
+        errors += get_sprout_package_errors(properties_dict)
+        if isinstance(properties_dict.get("resources"), list):
+            for index, resource in enumerate(properties_dict.get("resources")):
+                if isinstance(resource, dict):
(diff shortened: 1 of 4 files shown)


src/seedcase_sprout/core/sprout_checks/check_properties.py
```diff
@@ -44,11 +44,13 @@ def check_properties(
...
 
-    if isinstance(properties_dict.get("resources"), list):
-        for index, resource in enumerate(properties_dict.get("resources")):
-            errors += get_sprout_resource_errors(resource, index)
+    if isinstance(properties_dict, dict):
+        errors += get_sprout_package_errors(properties_dict)
+        if isinstance(properties_dict.get("resources"), list):
+            for index, resource in enumerate(properties_dict.get("resources")):
+                if isinstance(resource, dict):
+                    errors += get_sprout_resource_errors(resource, index)
```
lwjohnst86, line 53: This tells me we really need to take a step back and think of the design if we resort to a fourth-level nested code.

martonvago: Some other options:
- move the `isinstance` checks inside `get_sprout_package_errors` and `get_sprout_resource_errors`. Okay, although I think these checks belong on this outer level logically.
- unnest the structure buy repeating and ANDing `isinstance` checks (e.g. `if isinstance(properties_dict, dict) and isinstance(properties_dict.get("resources"), list):`)
- end early after each failed `isinstance` check. Seems nice, but then you have to extract the tail end of the function into a helper function like `process_errors`, which makes it a bit inelegant again.

lwjohnst86 (approved):
I'll approve for now, since we'll do a refactor of the checks later anyway.

Issue #1181, opened 2025-03-23
Do a brainstorming session on design of data check functionality


lwjohnst86:
I am quite at fault for not prioritizing/not recognizing that we should have spent a good chunk of time on the design of the (mainly Sprout-specific) check data functionality. This is part of the reason why it is taking me so long to review #1125. After spending a bit more time looking into it, I think I see a problem of not having a wider design in place for it. So we should spend some effort to brainstorm the design.

Some things to discuss/brainstorm, starting from the end and working backwards:

- What are some explicit/detailed flows of objects and actions on those objects when it comes to checking data
- Make clear how functional programming is involved/looks like in these tasks
- Identify steps that need to happen along the way (backwards)

This is a good starting point, though there are probably many things I'm missing here. Ideally we have a flow diagram, the inputs and outputs of the exposed function (I think these data checks should be public), and then some pseudocode of what the implementation looks like, so that we have something to discuss and agree on before spending time creating more functions or refactoring existing ones.

lwjohnst86:
We agree on a hard requirement of Tidy Data, as a Polars DataFrame.

PR #874, opened 2024-11-13
docs: :building_construction: design of the check functions

lwjohnst86:
## Description

This PR starts adding the design/potential functions of the various check functionality.

Closes #834 and closes #849

This PR needs an in-depth review.

PR #1191, opened 2025-03-24
docs: :building_construction: flow for checking data against the properties

lwjohnst86:
## Description

This creates a flow diagram of how the `check_data()` may work. Connected to #874 and #1181.


docs/design/interface/flows.qmd
```diff
@@ -113,3 +113,40 @@ flowchart
...
+if the check fails, otherwise it outputs a data frame with the same data
+as the input. The `check_data()` then gathers together all the messages
+and gives an error using those messages if there are any.
+
+```{mermaid}
+%%| label: fig-flow-check-data-against-properties
+%%| fig-cap: "Diagram showing the flow of objects and functions to check data against the properties."
+%%| fig-alt: "A flow diagram showing the steps to check data against the properties."
+flowchart
+    data[(DataFrame)]
```
martonvago, line 134: Just putting an earlier question here for reference: which data frame is this input? A data frame with values converted to Polars/Parquet data types or a data frame with unconverted (string type) values?

If it is a data frame with values converted to Polars/Parquet data types:
- When did we do that conversion?
- Before doing the conversion, we have to check that the conversion is possible (e.g., that in a date column, all strings can be parsed as dates). When did we do that check?
- If the values are already converted to the target data type, what do we mean by checking their types against the properties? Why isn't checking the column's data type enough? 

(Related [comment](https://github.com/seedcase-project/seedcase-sprout/pull/1164#discussion_r2007666867))

martonvago (commented):
I haven't yet read #1192, sorry if that makes some of these points moot

docs/design/interface/flows.qmd
```diff
@@ -113,3 +113,40 @@ flowchart
...
+    resource_properties[(ResourceProperties)]
+    check_data("check_data()")
+    subgraph Internal
+        check_column_names("_check_column_names()")
+        check_column_types("_check_column_types()")
+        check_column_values_types("_check_column_values_types()")
```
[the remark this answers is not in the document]
lwjohnst86: E.g. `check_column_types()` is checking that the column `age` has the same data type of `int` in properties as in the column. `_values_types()` is what you've been doing in the `check_column_data_types()`: It checks the individual values in the column if they match the data type or not. Though, thinking about this more and in response to your questions/comments above, this check on the data made sense when we were importing from CSV and preparing things for storing into Parquet. But this functionality sounds like it makes more sense in the pre-processing package rather than here. :thinking:

docs/design/interface/flows.qmd
lwjohnst86: Hmm, I think I can see where some of the confusion is coming from now. I was really trying to understand your previous comments but now I get it.

I think, we've slowly been moving over to taking *already pre-processed and tidy* data frames and storing them into Sprout. We've made some comments about this here and there over the last few months. I've just made an issue about it here #1193 

So now, checking the data values themselves and checking that the conversion work doesn't make sense. As what @K-Beicher has been doing, much of the work is tidying up the data outside of Sprout in Polars (we shouldn't have that be the focus of Sprout anyway, there's already lots of packages to help with cleaning, plus it is *very* domain specific). So maybe, we could just drop the whole checking the data values themselves. Which would simplify things a lot!

So, to answer almost all the questions/comments, the data is already tidy when Sprout takes over. This includes dealing with missing values, correct conversion, etc.

(small comment about Frictionless and user-defined missing values... that is a very strong opposition from me :laughing: missing values should not be decided by the user, it should be an agreed on standard, which is `NA`/`Null`/`None` (whatever language implements "not available").

martonvago: Ah, alright, that is a different kettle of fish! 
Then, I guess, the exercise is more about defining and documenting what Sprout means by tidy data (you're probably already doing that somewhere :sweat_smile: ).

I think much of this misunderstanding is originating from the fact that the Frictionless standard is written for text-based data and is concerned very much with how string data items in that text are formatted. If input data in Sprout is not text-based, then a large chunk of the Frictionless standard is irrelevant (e.g., I cannot check the format of a date object).

So... What we want is a mapping between Frictionless types and Polars/Parquet types. Then, when we check the data type of a column, we just* check that the corresponding Frictionless field's type maps to the column's type? That sure is simpler!

\* I'm sure there will be some delicate edge cases, e.g., Polars allowing negative dates but this creating dates in Parquet which are not legible by other Parquet tools...

docs/design/interface/flows.qmd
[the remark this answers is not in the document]
martonvago: Yeah, exactly! Generally, I don't think it makes sense to check that each value is an `int` in an `int` column, because it is definitionally that.

The only case I can think of when an extra check is necessary is when we have additional restrictions on a data type. E.g., as in the other comment, the case of negative dates. If we even care about preventing errors caused by that. :shrug:

PR #1192, opened 2025-03-24
docs: :building_construction: pseudocode for `check_data()`

lwjohnst86:
## Description

To help in the brainstorming of the check data functionality, I've added pseudocode for how the `check_data()` function might look like in the end.

### DIFF
--- a/docs/design/interface/pseudocode/check_data.py
+++ b/docs/design/interface/pseudocode/check_data.py
@@ -1,7 +1,5 @@
 # ruff: noqa
-def check_data(
-    data: DataFrame, resource_properties: ResourceProperties
-) -> DataFrame:
+def check_data(data: DataFrame, resource_properties: ResourceProperties) -> DataFrame:
     """Checks that the DataFrame matches the requirements in the resource properties.
 
     Runs a few checks to compare between the data and the properties on the items:
@@ -46,19 +44,8 @@ def check_data(
     check_resource_properties(resource_properties)
 
     # TODO: These individual checks have their own pseudocode files.
-    error_messages = _check_column_names(data, resource_properties)
-
-    error_messages += _check_column_types(data, resource_properties)
-
-    # TODO: If there are errors in the types above, this next line won't matter. Maybe have an error output here first?
-    if error_messages:
-        raise ExceptionGroup(error_messages, ...)
-
-    error_messages += _check_column_values_types(data, resource_properties)
-
-    error_messages += _check_column_values_constraints(data, resource_properties)
-
-    if error_messages:
-        raise ExceptionGroup(error_messages, ...)
+    _check_column_names(data, resource_properties)
+    _check_column_types(data, resource_properties)
+    _check_column_values_constraints(data, resource_properties)
 
     return data
--- a/docs/design/interface/pseudocode/check_data.py
+++ b/docs/design/interface/pseudocode/check_data.py
@@ -1,14 +1,14 @@
 # ruff: noqa
 def check_data(
     data: DataFrame, resource_properties: ResourceProperties
-) -> DataFrame | Error:
+) -> DataFrame:
     """Checks that the DataFrame matches the requirements in the resource properties.
 
     Runs a few checks to compare between the data and the properties on the items:
 
     | Data | Properties |
     |:------|:------------|
-    | Column names | `field.names` |
+    | Column names | `field.name` |
     | Column types | `field.types` |
     | Column values' types | `field.types` |
     | Column values' constraints | `field.constraints` |
@@ -27,7 +27,7 @@ def check_data(
         resource_properties: The specific `ResourceProperties` for the `data`.
 
     Returns:
-        If there are any issues, it gives an Error. Otherwise, will output the `data`.
+        Output the `data` if checks all pass.
 
     Raises:
         ExceptionGroup: A list of messages that highlight where there are differences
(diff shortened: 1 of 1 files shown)


docs/design/interface/pseudocode/check_data.py
```diff
@@ -0,0 +1,64 @@
+# ruff: noqa
+def check_data(
+    data: DataFrame, resource_properties: ResourceProperties
+) -> DataFrame | Error:
```
martonvago, line 4:
```suggestion
) -> DataFrame:
```
We want to throw an error, not return it, right?

docs/design/interface/pseudocode/check_data.py
```diff
@@ -0,0 +1,64 @@
...
+    """Checks that the DataFrame matches the requirements in the resource properties.
+
+    Runs a few checks to compare between the data and the properties on the items:
+
+    | Data | Properties |
+    |:------|:------------|
+    | Column names | `field.names` |
```
martonvago, line 11:
```suggestion
    | Column names | `field.name` |
```
I understand this means the names of all fields, but there is no `names` attribute. Same for type.

docs/design/interface/pseudocode/check_data.py
```diff
@@ -0,0 +1,64 @@
...
+    >
+    > - In the properties: {mismatch}
+    > - In the data: {mismatch}
+
+    Args:
+        data: A Polars DataFrame.
+        resource_properties: The specific `ResourceProperties` for the `data`.
+
+    Returns:
+        If there are any issues, it gives an Error. Otherwise, will output the `data`.
```
martonvago, line 30: As above, it raises the error, not returns it, right?

docs/design/interface/pseudocode/check_data.py
```diff
@@ -0,0 +1,64 @@
...
+        ```
+    """
+    check_resource_properties(resource_properties)
+
+    # TODO: These individual checks have their own pseudocode files.
+    error_messages = _check_column_names(data, resource_properties)
+
+    error_messages += _check_column_types(data, resource_properties)
+
+    # TODO: If there are errors in the types above, this next line won't matter. Maybe have an error output here first?
```
martonvago, line 53: Good point! And I would actually let it terminate early already after mismatched column names, because if we cannot pair up columns in the data frame and fields in the properties, then we cannot check anything about the columns, not even their types.
(In that case, it's enough to let that function throw, it doesn't have to return its errors. When it throws, that error will travel up the callstack and terminate the operation.)

The same is true for checking data types and constraints: we cannot check the constraints on data that has the wrong types. So we can exit early after a failed type check as well.

docs/design/interface/pseudocode/check_data.py
```diff
@@ -0,0 +1,64 @@
...
+    if error_messages:
+        raise ExceptionGroup(error_messages, ...)
+
+    error_messages += _check_column_values_types(data, resource_properties)
+
+    error_messages += _check_column_values_constraints(data, resource_properties)
+
+    if error_messages:
+        raise ExceptionGroup(error_messages, ...)
```
martonvago, line 62: Minor thing: this will take a list of errors and not error messages

docs/design/interface/pseudocode/check_data.py
```diff
@@ -0,0 +1,64 @@
...
+    | Column values' constraints | `field.constraints` |
+
+    Error messages output generally in the format of:
+
+    > # {data item}:
+    >
+    > There is a mismatch found:
+    >
+    > - In the properties: {mismatch}
+    > - In the data: {mismatch}
```
martonvago, line 23: Would we not want to group the errors somehow so that we don't print this for each failed data item? E.g., by type of error (was it the wrong type? was it a violated constraint?) and/or by column?

martonvago (changes requested):
Thanks, I think this will be really good as a basis for the brainstorming/discussion!
All my comments here are pretty minor, but I put 2 meatier ones on #1191, which are applicable here as well.

docs/design/interface/pseudocode/check_data.py
lwjohnst86: Ha, that's right hahahaha

docs/design/interface/pseudocode/check_data.py
lwjohnst86: Thanks for the catch! :grin:

docs/design/interface/pseudocode/check_data.py
lwjohnst86: Yea, the specifics of how the error messages are output can be brainstormed a lot I think (we do have a whole issue for it). And it also shows my lack of understanding/knowledge on there being different types of errors in Python (I still don't understand why that's needed, but that doesn't matter :stuck_out_tongue:).

And yea, I guess it isn't super clear, but the "format" I describe here would be modified for each data item in the table above. So it would give errors that are particular to the check.

docs/design/interface/pseudocode/check_data.py
lwjohnst86, line 30:
```suggestion
        Output the `data` if checks all pass.
```

docs/design/interface/pseudocode/check_data.py
lwjohnst86: This is pseudocode :wink: doesn't have to work yet :stuck_out_tongue:

docs/design/interface/pseudocode/check_data.py
lwjohnst86: Yea, that's a good point... Ok, so then each `_check_*()` function should contain the exception rather than output a string? Yea, that makes sense in this context :+1:

martonvago: Yeah, I think in general: raise an exception if the error should terminate the operation, return an exception(s) if we want to collect all exceptions before finishing.

PR #1198, opened 2025-03-25
docs: :pencil2: add resource guide to guide sidebar

signekb:
## Description

Seems like we forgot to add the resource guide to the sidebar, when we added the guide. 

Alternatively, we could do something like: 

```
contents: 
          - auto: "docs/guide/*.qmd” 
```

So we don’t have to update the sidebar with new guides. But then we won’t get the order that we want.

This PR needs a quick review.

## Checklist

- [X] Updated documentation
- [X] Ran `just run-all`

### DIFF
--- a/_quarto.yml
+++ b/_quarto.yml
@@ -66,6 +66,7 @@ website:
           contents:
             - docs/guide/installation.qmd
             - docs/guide/packages.qmd
+            - docs/guide/resources.qmd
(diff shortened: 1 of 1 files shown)


K-Beicher (approved):
Good catch

lwjohnst86 (approved):
Yea, that whole ordering thing is why I hadn't done `*.qmd`.

PR #1199, opened 2025-03-26
chore: 👩‍💻 use asterisk to include all guides in sidebar

signekb:
## Description

The reason we thought the ordering didn't work with `*` was that we actually had two documents with `order: 2`.
Now, we only have to remember to add the order—we don’t have to remember to add it to the sidebar 🙌 

This PR needs a quick review.

## Checklist

- [X] Updated documentation
- [X] Ran `just run-all`

### DIFF
--- a/_quarto.yml
+++ b/_quarto.yml
@@ -64,10 +64,7 @@ website:
         - section: "Guide"
           href: docs/guide/index.qmd
           contents:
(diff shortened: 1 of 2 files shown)


lwjohnst86 (approved):
That works also :relaxed: we'll just have to keep in mind to add the order yaml key when adding a guide :+1:

Issue #806, opened 2024-10-23
Write and test `write_resource_parquet` function


martonvago:
- Signature: `write_resource_parquet(raw_files: list[Path], path: Path) -> Path`
- Arguments:
  - `raw_files`: a list of paths pointing to the raw files of the resource
  - `path`: path to the `data.parquet` file of the resource
- Returns the path to the parquet file.
- If the parquet file doesn't exist yet, it should create it and add the data from the raw data files. If the parquet file exists, data from the raw data files should be merged with the data already in the parquet file.
- We'll have to have a think about the merge strategy to use here. It's also an option to add the data to the parquet file in the simplest way possible for this issue and refine the merge behaviour later.
- We should be able to use `pandas` to work with csv and parquet files.

lwjohnst86:
@martonvago small comment, we'll use Polars instead of Pandas for importing and working with CSV and Parquet data.

PR #1200, opened 2025-03-26
docs: :memo: make resources guide run

martonvago:
## Description

This PR makes the resources guide run.
I also retouched the text a bit.
Code changes are very minor.

Closes #1094

This PR needs an in-depth review.

## Checklist

- [x] Added or updated tests
- [x] Updated documentation
- [x] Ran `just run-all`

Issue #1180, opened 2025-03-23
Collect all variables/constants into a single constants file


lwjohnst86:
It's very difficult to keep track of all the constants/variables being used. So to help keep track of them all, let's put them in a common location, like `constants.py`.

PR #1201, opened 2025-03-26
feat: :sparkles: add `read_resource_batches()` - without `check_data()`

signekb:
## Description

This PR adds the function `read_resource_batches()` without checking if the data aligns with the resource properties (`check_data()` isn’t implemented yet)

Related to #806 since before being able to write batch files to `data.parquet` (and remove duplicate rows), the batch files needs to be read.

EDIT: Sorry, this PR became a bit larger than expected. 

## Checklist

- [X] Added or updated tests
- [X] Ran `just run-all`


src/seedcase_sprout/core/read_resource_batches.py
```diff
@@ -0,0 +1,138 @@
...
+        _check_timestamp_format("2025-03-26T153402Z")
+        ```
+    """
+    try:
+        datetime.strptime(timestamp, format)
+        return timestamp
+    except ValueError as error:
+        raise ValueError(
+            f"Timestamp '{timestamp}' is not in the expected format {format}"
+        ) from error
```
signekb, line 138: Not sure whether this should be under `checks` but decided on having it here bc I don’t think it will be used elsewhere. Thoughts?

tests/core/test_read_resource_batches.py
```diff
@@ -0,0 +1,133 @@
...
+@fixture
+def test_package(tmp_path):
+    create_test_data_package(tmp_path)
+    batch_path = tmp_path / "resources" / "1" / "batch"
+    batch_path.mkdir(parents=True)
+
+    for index, batch_data in enumerate([batch_data_1, batch_data_2], start=1):
+        batch_data.write_parquet(
+            Path(batch_path / f"2025-03-26T100346Z-{index}").with_suffix(".parquet")
+        )
```
signekb, line 56: Ideally, I would add resources to the test package in a less manual way. Could add a function `setup_directory_struture.py`?

Didn’t want to use `create_test_resource()` yet bc I saw it was edited anyway in #1200

tests/core/test_read_resource_batches.py
```diff
@@ -0,0 +1,133 @@
...
+
+    assert True
+
+
+def test_raises_error_when_timestamp_is_incorrect_format():
+    """Raises ValueError when the timestamp is not in the correct format."""
+    incorrect_timestamp = "2025-03-26T10"
+
+    with raises(ValueError):
+        _check_timestamp_format(incorrect_timestamp)
```
signekb, line 133: Not sure if it’s good practice to test a helper function (one of several) here. Thoughts?

src/seedcase_sprout/core/read_resource_batches.py
```diff
@@ -0,0 +1,138 @@
...
+        _add_timestamp_as_column(pl.DataFrame(), "2025-03-26T153402Z")
+        ```
+    """
+    return data.insert_column(0, pl.Series("timestamp", [timestamp] * len(data)))
+
+
+def _check_timestamp_format(timestamp: str, format: str = "%Y-%m-%dT%H%M%SZ") -> str:
+    """Check if the timestamp is in the expected format.
+
+    The expected format is "%Y-%m-%dT%H%M%SZ", which is the format used by
```
martonvago, line 114: Could we then extract this into a constant and use it here and in `get_compact_iso_timestamp` (where it's coming from)

src/seedcase_sprout/core/read_resource_batches.py
```diff
@@ -0,0 +1,138 @@
...
+    data = _add_timestamp_as_column(data, timestamp)
+    return data
+
+
+def _extract_timestamp_from_batch_file_path(path: Path) -> str:
+    """Extracts the timestamp from the file name.
+
+    Since the batch file name has been created by `create_batch_file_name()`, with a
+    specific format, the timestamp can be extracted by taking the first 18 characters of
+    the file name.
```
martonvago, line 88: This is true, but what happens if someone renames the file while preserving the first 18 chars? For example:
original name: `2025-03-27T110222Z-21dbe81c-...`
edited name: `2025-03-27T110222Z-2025-03-27T110222Z-21dbe81c-...`

Maybe one approach would be to (1) check the file name as a whole (does it match the expected `<timestamp>-<uuid>` pattern?) and (2) parse the timestamp (now it's safe).

src/seedcase_sprout/core/read_resource_batches.py
martonvago: I think a check function for checking the whole file name could conceivably go under `sprout_checks`. For just the timestamp, I also don't mind.

tests/core/test_read_resource_batches.py
martonvago: Maybe using `write_resource_batch` once it's done?

If `create_test_resource` would help, I think it's okay to use! The args or the return value didn't change.

tests/core/test_read_resource_batches.py
```diff
@@ -0,0 +1,133 @@
...
+    timestamp_column = [data[0]["timestamp"], data[1]["timestamp"]]
+
+    # Then
+    assert len(data) == 2
+    assert data[0].shape and data[1].shape == (3, 3)
+    assert timestamp_column[0].name and timestamp_column[1].name == "timestamp"
+    assert (
+        timestamp_column[0].unique()[0]
+        and timestamp_column[1].unique()[0] == "2025-03-26T100346Z"
+    )
```
martonvago, line 82: Are these `and`s doing what is expected? Isn't the expression before the `and` just checked for truthy-ness?

tests/core/test_read_resource_batches.py
```diff
@@ -0,0 +1,133 @@
...
+    resource_paths.append(Path("non-existent-file.parquet"))
+
+    # When, Then
+    with raises(FileNotFoundError):
+        read_resource_batches(
+            paths=resource_paths, resource_properties=resource_properties
+        )
+
+
+def test_raises_error_when_properties_do_not_match_data(resource_paths):
```
martonvago, line 97: If you want a test that is disabled, you can call it `def xtest_...`
Then you don't have to fake the pass

martonvago (changes requested):
This is great!! :tulip: 
I put lots of comments, but take only what you want!

src/seedcase_sprout/core/read_resource_batches.py
```diff
@@ -0,0 +1,138 @@
...
+from pathlib import Path
+
+import polars as pl
+
+from seedcase_sprout.core.check_is_file import check_is_file
+
+# from seedcase_sprout.core.checks.check_data import check_data
+from seedcase_sprout.core.checks.check_resource_properties import (
+    check_resource_properties,
+)
```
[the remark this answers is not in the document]
signekb: You’re right! 
```suggestion
from seedcase_sprout.core.sprout_checks.check_resource_properties import (
    check_resource_properties,
)
```

src/seedcase_sprout/core/read_resource_batches.py
```diff
@@ -0,0 +1,138 @@
...
+        import seedcase_sprout.core as sp
+
+        sp.read_resource_batches(
+            paths=sp.path_resources_batch_files(1),
+            resource_properties=sp.example_resource_properties(),
+        )
+        ```
+    """
+    list(map(check_is_file, paths))
+    check_resource_properties(resource_properties.compact_dict)
```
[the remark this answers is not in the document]
signekb:
```suggestion
    check_resource_properties(resource_properties)
```

tests/core/test_read_resource_batches.py
signekb: Thanks for the catch!

src/seedcase_sprout/core/read_resource_batches.py
```diff
@@ -0,0 +1,138 @@
...
+        ``` python
+        import seedcase_sprout.core as sp
+
+        sp.read_resource_batches(
+            paths=sp.path_resources_batch_files(1),
+            resource_properties=sp.example_resource_properties(),
+        )
+        ```
+    """
+    list(map(check_is_file, paths))
```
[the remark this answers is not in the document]
signekb: You’re right! Leftover from some testing I did. Thanks!

src/seedcase_sprout/core/read_resource_batches.py
signekb: That’s true… I’m not sure what we want to do with these constants yet, but we seem to have a few in a lot of different places. Might be nice to collect them somewhere. Do we want them in a `constants.py` file on the module level? I..e, one in `core`, one in `checks`, one in `sprout_checks`? 🤔 
“Constants" is a quite generic name though, but I’m not sure we have enough constants to start splitting them up into separate files within modules. What do you think?

src/seedcase_sprout/core/read_resource_batches.py
```diff
@@ -0,0 +1,138 @@
...
+    Args:
+        path: Path to the Parquet batch file.
+
+    Returns:
+        The Parquet file as a DataFrame with a timestamp column added.
+
+    Examples:
+        ```{python}
+        _read_parquet_batch_file(Path("resources/1/batch/2025-03-26T153402Z-1.parquet"))
+        ```
```
lwjohnst86, line 74:
```suggestion

    Examples:
        ``` python
        _read_parquet_batch_file(Path("resources/1/batch/2025-03-26T153402Z-1.parquet"))
        ```
```

Either make sure this code chunk doesn't execute (if built on the website). Or we decide we don't use example code in internal function docstrings:

```suggestion
```

src/seedcase_sprout/core/read_resource_batches.py
lwjohnst86: Yea, I think that idea is good. Blindly extracting the first 18 characters can be a recipe for issues later if the first 18 aren't the timestamp.

src/seedcase_sprout/core/read_resource_batches.py
```diff
@@ -0,0 +1,138 @@
...
+        data: Data to add timestamp column to.
+        timestamp: Timestamp to add as values in the timestamp column.
+
+    Returns:
+        Data with added timestamp column.
+
+    Examples:
+        ```{python}
+        _add_timestamp_as_column(pl.DataFrame(), "2025-03-26T153402Z")
+        ```
```
lwjohnst86, line 106:
```suggestion
```

I think it's fine to not use examples for internal functions. Unless we want to agree on using examples in internal functions.

We also have a style guide saying to keep docstrings short for internal, small functions: https://design.seedcase-project.org/style/code

src/seedcase_sprout/core/read_resource_batches.py
```diff
@@ -0,0 +1,138 @@
...
+    """
+    return data.with_columns(pl.lit(timestamp).alias("__timestamp__"))
```
lwjohnst86, line 108:
```suggestion
    return data.with_columns(pl.lit(timestamp).alias("_batch_file_timestamp"))
```

Just to be clearer.

src/seedcase_sprout/core/read_resource_batches.py
lwjohnst86: There's already an issue about this #1180 so yes, one constants file for each folder module. Then it can be imported and referred to via:

```
import ...constants as constants

constants.timestamp
```

lwjohnst86 (changes requested):

src/seedcase_sprout/core/read_resource_batches.py
signekb: I’ll remove them 🔥

src/seedcase_sprout/core/read_resource_batches.py
signekb: I’ve removed it 🔥

src/seedcase_sprout/core/read_resource_batches.py
martonvago: And until we have a file for it, it's also okay to just put it right above the function, I think

src/seedcase_sprout/core/sprout_checks/constants.py
```diff
@@ -0,0 +1,11 @@
...
+
+"""regex for UUID4, from https://gist.github.com/johnelliott/cf77003f72f889abbc3f32785fa3df8d?permalink_comment_id=4318295#gistcomment-4318295"""
+UUID4_PATTERN = r"[0-9a-f]{8}-[0-9a-f]{4}-4[0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}"
+
+"""regex for timestamp with format '%Y-%m-%dT%H%M%SZ'. Should match the format used by
+BATCH_TIMESTAMP_FORMAT in sprout.core"""
+TIMESTAMP_PATTERN = r"^\d{4}-\d{2}-\d{2}T\d{6}Z"
+
+"""regex for batch file name"""
+BATCH_FILE_NAME_PATTERN = rf"^{TIMESTAMP_PATTERN}-{UUID4_PATTERN}.parquet$"
```
signekb, line 11: This regex matches the pattern currently created by `create_batch_file_name()`. Do we want it to be this strict? We could also just check that the filename includes a timestamp in the expected format? 

What do you think? :)

tests/core/test_read_resource_batches.py
signekb: Good point! Might do that going forward :) Isn’t it fine to keep these as is for now?

src/seedcase_sprout/core/sprout_checks/constants.py
lwjohnst86: Hmmmmm, great question. I'm leaning towards only requiring that the file as a timestamp in it.

src/seedcase_sprout/core/read_resource_batches.py
```diff
@@ -0,0 +1,119 @@
...
+        ValueError: If a column with the name _batch_file_timestamp_ already exists in
+        the data.
+    """
+    if "_batch_file_timestamp_" in data.columns:
+        raise ValueError(
+            "The provided resource batch files contains a column named "
+            "'_batch_file_timestamp_'. This column is used internally in Sprout to "
+            "remove duplicate rows across batches. Please rename it in the "
+            "batch files and resource properties to read the resource batches."
+        )
```
signekb, line 118: Writing this error message made me think that this is a bit of an awkward time to throw this error? 
Should we check for this column either when creating the resource properties or writing to batch?

Alternatively, we need something more flexible/crazy — like adding a uuid to the timestamp column minimise the risk of this happening tremendously?

This is very much an edge case though.

src/seedcase_sprout/core/sprout_checks/constants.py
signekb: And what, if for some reason, the batch filename includes two timestamps? Or is that something we just don’t expect to happen?

src/seedcase_sprout/core/read_resource_batches.py
```diff
@@ -0,0 +1,119 @@
...
+        ``` python
+        import seedcase_sprout.core as sp
+
+        sp.read_resource_batches(
+            paths=sp.path_resources_batch_files(1),
+            resource_properties=sp.example_resource_properties(),
+        )
+        ```
+    """
+    map(check_is_file, paths)
```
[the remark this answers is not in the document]
lwjohnst86: Ha, that's funny. But it is right that Python does not actually evaluate `map()` right away, unlike other programs like R. So we need to wrap it in `list()` to force execution.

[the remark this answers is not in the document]
lwjohnst86: I updated it to use `list()` instead

tests/core/test_read_resource_batches.py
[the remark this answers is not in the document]
signekb: I’ve added a test for 1 in this file, and a test for when the filename doesn’t fit BATCH_FILE_NAME_PATTERN in the tests for `check_batch_file_name()`

tests/core/test_read_resource_batches.py
```diff
@@ -94,6 +100,31 @@ def test_raises_error_when_file_does_not_exist(resource_paths):
...
 
 
+def test_raises_error_when_timestamp_column_matches_existing_column(resource_paths):
+    """Raises ValueError when the timestamp column name matches an existing column."""
+    # Given
+    batch_path = resource_paths[0].parent
+
+    for file in batch_path.iterdir():
+        if file.is_file():
+            file.unlink()
```
martonvago, line 110: Is this necessary?

tests/core/test_read_resource_batches.py
martonvago: (((For sure! The difference is that if you cared about seeing how many of your tests pass then it would be misleading.)))

martonvago (changes requested):
Minor things again, take them or leave them :grin:

src/seedcase_sprout/core/read_resource_batches.py
martonvago: Very good idea about checking for clashing col names being part of a check! We might have other reserved col names later (e.g. when we get to checking constraints or other row-level checks), so checking for all these could be done in one go. Would make sense to put it in the Sprout-specific properties checks. Along with checking for col name uniqueness perhaps.

martonvago: I'm not sure how I feel about adding a timestamp column in all batch parquet files. 
- The great advantage would be that it's less likely they will be modified by a rename and are guaranteed to be valid datetimes. That's super useful!
- The disadvantage is that it's mixing metadata with data a bit, right?

Parquet files can also contain custom metadata (just plain old key-value pairs), so we could put it there as well. That would be as a string, so it would still need validation.

src/seedcase_sprout/core/read_resource_batches.py
```diff
@@ -100,39 +105,15 @@ def _add_timestamp_as_column(data: pl.DataFrame, timestamp: str) -> pl.DataFrame
...
+    if "_batch_file_timestamp_" in data.columns:
         raise ValueError(
-            f"Timestamp '{timestamp}' is not in the expected format {format}"
-        ) from error
+            "The provided resource batch files contains a column named "
+            "'_batch_file_timestamp_'. This column is used internally in Sprout to "
+            "remove duplicate rows across batches. Please rename it in the "
+            "batch files and resource properties to read the resource batches."
+        )
+    return data.with_columns(pl.lit(timestamp).alias("_batch_file_timestamp_"))
```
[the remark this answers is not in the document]
signekb: Yesssss! Also just thought of that today (working on `join_resource_batches()`)

src/seedcase_sprout/core/read_resource_batches.py
lwjohnst86: Yea, on the one hand, having the timestamp be in the data itself would solve some of the issues. But on the other hand, then we would have to figure out a way to add it to the resource properties, and make sure it is always there when/if a user modifies the properties.

src/seedcase_sprout/core/sprout_checks/check_batch_file_name.py
```diff
@@ -0,0 +1,27 @@
+import re
+from pathlib import Path
+
+from seedcase_sprout.core.sprout_checks.constants import BATCH_FILE_NAME_PATTERN
+
+
+def check_batch_file_name(path: Path) -> Path:
```
lwjohnst86, line 7: This is internal, so let's start following that `_` prefix practice.

lwjohnst86 (changes requested):
Just some initial comments, I'll review again after.

src/seedcase_sprout/core/sprout_checks/constants.py
```diff
@@ -0,0 +1,11 @@
+"""Constants for the sprout_checks module."""
+
+"""regex for UUID4, from https://gist.github.com/johnelliott/cf77003f72f889abbc3f32785fa3df8d?permalink_comment_id=4318295#gistcomment-4318295"""
+UUID4_PATTERN = r"[0-9a-f]{8}-[0-9a-f]{4}-4[0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}"
+
+"""regex for timestamp with format '%Y-%m-%dT%H%M%SZ'. Should match the format used by
+BATCH_TIMESTAMP_FORMAT in sprout.core"""
+TIMESTAMP_PATTERN = r"^\d{4}-\d{2}-\d{2}T\d{6}Z"
```
signekb, line 8:
```suggestion
TIMESTAMP_PATTERN = r"^\d{4}-(0[1-9]|1[0-2])-(0[1-9]|[12][0-9]|3[01])T([01][0-9]|2[0-3])[0-5][0-9][0-5][0-9]Z$"
```
This will match any 4 digit year, but only possible months (01-12), possible dates (01-31) and possible times (hour: 01-23, minutes: 00-59, seconds: 00-59). We could go stricter on the year as well.

martonvago: How about combining your original approach and your new one?
So in one check something like:
- Check that the first 18 chars are a real datetime using `strptime`. This way you don't have to write a regex that can tell apart real datetimes from nonexistent ones (e.g. 2024-02-31T050000Z looks real but doesn't exist).
- Check the rest of the filename. This bit depends on how strict we want to be (as you wrote above). E.g. check that the rest is a uuid.

src/seedcase_sprout/core/read_resource_batches.py
signekb: We'll scratch that idea then! :)

src/seedcase_sprout/core/read_resource_batches.py
```diff
@@ -0,0 +1,119 @@
+import re
+from pathlib import Path
+
+import polars as pl
+
+from seedcase_sprout.core.check_is_file import check_is_file
+from seedcase_sprout.core.properties import ResourceProperties
+from seedcase_sprout.core.sprout_checks.check_batch_file_name import (
+    check_batch_file_name,
```
signekb, line 10:
```suggestion
    _check_batch_file_name,
```

src/seedcase_sprout/core/read_resource_batches.py
```diff
@@ -0,0 +1,119 @@
...
+    a timestamp column to the DataFrame, extracted from the file name.
+
+    Args:
+        path: Path to the Parquet batch file.
+
+    Returns:
+        The Parquet file as a DataFrame with a timestamp column added.
+    """
+    data = pl.read_parquet(path)
+    check_batch_file_name(path)
```
signekb, line 83:
```suggestion
    _check_batch_file_name(path)
```

tests/core/sprout_checks/test_check_batch_file_name.py
```diff
@@ -0,0 +1,29 @@
+from pathlib import Path
+
+from pytest import raises
+
+from seedcase_sprout.core.create_batch_file_name import create_batch_file_name
+from seedcase_sprout.core.sprout_checks.check_batch_file_name import (
+    check_batch_file_name,
```
signekb, line 7:
```suggestion
    _check_batch_file_name,
```

tests/core/sprout_checks/test_check_batch_file_name.py
```diff
@@ -0,0 +1,29 @@
...
+)
+
+
+def test_file_name_created_by_create_batch_file_name_passes(tmp_path):
+    """A batch file name created by `create_batch_file_name` should pass the check."""
+    # Given
+    batch_file_name = (Path(tmp_path) / create_batch_file_name()).with_suffix(
```
signekb, line 14:
```suggestion
    batch_file_name = (Path(tmp_path) / _create_batch_file_name()).with_suffix(
```

tests/core/sprout_checks/test_check_batch_file_name.py
```diff
@@ -0,0 +1,29 @@
...
+
+
+def test_file_name_that_does_not_fit_the_expected_format_fails(tmp_path):
+    """A batch file name that doesn't fit the expected format should fail the check."""
+    # Given
+    batch_file_name = (Path(tmp_path) / "data").with_suffix(".parquet")
+
+    # When, Then
+    with raises(ValueError):
+        check_batch_file_name(batch_file_name)
```
signekb, line 29:
```suggestion
        _check_batch_file_name(batch_file_name)
```

src/seedcase_sprout/core/sprout_checks/check_batch_file_name.py
signekb: And should the file be renamed as well? 🤔

src/seedcase_sprout/core/sprout_checks/constants.py
signekb: @martonvago Is this to avoid the regex being too complex, or? 

Maybe a nice middle group could be to: 
1. Keep the approach of not expecting the datetime to be in a specific position (so use a rather simple regex like `r"^\d{4}-\d{2}-\d{2}T\d{6}Z"` to extract it from the batch file name)
2. Use `striptime` to check if its a real/possible datetime?

src/seedcase_sprout/core/read_resource_batches.py
signekb: About your comment, @martonvago, about making this a part of a Sprout-specific properties check, I’ve added an issue on this here #1238

tests/core/test_read_resource_batches.py
signekb: No! Will remove 🔥

src/seedcase_sprout/core/sprout_checks/constants.py
martonvago: Yeah, exactly! If we keep making the regex more and more complex, we’ll end up reimplementing strptime.

Re your new plan, I solemnly give my blessing 👼

tests/core/test_read_resource_batches.py
signekb: Good point. I'll change it :+1:

martonvago (approved):
:tada: :tada: :tada:

lwjohnst86 (approved):
:tada:

PR #1207, opened 2025-03-27
docs: :memo: add section on creating the README

### LINKED ISSUE #1120: Add `as_readme_text` to package guide
So far it is only a TODO item.

lwjohnst86:
## Description

Section describing building the README.

Closes #1120


docs/guide/packages.qmd
```diff
@@ -129,32 +131,55 @@ properties you added to it. The newly created file would be:
 print(package_path.glob("**/*"))
 ```
 
-<!-- TODO: Add section on building the README -->
+## Creating a README of the properties
+
+Having a *human-readable* version of what is contained in the
+`datapackage.json` file is useful for others who may be working with or
+learning more about your data package. You can use `as_readme_text()` to
```
signekb, line 138:
```suggestion
wanting to learn more about your data package. You can use `as_readme_text()` to
```

docs/guide/packages.qmd
```diff
@@ -129,32 +131,55 @@ properties you added to it. The newly created file would be:
...
+package you just created:
+
+```{python}
+readme_text = sp.as_readme_text(properties)
+print(readme_text)
+```
+
+It outputs a text string so that, if you want to, you can add other,
+maybe project specific, information before writing to the file. Then to
+save it to the file:
```
signekb, line 150: Add example on what that could be? And how to do it?

signekb (changes requested):

_quarto.yml
```diff
@@ -1,7 +1,9 @@
 project:
   type: seedcase-theme
   pre-render: rm -rf .storage
-  post-render: rm docs/reference/*.qmd
+  post-render:
+    # Delete the auto-generated reference files from quartodoc
+    - find docs/reference/ -type f -name "*.qmd" -o -name "*.quarto_ipynb" -delete
   render:
```
signekb, line 7: Ideally this and the fix about the args in `update_package_properties()` would be separate PRs that could be merged in almost right away. At least, I’m having these issues locally currently when I run `just run-all()`

docs/guide/packages.qmd
lwjohnst86: I think we need to see how we might use it before we add something like this. So I added a clarification text about it.

PR #1212, opened 2025-03-28
feat: :sparkles: add `example_resource_properties()`

signekb:
## Description

This PR adds an example resource properties object that we can use in the “Examples” sections in our docstrings.
Now that we’re adding functions related to resources, I thought it would be nice to have, similarly to the `example_package_resources()`, we already have.

## Checklist

- [X] Added or updated tests
- [X] Ran `just run-all` (fails bc of error in `packages.qmd`—wrong arg is used in `update_package_properties()`—which is fixed in #1207)


signekb:
Thanks for the review, Copilot 😅

src/seedcase_sprout/core/example_properties.py
lwjohnst86, line 1: I think this would be better renamed to `examples.py` and it will contain all example code, including making a dataset, making a full data package later on, etc.

signekb: Damn, why did Copilot not get that? 😉  
(Done 👍 )

lwjohnst86 (approved):

Issue #1053, opened 2025-02-17
Function to make an example data package (with and without resources)


lwjohnst86:
Like the function to make the example properties, this would be useful to have when making the guides.

signekb:
Agreed. And only an example dataset (with one column of each data type we support)

signekb:
Is this a duplicate of #1048 ?

lwjohnst86:
That's a great idea about the "one column each data type"!! 🎉

signekb:
I think it was actually @martonvago’s idea originally 🎉

lwjohnst86:
This has already been done.

PR #1213, opened 2025-03-28
feat: :sparkles: `write_resource_batch()`

### LINKED ISSUE #804: Create and test `write_resource_batches()`
- Signature: `write_resource_data_to_raw(path: Path, data_path: Path) -> Path`

Depends on: 

- [x] #800
- [ ] #801
- [x] #802 
- [x] #803

signekb:
## Description

This PR adds `write_resource_batch()` along with its tests. 

Closes #804 

## Checklist

- [X] Added or updated tests
- [X] Updated documentation
- [X] Ran `just run-all`


tests/core/test_write_resource_batch.py
```diff
@@ -0,0 +1,71 @@
...
+    tmp_path, resource_properties
+):
+    """Throws ExceptionGroup if the data don't match the resource properties."""
+    # Given
+    os.chdir(tmp_path)
+    (Path("resources") / "1").mkdir(parents=True, exist_ok=True)
+
+    # When
+    with raises(ExceptionGroup):
+        write_resource_batch(pl.DataFrame(), resource_properties)
```
signekb, line 71: This basically tests the functionality of `check_resource_properties()` and `check_data()`. Remove?

tests/core/test_write_resource_batch.py
signekb, line 1: I found it difficult to think of what to test for here when we assume the data is tidy and the resource properties and data are checked via other functions. So, if you have any ideas, they are more than welcome 👐

src/seedcase_sprout/core/write_resource_batch.py
signekb, line 1: Not sure how this happened; that Git thinks this was a rename of the file 🤷‍♀️

src/seedcase_sprout/core/write_resource_batch.py
```diff
@@ -0,0 +1,58 @@
...
+            ```
+    """
+    check_resource_properties(resource_properties)
+    # check_data(data, resource_properties)
+
+    # Since `path` is to the `data.parquet` file, take the `parent` to get the `batch` folder
+    batch_dir = Path(resource_properties.path).parent / "batch"
+
+    # Make sure the batch directory exists
+    batch_dir.mkdir(exist_ok=True)
```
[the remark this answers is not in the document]
signekb: Or, maybe we use `resource_batch(self, resource_id: str)` from #1173  ?

src/seedcase_sprout/core/write_resource_batch.py
```diff
@@ -0,0 +1,83 @@
...
+                sp.write_resource_batch(
+                    data=tidy_data, resource_properties=resource_properties
+                )
+            ```
+    """
+    check_resource_properties(resource_properties)
+    # check_data(data, resource_properties)
+
+    # Since `path` is to the `data.parquet` file, take the `parent` to get the `batch` folder
+    batch_dir = Path(resource_properties.path).parent / "batch"
```
lwjohnst86, line 72: This can soon use `PackagePath().resource_batch()` that I'm working on :)

src/seedcase_sprout/core/write_resource_batch.py
```diff
@@ -0,0 +1,83 @@
...
+
+    # Make sure the batch directory exists
+    batch_dir.mkdir(exist_ok=True)
```
lwjohnst86, line 75: We create the directory and file in the `write_` functions.

tests/core/test_write_resource_batch.py
lwjohnst86: I think it's just enough to make sure the function works as we intend it to, so that we can auto-run the tests to make sure nothing breaks when we change things in the future.

And also, a tidy dataframe doesn't mean a dataframe that matches what is described in the resource properties. So you could make a dataframe that has an extra column or has one of the columns changes from int to str or something.

You don't need to add extensive tests for giving an incorrect properties object, but you could make one test where there properties is incorrect itself, and then add a field to the properties that doesn't exist in the dataframe.

Since we already have tests for many of the other functions, you don't need to be too extensive here :relaxed:

lwjohnst86 (changes requested):
Cool! Just some comments :relaxed:

src/seedcase_sprout/core/write_resource_batch.py
```diff
@@ -0,0 +1,83 @@
...
+            resource_properties: The properties object for the specific resource.
+                Use `read_properties()` to read the properties for the resource
+                and `get_resource_properties()` to get the correct resource properties.
+
+        Returns:
+            The path to the written Parquet resource file.
+
+        Raises:
+            ExceptionGroup: A group of `CheckError`s, if resource properties are incorrect.
+            ExceptionGroup: A group of `CheckError`s, if data don't align with the resource properties.
```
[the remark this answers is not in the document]
martonvago: Yes, but will `check_data` throw `CheckError`s? `CheckError`s (as they are now) are meant for errors in properties (longer [comment](https://github.com/seedcase-project/seedcase-sprout/pull/1194#discussion_r2011624597) on this).

src/seedcase_sprout/core/write_resource_batch.py
```diff
@@ -0,0 +1,83 @@
...
+            with tempfile.TemporaryDirectory() as temp_dir:
+                temp_path = Path(temp_dir)
+
+                tidy_data = pl.DataFrame(
+                    {
+                        "id": [0, 1, 2],
+                        "name": ["anne", "belinda", "catherine"],
+                        "value": [1.1, 2.2, 3.3],
+                    }
+                )
```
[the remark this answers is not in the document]
martonvago: :heavy_plus_sign: 
And it could have one column for each data type we support!

tests/core/test_write_resource_batch.py
martonvago: I think it's good to have these 2 tests: you obviously don't have to retest the check functions, but you want to test that they are called in the function you are testing. So just trip them once each, as you're already doing 😊

martonvago (changes requested):
Very nice, just minor comments and a question about the path we're using.

src/seedcase_sprout/core/write_resource_batch.py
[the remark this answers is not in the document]
signekb: I’ll add a TODO here and then, when `check_data` has been implemented, we can update it here.

src/seedcase_sprout/core/write_resource_batch.py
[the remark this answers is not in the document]
signekb: Yess! It makes sense if the example dataset have one of each ➕ 👍

tests/core/test_write_resource_batch.py
```diff
@@ -0,0 +1,71 @@
...
+    os.chdir(tmp_path)
+    (Path("resources") / "1").mkdir(parents=True, exist_ok=True)
+
+    # When
+    batch_path = write_resource_batch(tidy_data, resource_properties)
+    batch_data = pl.read_parquet(batch_path)
+
+    # Then
+    assert batch_path.exists()
+    assert assert_frame_equal(batch_data, tidy_data, check_exact=True) is None
```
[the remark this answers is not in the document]
martonvago: Oh true!!

src/seedcase_sprout/core/write_resource_batch.py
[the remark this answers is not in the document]
signekb: I’ve added a comment on this in #1053

src/seedcase_sprout/core/write_resource_batch.py
signekb: Sooooo, here? :)

tests/core/test_write_resource_batch.py
```diff
@@ -0,0 +1,98 @@
...
+
+    tidy_data = tidy_data.select(["name", "id"])
+
+    # When
+    batch_path = write_resource_batch(tidy_data, resource_properties)
+    batch_data = pl.read_parquet(batch_path)
+
+    # Then
+    assert batch_path.exists()
+    assert_frame_equal(batch_data, tidy_data, check_exact=True)
```
signekb, line 68: Do we care about column order? This could be a part of `check_data()` (if we do care)?

lwjohnst86: No, column order isn't important.

src/seedcase_sprout/core/write_resource_batch.py
```diff
@@ -0,0 +1,85 @@
...
+                    "name": ["anne", "belinda", "catherine"],
+                    "value": [1.1, 2.2, 3.3],
+                }
+            )
+
+            resource_properties = sp.example_resource_properties()
+
+            sp.write_resource_batch(
+                data=tidy_data, resource_properties=resource_properties
+            )
```
signekb, line 65: Oh. When this example is run locally, the batch file is created outside the `tmp_path` since we’re using the path in `example_resource_properties()` to create the path we write to.

martonvago: Ohhh -- is this an example of the cwd =/= package path thing?

src/seedcase_sprout/core/write_resource_batch.py
lwjohnst86: Nvm, ignore me :stuck_out_tongue:

src/seedcase_sprout/core/write_resource_batch.py
signekb: I guess so? Since we use `resource_properties.path` (which is a relative path) to get the path to write to (we use the parent), I had to explicitly set the wd with `os.chdir()` to avoid this behaviour.

NB: This is when I build the website locally.

signekb: Kind of, yes, but not in relation to `PackagePath()`. That’s not used here.

This happens bc I use (the parent of) the relative path from `resource_properties.path` to get the path to write to (i.e., `data.write_parquet(batch_resource_path`)

martonvago (dismissed):
Looks good to me! :rocket:

lwjohnst86 (changes requested):
Nice! Just some small comments :relaxed:

tests/core/test_write_resource_batch.py
```diff
@@ -0,0 +1,95 @@
...
+    # Given
+    os.chdir(tmp_path)
+    (tmp_path / "resources" / resource_properties.name).mkdir(parents=True)
+
+    tidy_data.insert_column(2, pl.Series("extra_column", [1, 2, 3]))
+
+    # When
+    # TODO: What type of error will `check_data()` raise?
+    with raises(ExceptionGroup):
+        write_resource_batch(pl.DataFrame(), resource_properties, tmp_path)
```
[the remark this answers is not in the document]
martonvago: Sorry to butt in here, but would we not want to check that the check functions are called in `write_resource_batch`? So not retest them, just include one test each that trips them.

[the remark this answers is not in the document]
lwjohnst86: Yea, I guess that's a good idea. As long as it only checks that there is an error. No need to check further :relaxed:

tests/core/test_write_resource_batch.py
```diff
@@ -0,0 +1,52 @@
...
+
+    # Then
+    assert batch_path.exists()
+    assert_frame_equal(batch_data, data, check_exact=True)
+
+
+def test_throws_error_if_data_do_not_match_example_resource_properties(tmp_path):
+    """Throws ValueError if data don't match resource properties."""
+    with raises(ValueError):
+        write_resource_batch(pl.DataFrame(), example_resource_properties(), tmp_path)
```
signekb, line 52: @martonvago `check_data()` throws a `ValueError` here, not `ExceptionGroup`. Should it be changed to throw an `ExceptionGroup`? Alternatively, the docstring could be updated?

martonvago: @signekb  It's like:
- bad resource properties -> `CheckError`
- column names don't match -> `ValueError`
- column types don't match -> group of `ValueError`s, one for each mismatched type

But I think it's still up for discussion if we want to change these.
Oh and I see I didn't update the docstrings in #1263 -- I'll do that!

lwjohnst86 (approved):

PR #1239, opened 2025-04-04
refactor: :recycle: clean split of `check_datapackage` sub-package

### LINKED ISSUE #1224: Make clean split of check functionality so it is a "sub-package"
There are a few things that need to be updated from this, most of which is that there are some functions that are used from this sub-package that are not public in it's `__init__.py` file.

Let's also rename the folder to `check_datapackage` in prep for when we split it out.

lwjohnst86:
## Description

This PR makes it so there is a cleaner split with the sub-package and all other code. For instance, keeping a clear external function listing, importing from the sub-package directly rather than from the sub-modules, etc.

Closes #1224

This PR needs a bit of an in-depth review. It's mostly renaming and moving files around.

## Checklist

- [x] Added or updated tests
- [x] Updated documentation
- [x] Ran `just run-all`


martonvago (approved):
Looks good to me!

PR #1242, opened 2025-04-07
refactor: :recycle: make `check_column_types()` take data frame as input

### LINKED ISSUE #1182: Refactor `check_is_` functions in `check_column_data_type` to take a Series, not DataFrame
To make a cleaner abstraction, functions should operate on columns, not be dependent on the data frame it is part of. In Polars, you can make functions apply to Series (columns in Polars) ([here](https://docs.pola.rs/user-guide/expressions/user-defined-python-functions/)), which you can then make use of `map_batches()` to apply those functions to any given column as desired.

Right now, all the `check_column_data_types.py` functions use the field name as argument and output an expression. But the functions can be simplified to input a Series and output a Series, so only need one argument.

This is related to the #1181 discussion.

### LINKED ISSUE #1236: Create an example data frame
Create an example data frame with a column for each data type

### LINKED ISSUE #1240: Create `_check_column_types`

martonvago:
## Description

This PR refactors the check for column data types to take a data frame as input. This includes:
- adding a `check_column_types` function
- removing old column data type check functions
- adding an example data frame object (added here because it's used in the tests)

Closes #1240, closes #1236, closes #1182

## Checklist

- [x] Added or updated tests
- [x] Updated documentation
- [x] Ran `just run-all`


src/seedcase_sprout/core/sprout_checks/_check_column_types.py
```diff
@@ -0,0 +1,121 @@
...
+    allowed_types: str
+
+
+_STRING_TYPE_CHECK: _PolarsTypeCheck = {
+    "check": lambda dtype: dtype == pl.String,
+    "allowed_types": f"'{pl.String}'",
+}
+
+# Mapping from Frictionless field types to checks on Polars data types
+_FRICTIONLESS_TO_POLARS_TYPE_CHECK: dict[FieldType, _PolarsTypeCheck] = {
```
martonvago, line 70: So far we have been asking the question "given a Frictionless type, what Polars type should we map to?". This is a bit different here. It's asking "which Polars types are allowable realisations of a given Frictionless type?".

Take the Frictionless `integer` type as an example. So far we've said we're mapping this to `pl.Int64`. But now we're asking if this is the only Polars type that can reasonably correspond to `integer`. Arguably not the only one, as there are other integer types in Polars (e.g. `pl.Int8`, `pl.Int16`, `pl.Int32`, etc.), and these shouldn't raise an error either.

So the mapping below tries to be as permissive as possible. For many types the mapping is 1-to-1, but there are some where it's a bit more complex.

The `check` property contains a function that returns True if the given Polars type is an acceptable way to express the given Frictionless type. The `allowed_types` property contains a textual description of what Polars types are allowed for the Frictionless type.

src/seedcase_sprout/core/sprout_checks/_check_column_types.py
```diff
@@ -0,0 +1,121 @@
...
+    "geopoint": {
+        "check": lambda dtype: (
+            isinstance(dtype, pl.Array) and dtype.size == 2 and dtype.inner.is_numeric()
+        ),
+        "allowed_types": "an array of a numeric type with size 2",
+    },
+    "duration": _STRING_TYPE_CHECK,
+    "object": _STRING_TYPE_CHECK,
+    "array": _STRING_TYPE_CHECK,
+    "geojson": _STRING_TYPE_CHECK,
```
martonvago, line 116: Up for discussion if we want to accept other Polars types for these than `pl.String`. `pl.Object` or any of their [nested types](https://docs.pola.rs/api/python/stable/reference/datatypes.html#nested)?

tests/core/sprout_checks/test_check_column_types.py
```diff
@@ -0,0 +1,120 @@
...
+    resource_properties.schema.fields += [
+        FieldProperties(name="my_categorical", type="string"),
+        FieldProperties(name="my_enum", type="string"),
+        FieldProperties(name="my_int8", type="integer"),
+        FieldProperties(name="my_int8_year", type="year"),
+        FieldProperties(name="my_uint64", type="integer"),
+        FieldProperties(name="my_float32", type="number"),
+        FieldProperties(name="my_decimal", type="number"),
+        FieldProperties(name="my_int_geopoint", type="geopoint"),
+    ]
```
martonvago, line 26: Just adding some extra test cases to the example set

src/seedcase_sprout/core/sprout_checks/_check_column_types.py
martonvago: Let me know if you don't like this dictionary-based mapping and would prefer it in another form!

lwjohnst86 (changes requested):
There's actually three separate things happening in this PR: examples, removing `check_column_data_types()`, and refactoring the `check_column_types()`. Could you split them out into three PRs? It doesn't need to be stacked either. It can be independent ones (even if one depends on the examples)

PR #1246, opened 2025-04-08
feat: :sparkles: `check_data()` with column type checks

### LINKED ISSUE #1182: Refactor `check_is_` functions in `check_column_data_type` to take a Series, not DataFrame

### LINKED ISSUE #1240: Create `_check_column_types`

martonvago:
## Description

This PR adds a new  `_check_column_types` function to the `check_data()`

Closes #1240, closes #1182

## Checklist

- [x] Added or updated tests
- [x] Updated documentation
- [x] Ran `just run-all`


src/seedcase_sprout/core/sprout_checks/_check_column_types.py
```diff
@@ -0,0 +1,121 @@
...
+
+    Args:
+        data: The data frame to check.
+        resource_properties: The resource properties to check against.
+
+    Returns:
+        The data frame, if all column types are correct.
+
+    Raises:
+        ExceptionGroup: A group of `ValueError`s, one per incorrectly typed column.
```
martonvago, line 32: Example output:
```
  | ExceptionGroup: The following columns in the data have data types that do not match the data types in the resource properties: (2 sub-exceptions)
  +-+---------------- 1 ----------------
    | ValueError: Expected type of column 'id' to be an integer type but found 'String'.
    +---------------- 2 ----------------
    | ValueError: Expected type of column 'colour' to be one of '(String, Categorical, Enum)' but found 'Int64'.
```

src/seedcase_sprout/core/sprout_checks/_check_column_types.py
```diff
@@ -0,0 +1,121 @@
...
+    all_column_props: list[FieldProperties] = get_nested_attr(
+        resource_properties, "schema.fields", default=[]
+    )
+    errors = []
+    for polars_type, column_props in zip(data.schema.dtypes(), all_column_props):
+        type_check = _FRICTIONLESS_TO_POLARS_TYPE_CHECK[column_props.type or "any"]
+        check, allowed_types = type_check["check"], type_check["allowed_types"]
+        if not check(polars_type):
+            errors.append(
+                ValueError(
```
martonvago, line 43: This is a very bland error to throw, so suggestions welcome. I know there's a `TypeError` but I couldn't quite convince myself that it is appropriate here.

src/seedcase_sprout/core/sprout_checks/_check_column_types.py
```diff
@@ -0,0 +1,121 @@
...
+    allowed_types: str
+
+
+_STRING_TYPE_CHECK: _PolarsTypeCheck = {
+    "check": lambda dtype: dtype == pl.String,
+    "allowed_types": f"'{pl.String}'",
+}
+
+# This mapping gives the allowed Polars types for each Frictionless type
+_FRICTIONLESS_TO_POLARS_TYPE_CHECK: dict[FieldType, _PolarsTypeCheck] = {
```
martonvago, line 70: So far we have been asking the question "given a Frictionless type, what Polars type should we map to?". This is a bit different here. It's asking "which Polars types are allowable realisations of a given Frictionless type?".

Take the Frictionless `integer` type as an example. So far we've said we're mapping this to `pl.Int64`. But now we're asking if this is the only Polars type that can reasonably correspond to `integer`. Arguably not the only one, as there are other integer types in Polars (e.g. `pl.Int8`, `pl.Int16`, `pl.Int32`, etc.), and these shouldn't raise an error either.

So the mapping below tries to be as permissive as possible. For many types the mapping is 1-to-1, but there are some where it's a bit more complex.

The `check` property contains a function that returns True if the given Polars type is an acceptable way to express the given Frictionless type. The `allowed_types` property contains a textual description of what Polars types are allowed for the Frictionless type.

martonvago: Let me know if you don't like this dictionary-based mapping and would prefer it in another form!

src/seedcase_sprout/core/sprout_checks/_check_column_types.py
```diff
@@ -0,0 +1,121 @@
...
+    "geopoint": {
+        "check": lambda dtype: (
+            isinstance(dtype, pl.Array) and dtype.size == 2 and dtype.inner.is_numeric()
+        ),
+        "allowed_types": "an array of a numeric type with size 2",
+    },
+    "duration": _STRING_TYPE_CHECK,
+    "object": _STRING_TYPE_CHECK,
+    "array": _STRING_TYPE_CHECK,
+    "geojson": _STRING_TYPE_CHECK,
```
martonvago, line 116: Up for discussion if we want to accept other Polars types for these than `pl.String`. `pl.Object` or any of their [nested types](https://docs.pola.rs/api/python/stable/reference/datatypes.html#nested)?

tests/core/sprout_checks/test_check_column_types.py
```diff
@@ -0,0 +1,120 @@
...
+    resource_properties = example_resource_properties_all_types()
+    resource_properties.schema.fields += [
+        FieldProperties(name="my_categorical", type="string"),
+        FieldProperties(name="my_enum", type="string"),
+        FieldProperties(name="my_int8", type="integer"),
+        FieldProperties(name="my_int8_year", type="year"),
+        FieldProperties(name="my_uint64", type="integer"),
+        FieldProperties(name="my_float32", type="number"),
+        FieldProperties(name="my_decimal", type="number"),
+        FieldProperties(name="my_int_geopoint", type="geopoint"),
```
martonvago, line 25: Just adding some extra test cases to the example set

src/seedcase_sprout/core/sprout_checks/_check_column_types.py
```diff
@@ -0,0 +1,118 @@
...
+
+
+class _PolarsTypeCheck(TypedDict):
+    """Typed dictionary representing a check on a Polars type."""
+
+    check: Callable[[pl.DataType], bool]
+    allowed_types: str
+
+
+def _get_check_for_polars_type(expected_dtype: pl.DataType) -> _PolarsTypeCheck:
```
martonvago, line 66: Polars often calls their data types `dtype`

src/seedcase_sprout/core/sprout_checks/_check_column_types.py
```diff
@@ -0,0 +1,118 @@
...
+    """
+    all_column_props: list[FieldProperties] = get_nested_attr(
+        resource_properties, "schema.fields", default=[]
+    )
+    errors = []
+    for polars_type, column_props in zip(data.schema.dtypes(), all_column_props):
+        type_check = _FRICTIONLESS_TO_POLARS_TYPE_CHECK[column_props.type or "any"]
+        check, allowed_types = type_check["check"], type_check["allowed_types"]
+        if not check(polars_type):
+            errors.append(
```
lwjohnst86, line 42: This needs to be redesigned to be more functional, there's a lot going on here. This is actually a good chance for us to take a step back and think about the design a bit more here.

src/seedcase_sprout/core/sprout_checks/_check_column_types.py
```diff
@@ -0,0 +1,118 @@
...
+        data: The data frame to check.
+        resource_properties: The resource properties to check against.
+
+    Returns:
+        The data frame, if all column types are correct.
+
+    Raises:
+        ExceptionGroup: A group of `ValueError`s, one per incorrectly typed column.
+    """
+    all_column_props: list[FieldProperties] = get_nested_attr(
```
lwjohnst86, line 34:
```suggestion
    field_properties: list[FieldProperties] = get_nested_attr(
```

Or `column_properties`

martonvago:
Putting this into "in progress" while I update it to match [this change](https://github.com/seedcase-project/seedcase-sprout/pull/1247#discussion_r2034553422)

src/seedcase_sprout/core/sprout_checks/_check_column_types.py
martonvago: If you don't mind using "field" in the name, then I'll just go for `fields`

src/seedcase_sprout/core/sprout_checks/_check_column_types.py
martonvago: Let me know if the new version is better

tests/core/sprout_checks/test_check_column_types.py
```diff
@@ -0,0 +1,144 @@
...
+    resource_properties = example_resource_properties()
+    resource_properties.schema.fields = [
+        FieldProperties(name="my_col", type=frictionless_type)
+    ]
+
+    with raises(ExceptionGroup) as error_info:
+        _check_column_types(data, resource_properties)
+
+    errors = error_info.value.exceptions
+    assert len(errors) == 1
```
signekb, line 95: Could also include a test of whether the type is included in the error message, maybe? :) 
(same for tests below)

signekb (changes requested):
Very nice 🌟 This overall looks good to me! I have a couple of suggestions.

src/seedcase_sprout/core/sprout_checks/_check_column_types.py
```diff
@@ -0,0 +1,126 @@
...
+
+
+def _check_column_types(
+    data: pl.DataFrame, resource_properties: ResourceProperties
+) -> pl.DataFrame:
+    """Checks that column data types match the data types specified in the properties.
+
+    The resource properties specify a Frictionless data type for each column.
+    This function checks if the Polars data type of each column in the data is an
+    allowed realisation of the corresponding Frictionless data type.
```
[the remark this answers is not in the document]
martonvago: I changed it to something similar but not exactly this. :+1:

tests/core/sprout_checks/test_check_column_types.py
martonvago: I added this in 2 tests! (Not sure if adding it in all tests would achieve very much :shrug: )

src/seedcase_sprout/core/examples.py
```diff
@@ -214,8 +214,8 @@ def example_data_all_types() -> pl.DataFrame:
                 '{"outer": "value", "inner": {"prop1": 123, "prop2": true}}',
             ],
             "my_string": ["some text", "æøåäöü£$%^&*()\\''", "μῆνιν ἄειδε θεὰ"],
-            "my_any": pl.Series([{"prop": "value"}] * 3, dtype=pl.Object),
-            "my_none": pl.Series([{"prop": "value"}] * 3, dtype=pl.Object),
+            "my_any": pl.Series([[1]] * 3, dtype=pl.List),
+            "my_none": pl.Series([[1]] * 3, dtype=pl.List),
```
martonvago, line 218: Changed this because `assert_frame_equal` can't compare `Object`s

src/seedcase_sprout/core/sprout_checks/_check_column_types.py
```diff
@@ -0,0 +1,126 @@
...
+        A `ValueError`.
+    """
+    allowed_types = _FRICTIONLESS_TO_ALLOWED_POLARS_TYPES[field.type or "any"]
+    return ValueError(
+        f"Expected type of column '{field.name}' "
+        f"to be {allowed_types} but found '{polars_type}'."
+    )
+
+
+# Mapping from Frictionless types to check functions for allowed Polars types
```
martonvago, line 88: This mapping is the meat of this PR

src/seedcase_sprout/core/sprout_checks/_check_column_types.py
lwjohnst86, line 1: After doing the refactoring in #1255 I think we need to be more strict about building backwards so we have a clear endgoal.

So, could you revise this to be the `check_data.py` file, move the pseudocode over into here, expose it as a public function, and start implementing that function with what you've done here. So you won't need to change the current `_check_column_types()` :relaxed:

src/seedcase_sprout/core/sprout_checks/_check_column_types.py
```diff
@@ -0,0 +1,126 @@
...
+    """
+    fields: list[FieldProperties] = get_nested_attr(
+        resource_properties, "schema.fields", default=[]
+    )
+    polars_schema = data.schema
+    errors = [
+        _get_column_type_error(polars_schema[field.name], field)
+        for field in fields
+        if _not_allowed_type(polars_schema[field.name], field)
+    ]
```
lwjohnst86, line 42: This is much better, thanks! Though still not very functional. At least it's easy to see it outputs as a list :grin: I think this is fine now, moving the codebase to a more functional style will take some time :relaxed:

tests/core/sprout_checks/test_check_column_types.py
lwjohnst86, line 1: Rather than have this file be for the column types, can you rename to `test_check_data.py` and internally only use the `check_data()` function rather than the private `_check_column_types()`? I've realized that we probably shouldn't be testing private functions and rather test them indirectly through public functions (esp after having to debug all the private tests after refactoring the check properties!!).

lwjohnst86 (changes requested):
Very nice! Some mostly structural requests :relaxed:

src/seedcase_sprout/core/sprout_checks/_check_column_types.py
martonvago: But do you just mean (here, at least) using `map` and `filter`? In this case wouldn't that be more complicated to read than a regular list comprehension?

src/seedcase_sprout/core/sprout_checks/_check_column_types.py
martonvago: I thiiiink I understand :sweat_smile: 
So just call `_check_column_types` from `check_data`. And then test `check_data`?

lwjohnst86 (approved):
Niiicccceeee :grin: :dancers:

PR #1247, opened 2025-04-08
feat: :sparkles: check column names in `check_data()`

### LINKED ISSUE #1241: Refactor `_check_column_names`

martonvago:
## Description

This PR updates `check_column_names()` to match the design and includes it into `check_data()`.

Closes #1241

## Checklist

- [x] Added or updated tests
- [x] Updated documentation
- [x] Ran `just run-all`


src/seedcase_sprout/core/sprout_checks/_check_column_names.py
```diff
@@ -0,0 +1,59 @@
...
+    Args:
+        data: The data to check.
+        resource_properties: The resource properties to check against.
+
+    Returns:
+        The data if the column names match.
+
+    Raises:
+        ValueError: If the column names are not the same as the names in
+            `resource_properties`.
```
martonvago, line 21: Example output:
```
E   ValueError: Column names in the data do not match column names in the resource properties:
E   - Extra columns: ['extra_col']
E   - Missing columns: ['my_string', 'my_number']
```

```
E   ValueError: The column names in the data are in the wrong order. Expected ['my_bool', 'my_string'] but found ['my_string', 'my_bool'].
```

tests/core/sprout_checks/test_check_column_names.py
martonvago, line 1: This file was renamed and modified

src/seedcase_sprout/core/sprout_checks/_check_column_names.py
```diff
@@ -0,0 +1,59 @@
...
+) -> str:
+    extra_columns_in_data = [
+        name for name in names_in_data if name not in names_in_resource
+    ]
+    missing_columns_in_data = [
+        name for name in names_in_resource if name not in names_in_data
+    ]
+    if not extra_columns_in_data and not missing_columns_in_data:
+        return (
+            "The column names in the data are in the wrong order. Expected"
```
lwjohnst86, line 48: Hm, I'm not sure what you mean by "wrong order" here. Do you mean "the data you provided doesn't match with what is in the properties"? The order that they are in doesn't matter.

lwjohnst86 (changes requested):
Nice! Only one question :relaxed:

src/seedcase_sprout/core/sprout_checks/_check_column_names.py
martonvago: I was just going by the Frictionless [default](https://datapackage.org/standard/table-schema/#fieldsMatch) that seems to say the columns have to be in the same order. But we can obviously relax this. I'll just quickly update PR #1246 if we do.

lwjohnst86: That is a super weird default behaviour. But looks like it's easy to change to `equal` intead.

signekb:
Reviewing this today 👓

src/seedcase_sprout/core/sprout_checks/_check_column_names.py
```diff
@@ -0,0 +1,57 @@
...
+
+    Raises:
+        ValueError: If the column names don't match the names in
+            `resource_properties`.
+    """
+    names_in_data = data.schema.names()
+    names_in_resource = [
+        field.name
+        for field in get_nested_attr(resource_properties, "schema.fields", default=[])
+    ]
```
signekb, line 29: Add `column_` or `col_` before these variable names to be clearer?

src/seedcase_sprout/core/sprout_checks/_check_column_names.py
```diff
@@ -0,0 +1,57 @@
...
+            _format_column_name_error_message(
+                extra_columns_in_data, missing_columns_in_data
+            )
+        )
+
+    return data
+
+
+def _format_column_name_error_message(
+    extra_columns_in_data: list[str], missing_columns_in_data: list[str]
```
signekb, line 48:
```suggestion
    extra_columns_in_data: list[str] | None , missing_columns_in_data: list[str] | None
```
I guess?

src/seedcase_sprout/core/sprout_checks/_check_column_names.py
```diff
@@ -0,0 +1,57 @@
...
+) -> str:
+    message = (
+        "Column names in the data do not match column names in the resource properties:"
+    )
+    if extra_columns_in_data:
+        message += f"\n- Extra columns: {extra_columns_in_data}"
+    if missing_columns_in_data:
+        message += f"\n- Missing columns: {missing_columns_in_data}"
```
signekb, line 56:
```suggestion
        message += f"\n- Column(s) in data but not in properties: {extra_columns_in_data}"
    if missing_columns_in_data:
        message += f"\n- Column(s) in properties missing in data: {missing_columns_in_data}"
```
Maybe? :)

tests/core/sprout_checks/test_check_column_names.py
```diff
@@ -0,0 +1,97 @@
...
+
+@fixture
+def resource_properties() -> ResourceProperties:
+    return ResourceProperties(
+        name="data",
+        title="data",
+        path=str(Path("resources", "1", "data.csv")),
+        description="My data...",
+        schema=TableSchemaProperties(),
+    )
```
signekb, line 27: Could the examples (from `examples.py`) be used here? 🤷‍♀️

tests/core/sprout_checks/test_check_column_names.py
```diff
@@ -0,0 +1,97 @@
...
+
+    assert _check_column_names(df, resource_properties) is df
+
+
+def test_throws_error_when_data_has_extra_columns(resource_properties):
+    """Should throw an error if the data has extra columns."""
+    df = pl.DataFrame({"extra_col1": [""], bool_field.name: [True], "extra_col2": [""]})
+    resource_properties.schema.fields = [bool_field]
+
+    with raises(ValueError, match=r"Extra columns.*extra_col1.*extra_col2") as error:
```
signekb, line 57: Do we want to avoid using exact wording from error messages in tests? 🤔 Maybe only match the names of the cols? .. But I also get that you want to test that it’s the expected part of the error message that is thrown.

signekb (changes requested):
Looks great 🌟  Just some small suggestions that you can implement if you agree (otherwise you can just resolve them).

src/seedcase_sprout/core/sprout_checks/_check_column_names.py
martonvago: But if these are empty aren't they `[]`?

src/seedcase_sprout/core/sprout_checks/_check_column_names.py
martonvago: How about:
"Unexpected column(s) in data:"
"Missing column(s) in data:"
?

tests/core/sprout_checks/test_check_column_names.py
martonvago: Yeah, this is an older file from before!

tests/core/sprout_checks/test_check_column_names.py
martonvago: Yeah, I think the message is complex enough that we should check that extra and missing columns don't get mixed up! Otherwise I would agree

src/seedcase_sprout/core/sprout_checks/_check_column_names.py
lwjohnst86, line 1: Following the suggestion to build this into the `check_data.py` file that I made in #1246, can you do the same thing here? For both this file and the tests. :relaxed:

src/seedcase_sprout/core/sprout_checks/_check_column_names.py
lwjohnst86: It also doesn't matter as much for internal functions.

src/seedcase_sprout/core/sprout_checks/_check_column_names.py
lwjohnst86: I like either :relaxed:

lwjohnst86 (dismissed):
Aside from the error message clarification from @signekb, looks good to me! :grin: :horse:

martonvago:
Okay, should be done now!

lwjohnst86 (approved):

PR #1248, opened 2025-04-08
build: :wrench: switch to using uv, rather than poetry

### LINKED ISSUE #1036: Convert `pyproject.toml` to follow PEP 621

lwjohnst86:
## Description

The `pyproject.toml` file needed to be updated to match PEP621. So in the process, I also converted things to use uv. This hopefully will also fix our issue we've had with commitizen bumping other packages too.

Closes #1036

## Checklist

- [x] Added or updated tests
- [x] Updated documentation
- [x] Ran `just run-all`


lwjohnst86:
It fails right now because the GitHub workflow uses Poetry. Updating that right now.

lwjohnst86:
Well, switching to uv already shortened the build times from around 50 sec to 40 sec. Not crazy different, but different!

pyproject.toml
```diff
@@ -1,55 +1,68 @@
-[tool.poetry]
+[project]
 name = "seedcase-sprout"
 version = "0.29.1"
 description = "Grow your research data in a structured, modern way that follows best practices."
 authors = [
```
lwjohnst86, line 5: I removed the others here because at this point, nothing of what we did with them is kept/used. Thoughts though?

martonvago (approved):
I have no background knowledge about this, but I tried it and it works! :tada: :star:

lwjohnst86:
I'll merge this in, to let the builds pass from now on.

pyproject.toml
signekb: Makes sense to me!

PR #1256, opened 2025-04-09
test: :fire: remove duplicate tests on `check_*` functionality

lwjohnst86:
## Description

If we extensively test a functionality that another function uses internally, we don't need to also duplicate the tests again in the other function. Since the move to refactor `check_properties()` and having the more extensive tests there, these tests are not needed any more as they are duplicates.

This PR needs a quick review.

## Checklist

- [x] Added or updated tests
- [x] Ran `just run-all`


tests/core/test_update_package_properties.py
```diff
@@ -101,90 +101,3 @@ def test_current_resources_not_modified(properties_updates):
     )
 
     assert updated_properties == current_properties
-
-
-def test_throws_error_if_current_properties_have_a_package_error():
```
martonvago, line 106: Keep this (for example) to test that current properties are checked. (This error will trigger the first `check_properties` call.)

tests/core/test_update_package_properties.py
```diff
@@ -101,90 +101,3 @@ def test_current_resources_not_modified(properties_updates):
...
-
-    assert "my-package" in error_info.value.message
-    assert "invalid name with spaces" in error_info.value.message
-    errors = error_info.value.exceptions
-    assert len(errors) == 1
-    assert errors[0].json_path == "$.resources[0].name"
-    assert errors[0].validator == "pattern"
-
-
-def test_throws_error_if_incoming_package_properties_are_malformed():
```
martonvago, line 149: Keep this (but rename it) to test that resulting properties are checked. (This error will trigger the second `check_properties` call.)

tests/core/test_write_package_properties.py
martonvago, line 1: I would keep one test to make sure that we check properties before we write them.

tests/core/test_write_resource_properties.py
```diff
@@ -134,77 +133,3 @@ def test_throws_error_if_properties_file_cannot_be_read(
 
     with raises(JSONDecodeError):
         write_resource_properties(file_path, resource_properties_1)
-
-
-def test_throws_error_if_resource_properties_have_missing_required_fields(
```
martonvago, line 139: Keep this (for example) to test that input resource properties are checked.

tests/core/test_write_resource_properties.py
```diff
@@ -134,77 +133,3 @@ def test_throws_error_if_properties_file_cannot_be_read(
...
-    with raises(ExceptionGroup) as error_info:
-        write_resource_properties(package_properties_path, resource_properties_2)
-
-    errors = error_info.value.exceptions
-    assert len(errors) == 1
-    assert errors[0].validator == "pattern"
-    assert errors[0].json_path == "$.resources[0].path"
-
-
-def test_throws_error_if_package_properties_have_missing_required_fields(
```
martonvago, line 198: Keep this to test that the package properties are checked.

martonvago (changes requested):

martonvago (approved):

PR #1263, opened 2025-04-11
feat: :sparkles: use `check_data()` in `read_resource_batches()`

### LINKED ISSUE #1237: Create `check_data()` function
There's pseudocode of this already

martonvago:
## Description

This PR finishes `check_data()` and adds it to `read_resource_batches()` plus:

- tidies tests
- uncomments references in calling code
- fixes resulting errors

Closes #1237 

## Checklist

- [x] Added or updated tests
- [x] Updated documentation
- [x] Ran `just run-all`

### DIFF
--- a/src/seedcase_sprout/core/read_resource_batches.py
+++ b/src/seedcase_sprout/core/read_resource_batches.py
@@ -11,8 +11,7 @@ from seedcase_sprout.core.constants import (
     BATCH_TIMESTAMP_PATTERN,
 )
 from seedcase_sprout.core.properties import ResourceProperties
-
-# from seedcase_sprout.core.checks.check_data import check_data
+from seedcase_sprout.core.sprout_checks.check_data import check_data
 from seedcase_sprout.core.sprout_checks.check_properties import (
     check_resource_properties,
 )
@@ -87,8 +86,7 @@ def read_resource_batches(
 
     data_list = list(map(_read_parquet_batch_file, paths))
 
-    # TODO: Uncomment and test when `check_data` is implemented
-    # list(map(check_data, data_list, resource_properties))
+    list(map(lambda data: check_data(data, resource_properties), data_list))
 
     return data_list
 
--- a/src/seedcase_sprout/core/read_resource_batches.py
+++ b/src/seedcase_sprout/core/read_resource_batches.py
@@ -83,15 +83,12 @@ def read_resource_batches(
     """
     list(map(check_is_file, paths))
     check_resource_properties(resource_properties)
-
-    data_list = list(map(_read_parquet_batch_file, paths))
-
-    list(map(lambda data: check_data(data, resource_properties), data_list))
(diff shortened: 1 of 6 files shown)


src/seedcase_sprout/core/read_resource_batches.py
martonvago, line 1: The problem here was that `check_data` was called after adding the timestamp column, which made it fail. An easy fix was to call it before adding the column.

tests/core/assert_raises_errors.py
martonvago, line 1: Just a helper function for exception groups

tests/core/test_examples.py
```diff
@@ -21,15 +24,11 @@ def test_creates_correct_resource_properties_object_all_types():
...
 
 
-def xtest_example_data_matches_resource_properties():
-    # TODO: Update when `check_data()` has been finished
-    # data = example_data()
-    # assert check_data(data, example_resource_properties()) is data
-    ...
+def test_example_data_matches_resource_properties():
+    data = example_data()
+    assert check_data(data, example_resource_properties()) is data
```
martonvago, line 29: We just care about this not throwing an error, so we don't need `assert_frame_equal`

src/seedcase_sprout/core/read_resource_batches.py
```diff
@@ -84,28 +83,27 @@ def read_resource_batches(
     """
     list(map(check_is_file, paths))
     check_resource_properties(resource_properties)
+    return [_read_parquet_batch_file(path, resource_properties) for path in paths]
```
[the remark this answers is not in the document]
lwjohnst86: You don't need a lambda here:

```
list(map(_read_parquet_batch_file, paths, resource_properties))
```

Which is nice :relaxed: :stuck_out_tongue:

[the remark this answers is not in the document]
martonvago: I think that wouldn't work because `resource_properties` is not an iterable.
But you could do something like:
```python
from itertools import repeat
list(map(_read_parquet_batch_file, paths, repeat(resource_properties)))
```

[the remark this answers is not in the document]
lwjohnst86: Hmm, wow, Python's built-in tools for this are really lacking :anger: :rage: 

Ok, leave as is, but I'm going to spend some time creating some internal, easier functional programming tools and then refactor some things to make use of them. It's hard to show how powerful these are when Python seems to actively make them sub-optimal from it's interface and usage. So hopefully it will be easier to see how useful they are when they are designed better (I'll be just taking inspiration from R here, they do it very well).

tests/core/assert_raises_errors.py
lwjohnst86, line 1: Very nice!

lwjohnst86 (approved):

Issue #1210, opened 2025-03-28
Have path argument default to the appropriate path relative to the cwd


martonvago:
We've standardised the order of arguments in functions that take a path so that the path comes first.
I think this makes sense logically/stylistically, but, with the local-first approach, wouldn't we want to provide defaults for the path?
In that case, the path has to go last.

For example, let's take `write_resource_properties`:
```python
# Without default
properties_path = ...
resource_properties = sp.example_resource_properties()
sp.write_resource_properties(properties_path, resource_properties)

# With default
resource_properties = sp.example_resource_properties()
sp.write_resource_properties(resource_properties)
```

The default path being the path to `datapackage.json` in the cwd (or the cwd itself, according to taste): 
```python
# Without a SproutPath class
def write_resource_properties(
    resource_properties: ResourceProperties,
    path: Path = Path.cwd() / "datapackage.json",
) -> Path:

# With a SproutPath class
def write_resource_properties(
    resource_properties: ResourceProperties,
    path: Path = SproutPath().properties(),
) -> Path:
```

lwjohnst86:
We actually have properties or data come first, not the path. But this idea of using a default path is fantastic!!

martonvago:
Yay!!

PR #1265, opened 2025-04-14
feat: :sparkles: add example package

### LINKED ISSUE #1048: Function to create an example data package
Some of the docstring example codes could really benefit from having template code that generates an example data package (like the example properties) with and without resources.

martonvago:
## Description

This PR adds an example package class that can be used like:
```python
with ExamplePackage() as package_path:
    properties = read_properties(package_path.properties())
    ...
```

- can take a `with_resources` flag
- creates datapackage.json, README, resources folder, folder for one resource (if `with_resources` is True)
- puts everything into a temporary folder
- cleans up after itself

It's easy to add more customisations. 
It's easy to add data batches and data.parquet if we want.

I'll update example code to use this after it's OK'd

Closes #1048 

## Checklist

- [x] Added or updated tests
- [x] Updated documentation
- [x] Ran `just run-all`


src/seedcase_sprout/core/example_package.py
```diff
@@ -0,0 +1,87 @@
...
+            package_properties.resources = [resource_properties]
+
+            # Create resource folders
+            # TODO: update after resource creation is refactored
+            package_path.resources().mkdir(exist_ok=True)
+            resource_path, _ = create_resource_structure(path=package_path.path)
+            resource_properties = create_resource_properties(
+                path=resource_path, properties=resource_properties
+            )
+            resource_properties.name = resource_path.stem
```
martonvago, line 70: As the TODO says, this is using the current flow and will need to be updated

src/seedcase_sprout/core/example_package.py
```diff
@@ -0,0 +1,87 @@
...
+
+        # Save properties
+        write_package_properties(
+            properties=package_properties, path=package_path.properties()
+        )
+
+        # Write README
+        write_file(as_readme_text(package_properties), package_path.readme())
+
+        os.chdir(package_path.path)
```
martonvago, line 80: Do we want to add a `root()` method in `PackagePath`? So that we could write `package_path.root()`, matching the pattern of other paths?

martonvago: I'm changing the cwd to the package root so that functions called inside this context can use their default path args (when that is implemented).
I.e.:
```python
with ExamplePackage():
    properties = read_properties()
```
instead of
```python
with ExamplePackage() as package_path:
    properties = read_properties(package_path.properties())
```

lwjohnst86: I think that's a nice idea about the `root()`! :grin:

tests/core/test_example_package.py
lwjohnst86, line 1: Move this over into the `test_examples.py` file.

The tests inside that file (plus these tests) could actually form the basis of an "end to end" testing script. We could even rename the file to `test_end_to_end.py` :stuck_out_tongue:

lwjohnst86 (changes requested):
Verrryyy nice! :+1: Just some small comments, mostly about putting into the `examples.py` file with the other examples

src/seedcase_sprout/core/read_resource_batches.py
martonvago, line 1: This was breaking the build because it was not using the correct data type. I'll add `ExamplePackage` to the examples in another PR.

signekb: Ah! This was fixed in #1269

tests/core/test_examples.py
```diff
@@ -32,3 +40,80 @@ def test_example_data_matches_resource_properties():
...
+    assert not package_path.path.exists()
+
+
+# TODO: run test when path functions have default paths
+def xtest_can_use_context_without_referencing_package_path_explicitly():
+    """It should be possible to omit the package path from function calls within the
+    context."""
+    with ExamplePackage():
+        properties = read_properties()
+        assert check_properties(properties)
```
signekb, line 119: Since #1266 has been merged, is it time for this being added? 🚀

signekb (dismissed):
Nice! Looks good to me 🚀  I’ve added a comment about executing the last test.

tests/core/test_examples.py
martonvago: This is still pending #1210 alas

lwjohnst86 (approved):
Very nice :tada: :stars:

PR #1266, opened 2025-04-14
refactor: :recycle: get cwd at call time in `PackagePath`

martonvago:
## Description

This PR makes `PackagePath` usable in more situations by setting the default path (the cwd) at call time (not at definition time).

For a use case, see the test I added.

This PR needs a quick review.

## Checklist

- [x] Added or updated tests
- [x] Updated documentation
- [x] Ran `just run-all`


lwjohnst86 (approved):
Oooo very nice!!

PR #1270, opened 2025-04-14
refactor: :recycle: create a simpler map functional

### LINKED ISSUE #1267: Create simpler, internal FP tools
The built-in Python implementation of FP is, well, very much less than desirable. So create an internal set that mimics how the R package purrr does it (and R in general).

lwjohnst86:
## Description

It bothers me to no end that Pythons built-in functionals are so user-unfriendly. So this is the start of some refactoring I will do, with simpler user-friendly versions of map that are inspired from the R package purrr, which implements and designs functionals so nicely. So these functionals are mostly a copy of what purrr has.

I will be adding other functionals here over time.

Closes #1267 

This PR needs a quick review.

## Checklist

- [x] Added or updated tests
- [x] Updated documentation
- [x] Ran `just run-all`

### DIFF
--- a/src/seedcase_sprout/core/create_dirs.py
+++ b/src/seedcase_sprout/core/create_dirs.py
@@ -1,6 +1,6 @@
 from pathlib import Path
 
-import seedcase_sprout.core.fp as fp
+import seedcase_sprout.core.internals as _map
 
 
 def create_dir(path: Path) -> Path:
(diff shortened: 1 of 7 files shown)


src/seedcase_sprout/core/create_dirs.py
```diff
@@ -1,5 +1,7 @@
 from pathlib import Path
 
+import seedcase_sprout.core.fp as fp
```
lwjohnst86, line 3: I don't love the name `fp` and the way this is imported, so any ideas would be appreciated!

Alternatively, we have a new folder called `internals` that has a `__init__.py` that imports internal functions so we could do:

``` python
import seedcase_sprout.core.internals
```

Then use the function as `_map()` directly.

martonvago (changes requested):

PR #1273, opened 2025-04-14
refactor: :recycle: move `read_*` functions (one) into internal module

lwjohnst86:
## Description

Along with the other PRs, just tidying things up.

This PR needs a quick review.

## Checklist

- [x] Added or updated tests
- [x] Updated documentation
- [x] Ran `just run-all`


tests/core/test_read_json.py
martonvago, line 1: For such a general, multipurpose function, I think it's nice to have separate tests. Just because they don't rely on any context and they explore error cases that tests for calling functions might not care about. But it's not a hill I'll die on, especially not for this very simple function.

martonvago (dismissed):

tests/core/test_read_json.py
lwjohnst86: Since users won't be using this function and we will only use it in very specific ways that fit our purposes, we don't need to test edge cases etc.

I'm being firm on this because, from a refactoring point of view, its a major pain to have to figure out issues with internal tests failing when, really, they don't matter. Plus, code coverage and checks for unused code, putting these into tests can hide unused code and can falsely inflate code coverage.

PR #1276, opened 2025-04-14
refactor: :recycle: use `_map2()` in `read_resource_batches()`

lwjohnst86:
## Description

Related to my other PR, this adds another functional that takes two lists and a function. Again, this is inspired from the purrr R package.

This PR needs a quick review.

## Checklist

- [x] Added or updated tests
- [x] Updated documentation
- [x] Ran `just run-all`


src/seedcase_sprout/core/internals/functionals.py
```diff
@@ -0,0 +1,47 @@
+"""Mimicking the functional programming tools from R and the R package purrr."""
+
+from itertools import repeat
+
+
+def _map2(x: list, y: list, fn: callable) -> list:
```
martonvago, line 6:
```suggestion
def _map2(x: list, y: list, fn: Callable) -> list:
```

src/seedcase_sprout/core/internals/functionals.py
martonvago, line 1: For me, this is approaching the territory where it would be nice to have some sanity checks on it in the form of a test.

martonvago (changes requested):

src/seedcase_sprout/core/internals/functionals.py
lwjohnst86: As long as our public functions are doing as we expect and want, these internal functions are doing what we want them to do. If we continue adding functions to these to make functional programming a nicer experience in Python, than maybe we'd convert this into a package and formally test them there. But we really don't need to add maintenance burden that comes with more tests that are not needed for internal functions.

PR #1282, opened 2025-04-15
feat: :sparkles: finish up `PackagePath` and `resource_batch_files()`

martonvago:
## Description

This PR finishes up `PackagePath`, especially `resource_batch_files()`. I made it not throw an error if there are no batch files for the resource because the other methods don't throw either. I think returning `[]` when there are no batch files is in keeping with the spirit of the class. But we can always change it!

I also did some general tidying.

This is groundwork for #1210

This PR needs a quick review.

## Checklist

- [x] Added or updated tests
- [x] Updated documentation
- [x] Ran `just run-all`

### DIFF
--- a/src/seedcase_sprout/core/paths.py
+++ b/src/seedcase_sprout/core/paths.py
@@ -28,7 +28,7 @@ class PackagePath:
             directory.
 
     Returns:
-        The absolute path to the data package's file or folder.
+        A `PackagePath` object representing the structure of a data package.
 
     Examples:
         ```{python}
@@ -56,15 +56,15 @@ class PackagePath:
 
     def properties(self) -> Path:
         """Path to the `datapackage.json` file."""
-        return Path(self.path) / "datapackage.json"
+        return self.root() / "datapackage.json"
 
     def readme(self) -> Path:
         """Path to the `README.md` file."""
-        return Path(self.path) / "README.md"
+        return self.root() / "README.md"
 
     def resources(self) -> Path:
         """Path to the `resources/` folder."""
-        return Path(self.path) / "resources"
+        return self.root() / "resources"
 
     def resource(self, resource_name: str) -> Path:
         """Path to the specified `resources/<name>/` folder.
@@ -73,7 +73,7 @@ class PackagePath:
             resource_name: The name of the resource. Use `ResourceProperties.name` to
                 get the correct resource name.
         """
-        return Path(self.path) / "resources" / str(resource_name)
+        return self.resources() / str(resource_name)
 
     def resource_data(self, resource_name: str) -> Path:
         """Path to the specific resource's data file.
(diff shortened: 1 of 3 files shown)


src/seedcase_sprout/core/paths.py
```diff
@@ -28,7 +28,7 @@ class PackagePath:
             directory.
 
     Returns:
-        The absolute path to the data package's file or folder.
+        A `PackagePath` object representing the structure of a data package.
```
martonvago, line 31: Because the class doesn't return a `Path` I guess, but, well, a `PackagePath`.
But I don't feel strongly about this.

src/seedcase_sprout/core/paths.py
```diff
@@ -56,15 +56,15 @@ def root(self) -> Path:
 
     def properties(self) -> Path:
         """Path to the `datapackage.json` file."""
-        return Path(self.path) / "datapackage.json"
+        return self.root() / "datapackage.json"
```
martonvago, line 59: Just to avoid repeating ourselves

src/seedcase_sprout/core/paths.py
```diff
@@ -91,16 +91,14 @@ def resource_batch(self, resource_name: str) -> Path:
             resource_name: The name of the resource. Use `ResourceProperties.name` to
                 get the correct resource name.
         """
-        return Path(self.path) / "resources" / str(resource_name) / "batch"
+        return self.resource(resource_name) / "batch"
 
-    def resource_batch_files(self, resource_name: str) -> Path:
+    def resource_batch_files(self, resource_name: str) -> list[Path]:
```
martonvago, line 96: This returns all items in the folder, not only Parquet files. Could filter for Parquet files?

tests/conftest.py
martonvago, line 1: Quick shared fixture. This will be use _a lot_, here it's used once.

tests/core/test_paths.py
```diff
@@ -16,16 +14,48 @@ def test_package_path_outputs_an_absolute_path(tmp_path):
...
+    for file in ["test1", "test2"]:
+        batch_folder = package_path.resource_batch(file)
+        batch_folder.mkdir(parents=True)
+        (batch_folder / "file.txt").touch()
+
+    # Only the batch file for the given resource should be returned
+    assert package_path.resource_batch_files("test2") == [batch_folder / "file.txt"]
+
+
+def test_path_defaults_to_cwd_at_call_time(tmp_cwd):
```
martonvago, line 57: Note the `tmp_cwd`

src/seedcase_sprout/core/paths.py
```diff
@@ -91,16 +91,14 @@ def resource_batch(self, resource_name: str) -> Path:
...
         """Path to all the files in the specific resource's `batch/` folder.
```
martonvago, line 97:
```suggestion
        """Paths to all the files in the specific resource's `batch/` folder.
```

src/seedcase_sprout/core/paths.py
lwjohnst86: Yes, definitely filter for Parquet! :grin:

tests/conftest.py
lwjohnst86, line 1: Ahh, neat!

lwjohnst86 (approved):
Very nice! :tada: :star_struck:

lwjohnst86:
Oops, I merged after saying filter for Parquet!

martonvago:
No worries, will do! :robot:

PR #1283, opened 2025-04-16
feat: :sparkles: only return `.parquet` files in the batch folder

martonvago:
## Description

This PR makes `resource_batch_files()` return only `.parquet` files.

This PR needs a quick review.

## Checklist

- [x] Added or updated tests
- [x] Updated documentation
- [x] Ran `just run-all`

### DIFF
--- a/src/seedcase_sprout/core/paths.py
+++ b/src/seedcase_sprout/core/paths.py
@@ -100,5 +100,4 @@ class PackagePath:
             resource_name: The name of the resource. Use `ResourceProperties.name` to
                 get the correct resource name.
         """
-        batch_folder = self.resource_batch(resource_name)
-        return list(batch_folder.iterdir()) if batch_folder.is_dir() else []
+        return list(self.resource_batch(resource_name).glob("*.parquet"))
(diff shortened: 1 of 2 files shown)


src/seedcase_sprout/core/paths.py
```diff
@@ -100,5 +100,4 @@ def resource_batch_files(self, resource_name: str) -> list[Path]:
             resource_name: The name of the resource. Use `ResourceProperties.name` to
                 get the correct resource name.
         """
-        batch_folder = self.resource_batch(resource_name)
-        return list(batch_folder.iterdir()) if batch_folder.is_dir() else []
+        return list(self.resource_batch(resource_name).glob("*.parquet"))
```
martonvago, line 103: - This only checks the extension, it could still be a non-Parquet file with a Parquet extension, but I don't know if we care about that.
- I wasn't sure if we wanted to include extensions like `.parq` or `.pq`? We only use `.parquet`, so I only included that.

lwjohnst86: Let's be opinionated :wink: plus, Sprout is the one doing the managing/creation, so we decide what the extension is. `.parquet` is very clear :grin:

lwjohnst86 (approved):

PR #1264, opened 2025-04-11
refactor: :recycle: update `extract_resource_properties()` to take `pl.DataFrame`

signekb:
## Description

This PR updates `extract_resource_properties()` to take a `pl.DataFrame` instead of a `Path` to raw data. 

In doing this, I:

- added a mapping from `POLARS_TO_FRICTIONLESS` (which I’m not entirely sure about, so please take an in-depth look at that ☀️ ). 
- updated the resources guide. 
- updated the all types example to have a `my_list` field and updated the `dtypes` of `my_none` to `pl.Null` and `my_any` to `pl.Unknown`. 
- removed `frictionless py` as a dependency 🎉 

Closes #1203 

This PR needs an in-depth review.

## Checklist

- [X] Added or updated tests
- [X] Updated documentation
- [X] Ran `just run-all`

PR #1284, opened 2025-04-16
refactor: :recycle: make master Frictionless - Polars type map

martonvago:
## Description

This PR tidies up the Frictionless - Polars type maps:
- Removes the old `FRICTIONLESS_TO_POLARS`. This may be useful in e.g. shears, but no longer necessary in Sprout.
- Updates and renames `_FRICTIONLESS_TO_POLARS_TYPE_CHECK`. This is the master map. Updated with Signe's insights from #1264 .
- Adds a short function `_polars_to_frictionless()` to get the reverse mappings from the master map. This replaces `POLARS_TO_FRICTIONLESS` in #1264 .

## Checklist

- [x] Added or updated tests
- [x] Updated documentation
- [x] Ran `just run-all`

### DIFF
--- a/src/seedcase_sprout/core/map_data_types.py
+++ b/src/seedcase_sprout/core/map_data_types.py
@@ -94,6 +94,7 @@ def _get_allowed_polars_types(
         # added directly to the error message instead.
         return [pl.Array]
 
+    # TODO: Revise once `_map_keep()` has been implemented
     return [
         polars_type
         for polars_type, datapackage_types in _POLARS_TO_DATAPACKAGE.items()
--- a/src/seedcase_sprout/core/map_data_types.py
+++ b/src/seedcase_sprout/core/map_data_types.py
@@ -1,101 +1,114 @@
-from collections import OrderedDict
-from typing import Callable
-
 import polars as pl
 
 from seedcase_sprout.core.properties import FieldType
 
 """
-Mapping between Frictionless types and Polars types.
+Mapping between Polars types and Data Package types.
 
-Each Frictionless type maps to a function that checks if a given Polars type is an
-allowed representation of that Frictionless type. There are often multiple allowed
-Polars types for a single Frictionless type.
-
-To allow reverse lookups, i.e. mapping a Polars type to a single Frictionless type,
-entries are listed in order of precedence. The Polars type should be compared against
-Frictionless types in the order they are listed. The first entry with a passing check
-should determine the Frictionless type. Use `polars_to_frictionless()` to do the reverse
-lookup.
+Each Polars type can be described by any of the specified Data Package types.
+When extracting Data Package types from a data frame, each Polars type is mapped to
+the first Data Package type in the corresponding entry. 
 """
-_FRICTIONLESS_TO_POLARS: OrderedDict[FieldType, Callable[[pl.DataType], bool]] = (
-    OrderedDict(
-        {
-            # Primary types: reverse lookup always gives back Frictionless type
-            "number": lambda dtype: dtype.is_float() or dtype.is_decimal(),
-            "integer": lambda dtype: (
-                dtype.is_integer() or isinstance(dtype, pl.Duration)
-            ),
-            "string": lambda dtype: isinstance(
-                dtype, (pl.String, pl.Categorical, pl.Enum, pl.Binary)
-            ),
-            "date": lambda dtype: isinstance(dtype, pl.Date),
-            "datetime": lambda dtype: isinstance(dtype, pl.Datetime),
-            "time": lambda dtype: dtype == pl.Time,  # Polars stores Time as a class
-            "boolean": lambda dtype: isinstance(dtype, pl.Boolean),
-            # Secondary types: reverse lookup gives back Frictionless type only if
-            # primary type doesn't take precedence
-            "array": lambda dtype: isinstance(dtype, (pl.String, pl.Array, pl.List)),
-            "object": lambda dtype: isinstance(
-                dtype, (pl.String, pl.Struct, pl.Object)
-            ),
-            # Tertiary types: reverse lookup never gives back Frictionless type
-            "geopoint": lambda dtype: (
-                isinstance(dtype, pl.Array)
-                and dtype.size == 2
-                and dtype.inner.is_numeric()
-            ),
-            "geojson": lambda dtype: isinstance(
-                dtype, (pl.String, pl.Struct, pl.Object)
-            ),
-            "year": lambda dtype: dtype.is_integer(),
-            "yearmonth": lambda dtype: isinstance(dtype, pl.Date),
-            "duration": lambda dtype: isinstance(dtype, pl.String),
-            # Fallback type
-            "any": lambda _: True,
-        }
-    )
-)
-
-# Mapping from Frictionless types to descriptions of allowed Polars types
-_FRICTIONLESS_TO_POLARS_DESCRIPTION: dict[FieldType, str] = {
-    "number": "a float or decimal type",
-    "integer": f"an integer type or '{pl.Duration}'",
-    "string": (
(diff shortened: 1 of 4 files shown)


src/seedcase_sprout/core/map_data_types.py
martonvago, line 1: I moved all maps in here. They all follow the same structure / rely on each other, so this made sense for me, but let me know if they should be somewhere else.

src/seedcase_sprout/core/map_data_types.py
```diff
@@ -1,24 +1,101 @@
...
+    "duration": f"'{pl.String}'",
+    "any": "any type",
 }
+
+
+def _polars_to_frictionless(dtype: pl.DataType) -> FieldType:
+    """Map a Polars type to the corresponding Frictionless type.
+
+    Frictionless types are matched against the Polars type in the order set out in
+    `_FRICTIONLESS_TO_POLARS_TYPE_CHECK`. The first successful match is returned.
```
martonvago, line 86:
```suggestion
    `_FRICTIONLESS_TO_POLARS`. The first successful match is returned.
```

src/seedcase_sprout/core/map_data_types.py
```diff
@@ -1,24 +1,101 @@
...
+    Returns:
+        The corresponding Frictionless type.
+    """
+    return next(
+        (
+            frictionless_type
+            for frictionless_type, is_type in _FRICTIONLESS_TO_POLARS.items()
+            if is_type(dtype)
+        ),
+        "any",
```
martonvago, line 100: This is the default. Could technically be omitted because `_FRICTIONLESS_TO_POLARS` also has the same default, but maybe it's nice to have it here explicitly.

src/seedcase_sprout/core/map_data_types.py
```diff
@@ -1,24 +1,101 @@
...
+allowed representation of that Frictionless type. There are often multiple allowed
+Polars types for a single Frictionless type.
+
+To allow reverse lookups, i.e. mapping a Polars type to a single Frictionless type,
+entries are listed in order of precedence. The Polars type should be compared against
+Frictionless types in the order they are listed. The first entry with a passing check
+should determine the Frictionless type. Use `polars_to_frictionless()` to do the reverse
+lookup.
+"""
+_FRICTIONLESS_TO_POLARS: OrderedDict[FieldType, Callable[[pl.DataType], bool]] = (
```
martonvago, line 21: This is one response to the following requirements:
- Given a Frictionless type, we want to allow a range of Polars types (when checking data).
- Given a Polars type, we want to choose one Frictionless type to represent it (when extracting properties).
- We want the rules for both of these to come from the same place.

src/seedcase_sprout/core/map_data_types.py
```diff
@@ -1,24 +1,101 @@
...
+            "datetime": lambda dtype: isinstance(dtype, pl.Datetime),
+            "time": lambda dtype: dtype == pl.Time,  # Polars stores Time as a class
+            "boolean": lambda dtype: isinstance(dtype, pl.Boolean),
+            # Secondary types: reverse lookup gives back Frictionless type only if
+            # primary type doesn't take precedence
+            "array": lambda dtype: isinstance(dtype, (pl.String, pl.Array, pl.List)),
+            "object": lambda dtype: isinstance(
+                dtype, (pl.String, pl.Struct, pl.Object)
+            ),
+            # Tertiary types: reverse lookup never gives back Frictionless type
```
martonvago, line 42: E.g., given `pl.Int8`, I will always guess Frictionless type `integer` and not `year`.

tests/core/sprout_checks/test_check_data.py
```diff
@@ -41,44 +37,41 @@ def resource_properties() -> ResourceProperties:
...
-                [[3, 4], [5, 45], [12, -4]],
-                dtype=pl.Array(pl.Int16, 2),
-            ),
-        ]
+    data = pl.DataFrame(
+        {str(pl_type): pl.Series([], dtype=pl_type) for pl_type in pl_types}
     )
 
-    assert_frame_equal(check_data(data, resource_properties), data)
+    assert check_data(data, resource_properties) is data
```
martonvago, line 74: I need this to be able to work with `pl.Object`

src/seedcase_sprout/core/map_data_types.py
```diff
@@ -1,24 +1,101 @@
...
+    "geopoint": "an array of a numeric type with size 2",
+    "geojson": f"one of '{pl.String}', '{pl.Struct}', or '{pl.Object}'",
+    "year": "an integer type",
+    "yearmonth": f"'{pl.Date}'",
+    "duration": f"'{pl.String}'",
+    "any": "any type",
 }
+
+
+def _polars_to_frictionless(dtype: pl.DataType) -> FieldType:
```
martonvago, line 82: And I guess this will be tested when extracting properties in #1264

lwjohnst86 (changes requested):
Didn't review it all, but the comment I did make will mean some major changes, so I'll review it again after the changes.

src/seedcase_sprout/core/map_data_types.py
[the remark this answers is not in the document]
lwjohnst86: Hm, isn't one of the issues that we needed a two-way map? But here you're suggesting we don't? Hm, I'm a bit confused.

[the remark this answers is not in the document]
lwjohnst86: If we don't really need the two-way map, then what you suggests works very well :+1:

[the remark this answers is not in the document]
martonvago: It's because if we don't mind listing the polars types like this, then we can flip the original map direction, and this ends up working out nicely.

To the point where we're very close to being able to autogenerate the error messages as well. But for that we do need to go in the other direction. In any case, I'll put the next version up for review (based on the above) and you can look at it!

src/seedcase_sprout/core/map_data_types.py
```diff
@@ -2,23 +2,113 @@
...
-    "geojson": pl.String,
-    "any": pl.String,
+"""
+Mapping between Polars types and Data Package types.
+
+Each Polars type can be described by any of the specified Data Package types.
+When extracting Data Package types from a data frame, each Polars type is mapped to
+the first Data Package type in the corresponding entry.
+"""
+_POLARS_TO_DATAPACKAGE: dict[type[pl.DataType], list[FieldType]] = {
```
martonvago, line 12: This can live in `_get_allowed_datapackage_types()` if we don't care about having `_get_allowed_polars_types()`, which is used for generating the error messages.

src/seedcase_sprout/core/map_data_types.py
```diff
@@ -2,23 +2,113 @@
...
+) -> bool:
+    """Decide if the given Polars and Data Package types match.
+
+    Args:
+        polars_type: The Polars type to check.
+        datapackage_type: The Data Package type to check.
+
+    Returns:
+        Whether the given types match.
+
```
martonvago, line 75:
```suggestion
```

src/seedcase_sprout/core/map_data_types.py
```diff
@@ -2,23 +2,112 @@
...
+        The allowed Polars types.
+    """
+    datapackage_type = datapackage_type or "any"
+    if datapackage_type == "geopoint":
+        # Strictly speaking, this should be an Array of a numeric type with size 2, but
+        # it wouldn't be practical to list all combinations here. This information is
+        # added directly to the error message instead.
+        return [pl.Array]
+
+    return [
```
lwjohnst86, line 97:
```suggestion
    # TODO: Revise once `_map_keep()` has been implemented
    return [
```

I'm slowly updating some of the functionals, and this is next on my list. So just putting this TODO item to remind myself :relaxed:

lwjohnst86 (approved):
Very nice! :grin:

PR #1285, opened 2025-04-22
feat: :sparkles: add default path for path functions

### LINKED ISSUE #1108: Helper function to create a temporary directory?
Many of the examples and guide docs need to make a temporary directory for the data package. Would be nice to have a function to help with that, rather than write it out each time.

### LINKED ISSUE #1210: Have path argument default to the appropriate path relative to the cwd

### LINKED ISSUE #1250: [discussion]: Use `package_path` argument rather than `path`?

I've been thinking about this, especially in regard to #1213. What if instead of providing a path to the final point, we replace `path` argument in all functions with 
`package_path` that defaults to the working dir (or something similar). And that argument is giving to `PackagePath(package_path)`. Then, we internally use the paths from that class. I don't love that it would mean we couple strongly on location in the filesystem and couple strongly on the `PackagePath()` class and its paths.

But, it does go nicely with the idea @martonvago had about providing a default to the `path` arguments. And it would solve the issue in #1213 (since we use `properties.name` as an argument to `PackagePath().path_resource_batch(properties.name)`.

Thoughts?

martonvago:
## Description

This PR adds a default path to functions that take a path. So it's basically the same change in all the files.
It also updates some examples which wouldn't have built otherwise. Updating the rest of the examples will come in another PR.

Closes #1210 closes #1250 closes #1108

## Checklist

- [x] Added or updated tests
- [x] Updated documentation
- [x] Ran `just run-all`


src/seedcase_sprout/core/read_properties.py
```diff
@@ -2,18 +2,20 @@
 
 from seedcase_sprout.core.check_properties import check_properties
 from seedcase_sprout.core.internals import _check_is_file, _read_json
+from seedcase_sprout.core.paths import PackagePath
 from seedcase_sprout.core.properties import PackageProperties
 
 
-def read_properties(path: Path) -> PackageProperties:
+def read_properties(path: Path | None = None) -> PackageProperties:
```
lwjohnst86, line 9: Could the defaults not be `= PackagePath.properties()`?

src/seedcase_sprout/core/paths.py
```diff
@@ -32,17 +32,18 @@ class PackagePath:
...
+
         import seedcase_sprout.core as sp
 
+        # With default path
         sp.PackagePath().properties()
         sp.PackagePath().readme()
 
-        # Create a temporary directory for the example to show
-        # how to use the function with a different path
-        with tempfile.TemporaryDirectory() as temp_path:
```
martonvago, line 43: `temp_path` here was actually a string, but `PackagePath` now expects a `Path` strictly. 
But for this example I guess we don't even need a temp folder.

src/seedcase_sprout/core/read_properties.py
martonvago: No, because then we would get the cwd at definition time and it wouldn't always be the actual cwd of the caller

src/seedcase_sprout/core/create_resource_properties.py
```diff
@@ -20,13 +21,13 @@ def create_resource_properties(
...
-        path: The path to the resource `id` folder; use `PackagePath().resource()`
-            to provide the correct path or use the output of
-            `create_resource_structure()`.
         properties: The properties of the resource; must be given as a
             `ResourceProperties` object following the Data Package specification.
             See the `ResourceProperties` help documentation for details
             on what can or needs to be filled in.
+        path: The path to the resource `id` folder; use `PackagePath().resource()`
+            to provide the correct path or use the output of
+            `create_resource_structure()`.
```
lwjohnst86, line 30:
```suggestion
            to provide the correct path.
```

src/seedcase_sprout/core/read_resource_batches.py
```diff
@@ -76,13 +77,18 @@ def read_resource_batches(
...
             )
         ```
     """
-    _map(paths, _check_is_file)
+    paths = (
+        PackagePath().resource_batch_files(resource_properties.name)
+        if paths is None
+        else paths
+    )
```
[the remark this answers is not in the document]
lwjohnst86:
```suggestion
    paths = paths or PackagePath().resource_batch_files(resource_properties.name)
```

Or this like you used above?

tests/core/test_create_resource_structure.py
lwjohnst86, line 1: This file is deleted in #1295 so could you revert the changes so we don't have to deal with merge conflicts?

lwjohnst86 (changes requested):
Only a few comments, otherwise :tada: :grin:

src/seedcase_sprout/core/read_resource_batches.py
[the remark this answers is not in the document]
martonvago: Here you need the `is None` check because an empty `paths` (i.e. `[]`) is falsy. So we can have the first suggestion :+1:

lwjohnst86 (approved):

PR #1292, opened 2025-04-22
fix: :lock: fixed bandit flagged issues, moved dataclass gen into `tools/`

lwjohnst86:
## Description

Bandit flagged some things, which this PR fixes.

- Should have an escaping for reading in jinja files.
- Requests should have timeout.
- Isn't secure passing variables to subprocess.

For the last one, which was for generating the dataclasses from the JSON schema, I just moved it out of the package into a `tools/` directory. I also removed the use of subprocess and called `generate()` directly.

Unfortunately, even the original generator file didn't work for me after moving to tools, because it couldn't find the module tools, even with the `__init__.py` file. I decided to leave it as is since if/when they update the data package version, we probably would have to make some changes anyway.

This PR needs a quick review.

## Checklist

- [x] Added or updated tests
- [x] Updated documentation
- [x] Ran `just run-all`


martonvago (approved):
Yeah, I don't think what's going on in `generate_dataclasses` matters very much because it's just a helper script that is not really a part of the system :+1: :+1:

Issue #1144, opened 2025-03-18
[discussion]: Resource id's use resource's `name` field as an ID for folders?


lwjohnst86:
### What would you like to discuss?

Rather than use and create new number IDs for resources, we could use the resource's name to make and refer to the folder. The name needs to reflect the resource and there can't be duplicates. 

- Advantage: Don't have to create a new resource ID, plus it is clearer what the  resource folder contains than `resource/1/`.
- Disadvantage: If the `name` field is renamed, the resource folder also has to be renamed.

signekb:
From a user perspective, I think it would be nice that the resource’s folder name is the actual name of the resource as well. But I do recognise the disadvantage. 
Would that require that if a user renames the resource in the properties, we would have a function that renames the folder?

martonvago:
This would be a lot easier to code with for sure! 👍 👍 

True about the renaming of the folder. Maybe we could have a separate function for that to emphasise that it's different from changing other attributes. But maybe that's overkill.

martonvago:
I've had a look at this and there are 2 options:

1. Just change the function `create_relative_resource_data_path()` to use the resource name instead of the ID.
  ➕ Very easy to do.
  ➖ But then we will have to take care to set the path at the right point in the flow and warn/prevent users from changing the path to something else.
2. Make path a computed property that is autogenerated from the name. 
  ➕ Same `my_resource.path` interface, but read-only -> cannot set the path to something else
  ➕ No need to set the path at any point, it is always generated from the name.
  ➖ We would need to rewrite `compact_dict` to include computed properties as well. This is not difficult, but it's a more intrusive change than option 1.

@lwjohnst86 , do you have a preference?

lwjohnst86:
Hm, I think for now, go with the easier option. I have a feeling we'll need to refactor some of how the `create_relative_resource_data_path()` works. So getting it fixed this way works fine for the short term 😁 👍

PR #1295, opened 2025-04-23
refactor: :recycle: action of `create_resource_structure()` into `write_resource_batch()`

### LINKED ISSUE #1112: Consider the flow for making the `resources` dir
Right now, `create_resource_structure()` only creates new `id` resource folders, including the raw inside. So we need to manually create the resource folder it is in.

### LINKED ISSUE #1171: Update/delete/integrate `create_resource_structure()`
After the updates to the design, this doesn't match the flow anymore. Plus, now that we will go with using the `name` for the resource ID in #1144, we can probably move a lot of this functionality out into writing (where it will create the folders when needed).

lwjohnst86:
## Description

This removes the `create_resource_structure()` and moves its functionality into `write_resource_batch()`. This lead to deleting a lot of unneeded files and tests.

Unfortunately, tests are failing because it depends on @martonvago's work with #1144

Closes #1112, closes #1171

## Checklist

- [x] Added or updated tests
- [x] Updated documentation
- [x] Ran `just run-all`


src/seedcase_sprout/core/examples.py
```diff
@@ -277,7 +276,7 @@ def __enter__(self) -> PackagePath:
             # Create resource folders
             # TODO: update after resource creation is refactored
             package_path.resources().mkdir(exist_ok=True)
```
martonvago, line 278: So I was thinking that we could remove everything in the if-clause after this line (I can't select all of it)? Nothing is relying on it now, and the tests will pass. If I need something like this when refactoring the resource path, I can add it back later.

tests/core/test_examples.py
```diff
@@ -57,10 +57,7 @@ def test_creates_package_with_resources():
         properties = read_properties(package_path.properties())
         assert check_properties(properties)
         # Test resources
-        assert package_path.resources().is_dir()
```
martonvago, line 60: I think either leave this assert or delete the line making this folder in the code

martonvago (changes requested):
Nice!! :tada:

src/seedcase_sprout/core/examples.py
lwjohnst86: There is actually a test for it. I will put `x` in the front that we can come back to it later, how does that sound?

lwjohnst86: Ah, it's because `check_properties()` is checking if the path has the ID in it. That is in the `check_datapackage` sub-package. I'm thinking that shouldn't be in there, but in the Sprout checks.

docs/guide/resources.qmd
```diff
@@ -159,11 +167,10 @@ resource_properties.description = "This data resource contains data about..."
...
 Now that you have the properties for the resource, you can create the
-resource itself within your existing data package. As a data package can
-contain multiple resources, each resource is stored in a separate
-folder. So in this step you will make a folder for the new resource
-within the data package and save the resource properties to the
-`datapackage.json` file of the package.
+properties for the resource within your existing data package. A data
+package can contain multiple resources, so their `name` property
+must be unique. This `name` property is what will be used later
+to create the folder structure. for that resource.
```
martonvago, line 173:
```suggestion
to create the folder structure for that resource.
```

martonvago (changes requested):
Nice!! Just a single typo :star:

signekb (changes requested):
👍 Just a little misplaced `.`, otherwise approved.

docs/guide/resources.qmd
```diff
@@ -155,11 +163,10 @@ resource_properties.description = "This data resource contains data about..."
...
```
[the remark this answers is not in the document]
signekb: Duplicate of Marton's comment!

signekb (approved):

PR #1298, opened 2025-04-23
refactor: :recycle: moved functions into internals/location of use

lwjohnst86:
## Description

This moves functions either into `internals/` if they are used in multiple scripts, or moved into the script that are used in.

~Also, running `just run-all` I picked up an error with the `PackagePath()`. I needed to force the path to `Path()` if a string is given (like in `temp_path`).~

This PR needs a quick review.

## Checklist

- [x] Added or updated tests
- [x] Updated documentation
- [x] Ran `just run-all`


src/seedcase_sprout/core/paths.py
```diff
@@ -49,6 +48,7 @@ class PackagePath:
     def __init__(self, path: Path | None = None) -> Path:
         """Set the base path."""
         self.path = path or Path.cwd()
+        self.path = Path(self.path)
```
martonvago, line 51: Oh I've also seen this and changed the example in #1285 .
I think either
- accept a string but then let's put that in the type hints
- accept only a `Path` and never use this with a string input (now that way on main)

martonvago (changes requested):
Nice, just one comment :black_bird:

PR #1299, opened 2025-04-23
build: :hammer: add auto-spell checker

### LINKED ISSUE #1221: Include auto-spell checker
I've set up something similar in rostools, where the dictionary is in the repo rather than in another location. Unfortunate thing about the cSpell cli is that we need to install it from `npx`. Include this in both justfile and in action. (maybe link them together somehow?)

lwjohnst86:
## Description

Found a super nifty Rust package (with Python package) that does spell checking.

Removed use of cSpell as well, since it isn't needed anymore.

This will be included in the GitHub action shortly.

Closes #1221

This PR needs a quick review.

## Checklist

- [x] Added or updated tests
- [x] Updated documentation
- [x] Ran `just run-all`

### DIFF
--- a/.typos.toml
+++ b/.typos.toml
@@ -1,0 +1,9 @@
+[files]
+extend-exclude = [
+  "*.json",
+  "*.css",
+  ".quarto/*",
+  "_site/*",
+  "_extensions/*",
+  ".coverage-report/*"
+]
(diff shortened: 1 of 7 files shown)


martonvago (approved):
Looks good to me! :heavy_check_mark: :heavy_check_mark: :heavy_check_mark:

PR #1302, opened 2025-04-24
feat: :sparkles: `write_resource_data()`

### LINKED ISSUE #806: Write and test `write_resource_parquet` function

martonvago:
## Description

This PR adds `write_resource_data()`.

Closes #806 

## Checklist

- [x] Added or updated tests
- [x] Updated documentation
- [x] Ran `just run-all`


src/seedcase_sprout/core/write_resource_data.py
```diff
@@ -0,0 +1,56 @@
...
+            # TODO: delete after data path is refactored to use name
+            resource_properties.name = "1"
+            # Add and join batch files
+            sp.write_resource_batch(sp.example_data(), resource_properties)
+            batches = sp.read_resource_batches(resource_properties)
+            data = sp.join_resource_batches(batches, resource_properties)
+            # Write resource data file
+            sp.write_resource_data(data, resource_properties)
+    """
+    check_data(data, resource_properties)
```
martonvago, line 52: This also checks the resource properties

lwjohnst86 (approved):
Niicccceeee :tada:

PR #1303, opened 2025-04-28
refactor :recycle: use `name` in path

### LINKED ISSUE #1144: [discussion]: Resource id's use resource's `name` field as an ID for folders?

martonvago:
## Description

This PR makes `resource.path` use `resource.name` instead of an ID.

Closes #1144 

## Checklist

- [x] Added or updated tests
- [x] Updated documentation
- [x] Ran `just run-all`


src/seedcase_sprout/core/internals/_create_relative_resource_data_path.py
martonvago, line 1: This function could live in `PackagePath` maybe as a class method (as it doesn't need an instance). It returns a string, so it's not exactly like the other paths methods, so I wasn't sure.

src/seedcase_sprout/core/internals/_create_relative_resource_data_path.py
```diff
@@ -0,0 +1,25 @@
...
+
+from seedcase_sprout.core.internals._is_resource_name_incorrect import (
+    _is_resource_name_incorrect,
+)
+
+
+def _create_relative_resource_data_path(resource_name: Any) -> str | None:
+    """Creates a stringified relative path to the resource data file based on the name.
+
+    The path is created only if the resource name is correct.
```
martonvago, line 12: I thought the cleanest thing to do when the name has something wrong with it was to not create a path and leave it as `None`. The checks will flag the bad name, and we don't have to worry about the various ways in which a bad name can break the path.

src/seedcase_sprout/core/properties.py
```diff
@@ -510,6 +513,10 @@ class ResourceProperties(Properties):
     hash: str | None = None
     schema: TableSchemaProperties | None = None
 
+    def __post_init__(self):
+        """Generates the path from the resource name after object creation."""
+        self.path = _create_relative_resource_data_path(self.name)
+
```
martonvago, line 519: This is a halfway-house solution between (1) calling `_create_relative_resource_data_path` manually every time a resource properties is created and (2) making `path` a computed property.
- Advantage over (1): the path is set automatically based on the name when a resource properties is created (provided the name is correct). Doesn't allow setting a bad path at object creation.
- Disadvantage over (2): the path can still be changed after object creation to any value. If the name is updated, the path is not updated automatically.

Even with this solution we could iron out the disadvantages (by overriding `__setattr__`), but at that point we really should just be doing (2).

src/seedcase_sprout/core/write_resource_properties.py
```diff
@@ -57,11 +57,17 @@ def write_resource_properties(
     check_resource_properties(resource_properties)
 
     package_properties = _read_json(path)
-    check_package_properties(PackageProperties().from_dict(package_properties))
+    check_properties(PackageProperties().from_dict(package_properties))
```
martonvago, line 60: I changed this because I want to check existing resource properties too: if one of those has no name / a bad name, I cannot compare it with the input resource properties.

(This function will probably change when the update flow is worked out, so it doesn't matter that much.)

src/seedcase_sprout/core/write_resource_properties.py
```diff
@@ -57,11 +57,17 @@ def write_resource_properties(
...
-    resource_id = get_resource_id_from_properties(resource_properties)
-    current_resource = get_resource_properties(package_properties, resource_id)
+    current_resource = next(
+        (
+            resource
+            for resource in package_properties.get("resources", [])
+            if resource["name"] == resource_properties["name"]
+        ),
+        None,
+    )
```
martonvago, line 70: Gets the first matching existing resource or `None`

tests/core/sprout_checks/test_check_properties.py
```diff
@@ -256,33 +259,27 @@ def test_error_blank_resource_properties(properties, name, type):
...
 
 
-@mark.parametrize(
-    "item,value,validator",
-    [
-        ("name", "a name with spaces", "pattern"),
-    ],
-)
-def test_error_incorrect_resource_property_values(properties, item, value, validator):
+def test_error_incorrect_resource_property_values(properties):
```
martonvago, line 262: Just moving this to a simpler error case

tests/core/sprout_checks/test_check_resource_path_format.py
martonvago, line 1: Used to be `test_check_id_in_resource_path.py`

src/seedcase_sprout/core/check_datapackage/is_resource_name_correct.py
martonvago, line 1: I thought it made sense to put this here

martonvago, line 1: I thought it made sense to put this here

tests/core/sprout_checks/test_check_resource_path_format.py
martonvago, line 1: Used to be `test_check_id_in_resource_path.py`

src/seedcase_sprout/core/sprout_checks/check_resource_path_format.py
martonvago, line 1: Used to be `check_id_in_resource_path.py`

lwjohnst86 (changes requested):
Very nice! Just some small comments/suggestions

src/seedcase_sprout/core/properties.py
```diff
@@ -16,6 +16,9 @@
 from dacite import from_dict
 
 from seedcase_sprout.core.internals import _get_iso_timestamp
+from seedcase_sprout.core.internals._create_relative_resource_data_path import (
+    _create_relative_resource_data_path,
```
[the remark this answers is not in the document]
martonvago: It's also in `check_resource_path_format`

tests/core/sprout_checks/test_get_sprout_resource_errors.py
martonvago, line 1: Kept only one test from the deleted test file

lwjohnst86 (changes requested):
Nice, only a few comments

src/seedcase_sprout/core/check_datapackage/is_resource_name_correct.py
martonvago: I don't mind either way! Should I move it?

lwjohnst86 (approved):
I moved the file into the sprout checks :)

PR #1304, opened 2025-04-29
refactor: :recycle: have a default method only for `PackageProperties`

### LINKED ISSUE #1209: Remove `default` methods in `properties.py`
These methods made sense in the first iteration of working with `*Properties`, but now they are quite misleading. For the most part, they return properties that fail the checks and most fields don't have sensible defaults anyway.

There is a case for having default `PackageProperties` and maybe also `ResourceProperties`, but not really for the rest? 
And even for `PackageProperties`, `PackageProperties.default()`, is a pretty clunky way of setting defaults for only 3 fields. Some discussion on this [here](https://github.com/seedcase-project/seedcase-sprout/pull/1153#discussion_r2003604421).

martonvago:
## Description

This PR removes the default method from all properties classes except `PackageProperties`.

In `PackageProperties`, I changed the default method to be in line with how it's used in the guide. Based on an old [discussion](https://github.com/seedcase-project/seedcase-sprout/pull/1153#discussion_r2003604421).
Another approach would be to populate these fields automatically when creating the object in the usual way, but I'm not sure if we want that.

Closes #1209

## Checklist

- [x] Added or updated tests
- [x] Updated documentation
- [x] Ran `just run-all`

### DIFF
--- a/src/seedcase_sprout/core/properties.py
+++ b/src/seedcase_sprout/core/properties.py
@@ -478,8 +478,8 @@ class PackageProperties(Properties):
     ) -> Self:
         """Creates a `PackageProperties` dataclass with default values.
 
-        Default values can be overridden and unset values can be set using
-        keyword arguments.
+        Default values (`id`, `version`, and `created`) can be overridden and unset
+        values can be set using keyword arguments.
 
         Args:
             name (str | None): A simple name or identifier to be used for this package.
--- a/src/seedcase_sprout/core/properties.py
+++ b/src/seedcase_sprout/core/properties.py
@@ -8,7 +8,7 @@ are intended to help users with the correct structure and content of the propert
 # `generate_properties/generated_properties.py` file. Update the auto-generated
 # properties file to add more dataclasses and move them into this file.
 
-from abc import ABC, abstractmethod
+from abc import ABC
 from dataclasses import asdict, dataclass
 from typing import Any, Literal, Self
 from uuid import uuid4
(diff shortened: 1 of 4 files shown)


src/seedcase_sprout/core/properties.py
```diff
@@ -600,14 +459,74 @@ class PackageProperties(Properties):
...
+        description: str | None = None,
+        homepage: str | None = None,
+        version: str | None = None,
+        created: str | None = None,
+        contributors: list[ContributorProperties] | None = None,
+        keywords: list[str] | None = None,
+        image: str | None = None,
+        licenses: list[LicenseProperties] | None = None,
+        resources: list[ResourceProperties] | None = None,
+        sources: list[SourceProperties] | None = None,
```
martonvago, line 477: I'm listing all of them instead of doing `**properties` because it's nice to have the autocompletion when using the method.

lwjohnst86: Love that :star:

lwjohnst86 (dismissed):
Very nice :star: I guess we could also do them for the other properties like resource properties?

martonvago:
@lwjohnst86 Yeah, for resource properties there are maybe 2 fields that have sensible defaults? And there's `fields_match` on `TableSchemaProperties`. I'm not sure about the rest.

signekb (changes requested):
Very nice 🌟 
Just added a suggestion to the docstring, otherwise approved ✅

signekb (approved):

PR #1305, opened 2025-04-29
fix: :bug: do not escape non-ASCII characters when writing JSON

martonvago:
## Description

This PR fixes the encoding of special characters. Turns out `json.dumps` automatically escapes non-ASCII characters, which we don't want, we want to have them as UTF-8 (the default). This was pointed out by Kris a long time ago.

This PR needs a quick review.

## Checklist

- [x] Added or updated tests
- [x] Updated documentation
- [x] Ran `just run-all`


signekb (approved):
Yay! Super nice 🌟

lwjohnst86 (approved):
oooooo!!

PR #1308, opened 2025-05-01
fix: :bug: raise a more informative error when the data list is empty

martonvago:
## Description

This PR makes `join_resource_batches` raise an error explicitly and with an informative error message when the input data list is empty. Otherwise `pl.concat` will throw an error saying "cannot concat empty list".

This happens realistically when the batch files are loaded using `batches = sp.read_resource_batches(resource_properties)`, but there aren't any batch files. I just happened to run into this when running a version of the resource guide.

This PR needs a quick review.

## Checklist

- [x] Added or updated tests
- [x] Updated documentation
- [x] Ran `just run-all`

### DIFF
--- a/src/seedcase_sprout/core/join_resource_batches.py
+++ b/src/seedcase_sprout/core/join_resource_batches.py
@@ -40,6 +40,7 @@ def join_resource_batches(
             observational units removed.
 
     Raises:
+        ValueError: If an empty `data_list` is provided.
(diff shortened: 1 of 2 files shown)


lwjohnst86 (approved):

signekb (approved):

lwjohnst86:
There's a build error but it's because vulture needs to be installed here. I'll do that in another PR

PR #1309, opened 2025-05-01
feat: :sparkles: add custom package name to `ExamplePackage`

martonvago:
## Description

This PR adds a `package_name` parameter in `ExamplePackage` for specifying a custom package name.
The package name is set as the name of the root folder of the package and as the name of the package properties.

I'll use `ExamplePackage` in the resource guide in another PR.

This PR needs a quick review.

## Checklist

- [x] Added or updated tests
- [x] Updated documentation
- [x] Ran `just run-all`


src/seedcase_sprout/core/examples.py
```diff
@@ -241,18 +241,27 @@ class ExamplePackage(AbstractContextManager):
...
+
+        with sp.ExamplePackage(package_name="my-package"):
+            properties = sp.read_properties()
         ```
     """
 
-    def __init__(self, with_resources: bool = True):
+    def __init__(
+        self,
+        package_name: str = example_package_properties().name,
```
martonvago, line 254: Not super elegant calling `example_package_properties` both here and in `__enter__`, but maybe it doesn't matter that much.

lwjohnst86 (dismissed):

lwjohnst86 (approved):

PR #1322, opened 2025-05-06
fix: :bug: resolve temp paths so tests pass on MacOS as well

signekb:
## Description

Two tests were failing for me locally and it seems to be bc of how MacOS uses `/var` as a symlink to `/private/var` (thanks @martonvago 🌟 ). 
So, these two tests were failing bc the paths are compared directly and even though they refer to the same location, the paths aren’t identical. I solved this by resolving the paths with `.resolve()`. 

A different solution could be to change how we assert the paths. Like `assert Path.cwd().samefile(package_path.path)`. 

Thoughts?

## Checklist

- [X] Added or updated tests
- [X] Ran `just run-all`

### DIFF
--- a/tests/core/test_examples.py
+++ b/tests/core/test_examples.py
@@ -85,7 +85,9 @@ def test_changes_cwd_to_package_root_in_package_context():
     After exiting the context, the cwd should be reset to the original location."""
     original_cwd = Path.cwd()
     with ExamplePackage() as package_path:
-        assert Path.cwd() == package_path.path
+        # Resolve paths to ensure symlinks are compared correctly on MacOS as well
+        # (i.e., `/var` -> `/private/var` on MacOS)
+        assert Path.cwd().resolve() == package_path.path.resolve()
 
     assert Path.cwd() == original_cwd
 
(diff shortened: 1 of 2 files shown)


martonvago (approved):
Nice!! :+1: :+1: 

And yet another solution would be to put this resolve call directly into `ExamplePackage`: maybe l.261 `package_path = PackagePath(Path(self.temp_dir.name).resolve())`? (I haven't tested this.)

signekb:
Ah! I just tried it out, but only adding `.resolve()` to `ExamplePackage` makes the test `test_manages_temp_folder_correctly()` fail bc now `package_path` starts with `/private/` while `temp_root` does not 🤷‍♀️ 

https://github.com/seedcase-project/seedcase-sprout/blob/25e03eca5ce3256a61fe39c90b06b2746f02dd8c/tests/core/test_examples.py#L103-L110

Could change it there as well? Since this is only a test problem (related to how we assert the paths) maybe it’s fine to make changes to the tests only?

martonvago:
Ah okay! Yeah, I agree :+1: :+1:

lwjohnst86 (approved):

PR #1325, opened 2025-05-13
build: :hammer: build PDF locally

### LINKED ISSUE #1314: Build Sprout docs into PDF for eventual archiving to Zenodo

martonvago:
## Description

This PR adds a Quarto profile for rendering the website as a PDF.

I'm sure this can be improved a lot, but I don't know Quarto that well, so I'm putting this up for review now instead of sitting on it longer.

I've also included some tweaks to improve the output, but it's not perfect yet.
Outstanding issues:
- The `glossary` extension doesn't support PDF. Replacing the shortcode in the text as a post-processing step (e.g. `{{< glossary "data package">}}` >> `data package`, so that at least it reads normally) is not as trivial as it seems. It's still possible, but I didn't want to spend more time on it if we're moving away from this extension anyway. 
- Very long code output goes off the page. I mitigated this somewhat by auto-wrapping code output, but these two cases remain:
  1. super long words (like an absolute path) are not wrapped
  2.  `pprint`ed objects have lots of whitespace that pushes them off the page regardless of wrapping. 
Displaying big, nested properties nicely is a problem for the HTML website as well, so maybe we need a common solution? 
- We could have a nice book cover / first page. There are Quarto extensions that claim to help with that, but I haven't tried them.

This is what it looks like now:
[seedcase-sprout.pdf](https://github.com/user-attachments/files/20184031/seedcase-sprout.pdf)

Closes #1314 

## Checklist

- [x] Added or updated tests
- [x] Updated documentation
- [x] Ran `just run-all`

### DIFF
--- a/_quarto-pdf.yml
+++ b/_quarto-pdf.yml
@@ -1,0 +1,48 @@
+project:
+  type: book
+
+format:
+  pdf:
+    documentclass: scrbook
+    papersize: a4
+    geometry:
+      - top=3cm
+      - bottom=3cm
+    
+    number-sections: true
+    number-depth: 2
+    toc: true
+    toc-depth: 2
+    colorlinks: true 
+
+    shortcodes: 
+      - _extensions/seedcase-project/seedcase-theme/_extensions/quarto-ext/fontawesome/fontawesome.lua
+
+book:
+  output-file: seedcase-sprout
+  title: "Seedcase Sprout Documentation"
+  author: "Seedcase Project"
+  chapters:
+    - index.qmd
+    - part: "Guide"
+      chapters:
+        # - docs/guide/index.qmd
+        - docs/guide/installation.qmd
+        - docs/guide/packages.qmd
+        - docs/guide/resources.qmd
+        - docs/guide/checks.qmd
+    - part: "Design - Architecture"
+      chapters:
+        - docs/design/architecture/index.qmd
+        - docs/design/architecture/c4-models.qmd
+        - docs/design/architecture/input-data.qmd
+        - docs/design/architecture/naming.qmd
+        - docs/design/architecture/user-personas.qmd
+    - part: "Design - Interface"
+      chapters:
+        - docs/design/interface/index.qmd
+        - docs/design/interface/flows.qmd
+        - docs/design/interface/functions.qmd
+        - docs/design/interface/inputs.qmd
+        - docs/design/interface/outputs.qmd
+
--- a/_quarto-pdf.yml
+++ b/_quarto-pdf.yml
@@ -1,6 +1,9 @@
 project:
   type: book
 
+filters:
+  - filters/pdf-internal-link.lua
+
 format:
   pdf:
     documentclass: scrbook
--- a/_quarto-pdf.yml
+++ b/_quarto-pdf.yml
@@ -46,7 +46,7 @@ format:
 book:
   output-file: seedcase-sprout
   title: Seedcase Sprout Documentation
-  author: "Seedcase Project"
+  author: Seedcase Project
   chapters:
     - index.qmd
     - part: Guide
--- a/_quarto-pdf.yml
+++ b/_quarto-pdf.yml
@@ -22,10 +22,12 @@ format:
       - openany
 
     include-in-header: 
-      # Wrap text in code blocks
       text: |
+        % Wrap text in code blocks
         \usepackage{fvextra}
         \DefineVerbatimEnvironment{Highlighting}{Verbatim}{breaklines,commandchars=\\\{\}}
+        % Keep code blocks where they appear in the text
+        \AtBeginDocument{\floatplacement{codelisting}{H}}
    
     papersize: a4
     geometry:
--- a/_quarto-pdf.yml
+++ b/_quarto-pdf.yml
@@ -17,18 +17,18 @@ include-before-body:
 format:
   pdf:
     documentclass: scrbook
-    classoption: 
+    classoption:
       # Chapters can start on any page (no blank pages)
       - openany
 
-    include-in-header: 
+    include-in-header:
       text: |
         % Wrap text in code blocks
         \usepackage{fvextra}
         \DefineVerbatimEnvironment{Highlighting}{Verbatim}{breaklines,commandchars=\\\{\}}
         % Keep code blocks where they appear in the text
         \AtBeginDocument{\floatplacement{codelisting}{H}}
-   
+
     papersize: a4
     geometry:
       - top=3cm
@@ -38,9 +38,9 @@ format:
     number-depth: 2
     toc: true
     toc-depth: 2
-    colorlinks: true 
+    colorlinks: true
 
-    shortcodes: 
+    shortcodes:
       - _extensions/seedcase-project/seedcase-theme/_extensions/quarto-ext/fontawesome/fontawesome.lua
 
 book:
@@ -70,4 +70,3 @@ book:
         - docs/design/interface/functions.qmd
         - docs/design/interface/inputs.qmd
         - docs/design/interface/outputs.qmd
-
--- a/_quarto-pdf.yml
+++ b/_quarto-pdf.yml
@@ -7,6 +7,9 @@ filters:
 format:
   pdf:
     documentclass: scrbook
+    classoption: 
+      - openany
+    # keep-tex: true
     papersize: a4
     geometry:
       - top=3cm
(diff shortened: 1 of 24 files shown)


_quarto-pdf.yml
```diff
@@ -0,0 +1,72 @@
...
+      # Chapters can start on any page (no blank pages)
+      - openany
+
+    include-in-header:
+      text: |
+        % Wrap text in code blocks
+        \usepackage{fvextra}
+        \DefineVerbatimEnvironment{Highlighting}{Verbatim}{breaklines,commandchars=\\\{\}}
+        % Keep code blocks where they appear in the text
+        \AtBeginDocument{\floatplacement{codelisting}{H}}
```
martonvago, line 30: From https://github.com/quarto-dev/quarto-cli/issues/5142#issuecomment-2226281237

_quarto-pdf.yml
```diff
@@ -0,0 +1,72 @@
...
+      - bottom=3cm
+
+    number-sections: true
+    number-depth: 2
+    toc: true
+    toc-depth: 2
+    colorlinks: true
+
+    shortcodes:
+      - _extensions/seedcase-project/seedcase-theme/_extensions/quarto-ext/fontawesome/fontawesome.lua
```
martonvago, line 44: I couldn't make the extensions of the theme load by referencing the theme (and still keeping the `book` project type). Let me know if you know how to do it!

_quarto-pdf.yml
```diff
@@ -0,0 +1,72 @@
+project:
+  type: book
```
martonvago, line 2: To create a single PDF, not one for each document

_quarto-pdf.yml
```diff
@@ -0,0 +1,72 @@
...
+    toc-depth: 2
+    colorlinks: true
+
+    shortcodes:
+      - _extensions/seedcase-project/seedcase-theme/_extensions/quarto-ext/fontawesome/fontawesome.lua
+
+book:
+  output-file: seedcase-sprout
+  title: Seedcase Sprout Documentation
+  author: "Seedcase Project"
```
martonvago, line 49:
```suggestion
  author: Seedcase Project
```

_quarto-pdf.yml
```diff
@@ -0,0 +1,72 @@
...
+    toc-depth: 2
+    colorlinks: true
+
+    shortcodes:
+      - _extensions/seedcase-project/seedcase-theme/_extensions/quarto-ext/fontawesome/fontawesome.lua
+
+book:
+  output-file: seedcase-sprout
+  title: Seedcase Sprout Documentation
+  author: Seedcase Project
```
martonvago, line 49: Feel free to change or add more details

_quarto-pdf.yml
```diff
@@ -0,0 +1,72 @@
...
+  chapters:
+    - index.qmd
+    - part: Guide
+      chapters:
+        # - docs/guide/index.qmd
```
martonvago, line 54: This is empty, so I wasn't sure whether to include it just in case

_quarto-pdf.yml
```diff
@@ -0,0 +1,72 @@
...
+        - docs/guide/installation.qmd
+        - docs/guide/packages.qmd
+        - docs/guide/resources.qmd
+        - docs/guide/checks.qmd
+    - part: "Design: Architecture"
```
martonvago, line 59: You cannot have more levels of nesting (no sub-chapters) -- is this okay this way?

martonvago: Also, wildcards don't work here, so everything has to be listed

_variables.yml
```diff
@@ -1,2 +1,2 @@
-done: '[{{< fa circle-check size="l" title="Completed" >}}]{style="color: #47DC76"}'
+done: '[{{< fa check-circle size="l" title="Completed" >}}]{style="color: #47DC76"}'
```
martonvago, line 1: I thiiiink this is the name of this icon in an earlier version. In any case, this works in both PDF and HTML.

docs/design/architecture/c4-models.qmd
```diff
@@ -20,23 +20,24 @@ for more information.
 
 The Context diagram shows the users and any external systems that
 interact with Sprout. This includes the user roles described in the
-[User Personas](user-personas.qmd) page.
+[User Personas](user-personas.qmd#user-types) page.
```
martonvago, line 23: This is to make internal links work in the PDF. More details in `pdf-internal-link.lua`

docs/design/architecture/c4-models.qmd
```diff
@@ -20,23 +20,24 @@ for more information.
...
 
 ::: callout-caution
 For some reason, these diagrams don't display well on some browsers like
 Firefox. To see them, try using a different browser like Chrome or Edge.
 :::
 
-::: column-page
```
martonvago, line 30: This didn't work well in PDF, and now that I've wrapped the text in the diagrams, the HTML looks fine without it.

docs/design/architecture/c4-models.qmd
```diff
@@ -20,23 +20,24 @@ for more information.
...
 ```{mermaid}
+%%| label: fig-c4-context
```
martonvago, line 31: Figures need a label to be centred in PDF (this took me half a day to figure out :see_no_evil: )

docs/design/architecture/c4-models.qmd
```diff
@@ -20,23 +20,24 @@ for more information.
...
-    user_admin(["Administrator<br>[person]<br><br>Manages data and metadata from a project for an owner. Uses Sprout Core for more control."])
-    user_swe(["Software engineer<br>[person]<br><br>Creates and develops data infrastructure. Uses Sprout Core to build extensions and customizations."])
-    user_de(["Data engineer<br>[person]<br><br>Develops data pipelines and infrastructures. Uses Sprout Core to build and customize the pipelines."])
-    sprout["Seedcase Sprout<br>[system]<br><br>Allows users fine-grained control over building, managing, and structuring data in a way that makes it easier to share with collaborators and to analyze. Structures data following the Frictionless Data Package Standard."]
-    output[("Filesystem (local or server)<br>[system]<br><br>Structured data from Sprout that is stored in a filesystem for later use. This can be on a server or on a local, personal computer, depending on the privacy of the data.")]
+    user_owner(["Owner/creator<br>[person]<br><br>Creates and owns a<br>project that generates<br>data. Uses Sprout Core<br>for more control."])
```
martonvago, line 35: Adding manual breaks to reduce the width of this very wide diagram

docs/guide/packages.qmd
```diff
@@ -37,8 +37,8 @@ you need to establish our properties.
 Sprout has several helper classes, such as `PackageProperties`,
 `LicenseProperties`, and `ContributorProperties`, to make it easier for
 you to make properties with the correct fields filled in. See the guide
-on [properties](/docs/guide/properties.qmd) for more information about
-these classes.
+on [properties](/docs/guide/properties.qmd#properties) for more
```
martonvago, line 40: This doesn't exist, but that's OK

filters/pdf-internal-link.lua
martonvago, line 1: This removes the path to the qmd file in links and leaves only the `#my-section` bit. This way, internal links work both in HTML and PDF.
All titles are automatically labelled in the Latex output, so they can be targeted using `#...`.

index.qmd
```diff
@@ -1,6 +1,8 @@
 ---
 title: Grow structured, organized, and *usable* data
-toc: false
+format:
+  seedcase-theme-html:
+    toc: false
```
martonvago, line 5: To keep the TOC in PDF

index.qmd
```diff
@@ -62,6 +64,8 @@ share your ideas or contribute code. Your input makes us better!
 <!-- TODO: Add link to a guide on how to contribute in more detail -->
 :::
 
+::: {.content-hidden when-format="pdf"}
 ::: landing-page-block
 ## Image by [Storyset](https://storyset.com/data) {.appendix}
 :::
+:::
```
martonvago, line 71: The hero image is not rendered in PDF, so we don't want this showing up as a section.

justfile
```diff
@@ -35,6 +35,17 @@ build-website:
   uv run quartodoc build
   uv run quarto render --execute
 
+# Build the documentation as PDF using Quarto
+build-pdf:
+  # To let Quarto know where python is.
+  export QUARTO_PYTHON=.venv/bin/python3
+  uv run quarto install tinytex
+  # For generating PNGs from Mermaid diagrams
```
martonvago, line 43:
```suggestion
  # For generating images from Mermaid diagrams
```

justfile
```diff
@@ -35,6 +35,17 @@ build-website:
   uv run quartodoc build
   uv run quarto render --execute
 
+# Build the documentation as PDF using Quarto
+build-pdf:
+  # To let Quarto know where python is.
+  export QUARTO_PYTHON=.venv/bin/python3
+  uv run quarto install tinytex
+  # For generating images from Mermaid diagrams
+  uv run quarto install chromium
```
martonvago, line 44: No idea if this is an okay way of expressing that these are dependencies of this workflow. Is there another place I can declare them?

filters/pdf-internal-link.lua
martonvago, line 1: This removes the path to the qmd file in links and leaves only the `#my-section` bit. This way, internal links work both in HTML and PDF.
All titles are automatically labelled in the Latex output, so they can be targeted using `#...`.

_quarto-pdf.yml
lwjohnst86: Interesting... I think it might be because it needs to use a `seedcase-theme-pdf` version? Once we see how this works, maybe we can move some of this into the `seedcase-theme` repo.

lwjohnst86 (approved):
Looks amazing!! :tada: Thanks @martonvago :grin:

_quarto-pdf.yml
martonvago: I tried adding a pdf format in the theme (to create `seedcase-theme-pdf`), but I personally couldn't get it to work. But I agree that this is key!

PR #1330, opened 2025-05-14
build: ➕ add `vulture` to `dev` with `just` recipe to check for unused code

### LINKED ISSUE #1223: Include workflow/build for checking unused code
There's a package called vulture that checks for dead/unused code. Would be helpful for us to have an automatic way of checking this.

signekb:
## Description

This PR adds `vulture` to our `dev` dependencies and a `just` recipe that uses vulture to check for unused code in the `src/` and `tests/` directories. 

The `check-unused` recipe is also added to `run-all`.

I thought it made sense to the check for unused code add it as a `just recipe` instead of e.g., a workflow, but let me know if you agree or not :grin: 

Closes #1223 

## Checklist

- [X] Ran `just run-all`

### DIFF
--- a/justfile
+++ b/justfile
@@ -59,8 +59,8 @@ check-spelling:
 check-unused:
   # exit code=0: No unused code was found
   # exit code=3: Unused code was found
-  # Three confidence values: 
+  # Three confidence values:
   # - 100 %: function/method/class argument, unreachable code
-  # - 90 %: import 
+  # - 90 %: import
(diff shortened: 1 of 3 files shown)


signekb:
When this has been reviewed and merged, I’ll add it to `template-python-project` as well 👌

lwjohnst86 (approved):
Nice! You can add it to the workflow in `.github` repo easily because the workflow uses the just recipe.

PR #1345, opened 2025-05-25
feat: :sparkles: add properties template

### LINKED ISSUE #1343: Create template for package properties

martonvago:
# Description

This PR adds a template for package properties (and supporting code).

I imagine the flow for using this like:

1. In a new package without a `datapackage.json`: either copy the template to `scripts/properties.py` or call `create_properties_template()` from the package root.
2. Open the template and fill in the required fields. Required fields are uncommented. Comment out and fill in any optional fields you need.
3. Run `main.py`. This updates `datapackage.json` .

Closes #1343

## Checklist

- [x] Added or updated tests
- [x] Updated documentation
- [x] Ran `just run-all`


lwjohnst86 (commented):
just some small comments :)

src/seedcase_sprout/sync_package_properties_template.py
```diff
@@ -0,0 +1,72 @@
...
+    properties_path = package_path.properties()
+    package_properties = (
+        read_properties(properties_path)
+        if properties_path.exists()
+        else PackageProperties.from_default()
+    )
+
+    env = Environment(
+        loader=FileSystemLoader(TEMPLATES_PATH),
+        autoescape=select_autoescape(disabled_extensions=("py.jinja2")),
```
martonvago, line 40: We don't want various inputs to the template escaped (e.g. quote marks) because that will mess up the Python code. I think this is mostly appropriate for HTML anyway?

src/seedcase_sprout/sync_package_properties_template.py
```diff
@@ -0,0 +1,73 @@
...
+    )
+    env.filters["quote_str"] = lambda value: json.dumps(value, ensure_ascii=False)
+    env.filters["comment"] = _comment
+    env.filters["strip_n"] = lambda text: text.lstrip("\n").rstrip("    ")
```
martonvago, line 44: This is to remove empty lines between list items

src/seedcase_sprout/use_package_properties_template.py
```diff
@@ -0,0 +1,70 @@
...
+    Raises:
+        ValueError: If the `resources` field is set on the input properties.
+        ValueError: If the template path is incorrect.
+        ExceptionGroup: If there is an error in the properties. A group of
+            `CheckError`s, one error for each failed check.
+    """
+    if properties.resources is not None:
+        raise ValueError(
+            "The `resources` field must not be set. Use the resource template to define"
+            "resource properties."
```
martonvago, line 36:
```suggestion
            " resource properties."
```

src/seedcase_sprout/paths.py
```diff
@@ -46,7 +48,7 @@ class PackagePath:
         ```
     """
 
-    def __init__(self, path: Path | None = None) -> Path:
+    def __init__(self, path: Path | None = None):
```
lwjohnst86, line 51: The return type isn't needed here?

src/seedcase_sprout/templates/package-properties.py.jinja2
```diff
@@ -0,0 +1,53 @@
...
+    {{ required("description", properties.description) }}
+    {% call(license) required_list('licenses', properties.licenses) %}
+        sp.LicenseProperties(
+            ## Required:
+            {{ required("name", license.name) }}
+            ## Optional:
+            {{ optional("path", license.path) }}
+            {{ optional("title", license.title) }}
+        ),
+    {% endcall %}
```
lwjohnst86, line 19: Hm, is this all necessary? I'm trying to understand how this all would be used. Would something like this not be sufficient as a starting point for being a template to work from?

```suggestion
    name="",
    title="",
    description=(
        ""
    ),
    license=sp.LicenseProperties(
        # Required
        name="",
        # Optional
        path="",
        title=""
    ),
```

src/seedcase_sprout/templates/package-properties.py.jinja2
```diff
@@ -0,0 +1,53 @@
...
+            {{ optional("email", source.email) }}
+            {{ optional("version", source.version) }}
+        ),
+    {% endcall %}
+    ## Autogenerated:
+    {{ optional("id", properties.id) }}
+    {{ optional("version", properties.version) }}
+    {{ optional("created", properties.created) }}
+)
+
```
lwjohnst86, line 52:
```suggestion

sp.write_package_properties(package_properties)
```

To end the script with writing to the `datapackage.json`.

tests/test_package_properties_template.py
lwjohnst86, line 1: I think with the comments I've made, much of this might be removed or re-written. This test file only really needs to test that the file was created in the right location.

lwjohnst86 (changes requested):
I have mostly high level comments here, but I think that they will result in a lot of changes.

martonvago:
@lwjohnst86 just before I work through your comments, was it obvious that the jinja template is a template for the `scripts/package-properties.py` Sprout template? So the template from a user perspective is this (generated) file (uploading it as a txt because GitHub didn't allow a py file here):
[template.txt](https://github.com/user-attachments/files/20458731/template.txt)

src/seedcase_sprout/paths.py
martonvago: I realised this now, but the `__init__` method always returns `None`. (There's a `__new__` method for returning the instance.) mypy [says](https://mypy.readthedocs.io/en/stable/class_basics.html#annotating-init-methods) here it's ok to leave out `-> None`. But this is the `PackagePath` class anyway, not `Path`. :sweat_smile: 

(Not sure if there are misannotated inits elsewhere.)

src/seedcase_sprout/sync_package_properties_template.py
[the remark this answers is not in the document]
martonvago: I was thinking: if `datapackage.json` changes / is changed using a flow that doesn't go through the template, then the  template becomes out of sync with `datapackage.json`. So you then won't be able to use the template to edit  `datapackage.json`, unless you update it by hand.

Maybe resource properties are the better example: let's say you extract resource properties and write them to `datapackage.json`. Then you probably want to edit them using the template? That is only possible if the template is not empty, but contains the extracted fields.

Or is this not a concern?

src/seedcase_sprout/templates/package-properties.py.jinja2
martonvago: I think this would remove any resource properties in an existing `datapackage.json`.
It would also pass the (caller's) cwd as the `path` arg to `write_package_properties`.

src/seedcase_sprout/sync_package_properties_template.py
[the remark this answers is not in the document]
lwjohnst86: We can assume the `PackageProperties -> datapackage.json` is unidirectional, not the other way around. If someone does, we assume they know what they are doing and that they will fix it in the `PackageProperties`.

A note about the wording. Once the template gets created, as soon as it is filled it, it is no longer a template and just a script with the `PackageProperties` filled in.

And yes, for the resources, it would be "extract_properties -> insert into use_resource_properties_template -> makes a script with some of the extracted properties -> datapackage.json"

But there won't be any syncing. The template is used to get things started and after that we/Kris can take over.

src/seedcase_sprout/templates/package-properties.py.jinja2
lwjohnst86: Yea, I haven't been happy with the flow we've currently described for this... I think I will make something in the seed beetle repo to show a potential way of doing this.

src/seedcase_sprout/templates/package-properties.py.jinja2
```diff
@@ -0,0 +1,53 @@
...
+sp.use_package_properties_template(package_properties, Path(__file__))
```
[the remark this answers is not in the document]
lwjohnst86: "This should happen when the template is executed". No, the template function only creates a skeleton script that will contain the package properties. Only thing that gets inserted into the script is the `id`, `created`, and `version`. Once we/Kris fill in that script, we run it to write to datapackage.

lwjohnst86 (changes requested):
Hopefully some more clarifying comments :)

src/seedcase_sprout/templates/package-properties.py.jinja2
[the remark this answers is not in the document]
martonvago: But does the template (/ filled in script) write to `datapackage.json`? Because then it has to be able to locate it (even if there is no syncing), right? Or maybe I'm just completely misunderstanding it :sweat_smile:

src/seedcase_sprout/sync_package_properties_template.py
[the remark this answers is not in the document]
martonvago: Okay so just to summarise for myself:
- package properties template: only autogenerated fields filled in, rest empty. User is in charge of updating it.
- resource properties template: same as above, but extracted fields are also filled in.

So we still need a jinja template for both, and a function to populate that template, but only the fields mentioned above.

signekb (commented):
Very nice work, @martonvago ! This is going to be so great! 🌟 🌟 
I've gone through the comments and will hold off with my review until they have been resolved 🌱

martonvago:
> See also [seedcase-project/example-seed-beetle#29](https://github.com/seedcase-project/example-seed-beetle/pull/29) for an example of how things might work together.

I see, thanks, I've reviewed it now! I think it's a very valid approach :+1: :+1:

src/seedcase_sprout/templates/properties.py.jinja2
lwjohnst86, line 1: Niiiiccccceeeeee

lwjohnst86 (approved):
Super niceeeee :star:

signekb (approved):
Looks good to me 👍

PR #1348, opened 2025-05-27
refactor: :recycle: simplify and move into internals

lwjohnst86:
# Description

Moved the internal resource data path creator function into internals and split out the check functionality. Since it's use often co-incides with checking (like in `get_sprout_resource_errors.py`) and since it's cleaner if not too many distinct actions (creating but also checking) happen within an internal function, I moved that bit out. Also simplifies the return type.

This PR needs a quick review.

## Checklist

- [x] Added or updated tests
- [x] Updated documentation
- [x] Ran `just run-all`

### DIFF
--- a/src/seedcase_sprout/internals/create.py
+++ b/src/seedcase_sprout/internals/create.py
@@ -1,15 +1,9 @@
 from pathlib import Path
 from typing import Any
 
-from seedcase_sprout.sprout_checks.is_resource_name_correct import (
-    _is_resource_name_correct,
-)
-
 
 def _create_relative_resource_data_path(resource_name: Any) -> str | None:
     """Creates a stringified relative path to the resource data file based on the name.
-
-    The path is created only if the resource name is correct.
 
     Args:
         resource_name: The name of the resource.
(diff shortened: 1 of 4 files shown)


src/seedcase_sprout/internals/create.py
```diff
@@ -0,0 +1,15 @@
+from pathlib import Path
+from typing import Any
+
+
+def _create_resource_data_path(resource_name: Any) -> str:
```
martonvago, line 5:
```suggestion
def _create_resource_data_path(resource_name: str) -> str:
```

src/seedcase_sprout/internals/create.py
```diff
@@ -0,0 +1,15 @@
+from pathlib import Path
+from typing import Any
```
martonvago, line 2:
```suggestion
```

src/seedcase_sprout/internals/create.py
martonvago: Now we are making this assumption

martonvago (changes requested):
Just a small comment

signekb (approved):

PR #1366, opened 2025-05-30
refactor: ♻️ simplify to only `write_properties()`

### LINKED ISSUE #995: Have one `write_properties()` function

### LINKED ISSUE #1136: Revise so resource properties are created through `write_resource_properties` like with package properties

### LINKED ISSUE #1208: Simplify `write_resource_properties()` to not depend on reading properties
Right now, it reads the properties internally, which is not a great coupling going on.

### LINKED ISSUE #1253: Refactor `write_resource_properties()` to not read from the file

### LINKED ISSUE #1354: [discussion]: Consider simplifying things so we only have one `write_properties()` function

I made a PR in the seed beetle repo: https://github.com/seedcase-project/example-seed-beetle/pull/29 that shows one workflow that feels a bit better than our current ones.

The whole thing around the resource properties has been bothering me a lot over the many months, but I couldn't put my finger on it. Maybe this will resolve some of it more.

Instead of having `write_resource_properties()` and `write_package_properties()`, we simplify it into `write_properties()`. Then we also get rid of the `update_properties()` functions (as we've talked about previously). And then, the properties written in `PackageProperties` is the "source of truth". Whenever we make a change to it, the `write_properties()` overwrites everything in the `datapackage.json` (rather than our clumsy attempts at trying to update things whenever adding the resource properties).

### LINKED ISSUE #1358: Refactor to only have one `write_properties()` function
See [#1354](https://github.com/seedcase-project/seedcase-sprout/issues/1354).

signekb:
# Description

Closes #1354 
Closes #1358
Closes #1208 
Closes #1136 
Closes #1253
Closes #995 

## Checklist

- [X] Added or updated tests
- [X] Updated documentation
- [X] Ran `just run-all`


lwjohnst86 (approved):
Nice! I was about to say to add tests where there are no resources included. But then thought, maybe we keep it simple and say you *need* to have resources added. And with our current workflow, that will be fine, since we (or users) can write the package properties in the Python script and wait to write when the resources has been added too.

PR #1369, opened 2025-06-02
feat: :sparkles: add resource properties template

### LINKED ISSUE #1297: Create template files for making properties
Like a template Python script with the full list of properties to fill in plus which are required etc.

### LINKED ISSUE #1311: initial requirements capture for use of template in creation of datapackage and ressources
The way I (as a user) would like to work with datapackages (including ressources) follows below. 

**Datapackage:**
To start with, I would like to be able to easily see all the parameters I can set for a datapackage.json file. This could either be as a comprehensive guide on the website, or as a template in the repo where I can then remove the fields that I don't want to complete for the datapackage. I am leaning towards a guide on the website, with the option of making a copy of a blank set of code with all relevant fields. This way it would be possible to describe which fields must be completed and which can be taken out. 

I would then like to copy the blank code into a datapackage-properties.[something] file in the relevant GitHub folder, complete the fields, and save the file as my 'master' properties file. Then run a script which takes my properties and creates the .json file. Then when I want to update or change one or more properties I go back to my 'master' file and make the changes there (fx adding a contributor or updating a url), and run the script again to make the change.

Having the properties as a separate file will also allow me a quick overview of the state of the datapackage.json file without me having to know that it exists (less risk of me tampering with it directly).

**Ressources:**
Once I've done the initial datapackage work, I'd like to have two options of adding ressources to the datapackage. 

The first would be to take a script, point it at a data file, and then extract as much information from the file as possible. This should be formatted in the way needed for adding it to the datapackage.json file. I would then save it in the same way as the datapackage-properties file, and manually complete it. 

The second would be a guidance webpage like the one for the datapackage, where I can copy the relevant empty code into a new file, and complete all of it in advance of getting my data. It is often the case that you know what data you will be receiving before you get it, so to optimise the time it may be convenient to be able to complete this in advance. 

Once the resource properties have been completed I'd then run a script to add them to the datapackage, and if changes needs to be made, use the same procedure as for the datapackage. 

**Files:**
From an ease of use perspective I'd like all of my master files to be collated in a separate folder, so that they don't get mixed in with the rest of the files. I would also ideally have a file per ressource which I'm allowed to name, so that it isn't too difficult to go back to a specific resource properties file and make the needed edits/changes. I am aware that this will make it more complex to code, but I think the alternative with all resources in the same file is begging for the wrong set of properties to be altered. The same applies to the scenario where the files have names which are autogenerated and/or based on some form of numbering.

### LINKED ISSUE #1319: Output extracted properties into a template script to finish editing in the file
With #1264 the extracted properties will be in a `ResourceProperties` object. 

Next step is to output these properties into a template script, so users can edit and finish the resource properties.

### LINKED ISSUE #1344: Create template for resource properties

martonvago:
# Description

This PR adds the resource properties template.

Closes #1297, closes #1344, closes #1319, closes #1311

## Checklist

- [x] Added or updated tests
- [x] Updated documentation
- [x] Ran `just run-all`


src/seedcase_sprout/create_resource_properties_template.py
```diff
@@ -0,0 +1,56 @@
...
+from seedcase_sprout.paths import PackagePath
+from seedcase_sprout.properties import FieldProperties
+from seedcase_sprout.sprout_checks.is_resource_name_correct import (
+    _is_resource_name_correct,
+)
+from seedcase_sprout.write_file import write_file
+
+
+def create_resource_properties_template(
+    resource_name: str | None = None,
```
martonvago, line 16: I thought it made sense to have an optional param for this. Feels like people would know the name of the resource when creating a template. But it's okay to omit it as well.

src/seedcase_sprout/paths.py
```diff
@@ -105,3 +107,15 @@ def resource_batch_files(self, resource_name: str) -> list[Path]:
     def properties_template(self) -> Path:
         """Path to the properties template."""
         return self.root() / "scripts" / "properties.py"
+
+    def resource_properties_template(self, resource_name: str = "") -> Path:
```
martonvago, line 111: I'm happy to do it another way, but the idea was:
- If no name given > `resource_properties_.py`. This matches the name of a newly created nameless template  and the object in that template.
- If name given > `resource_properties_my_resource.py`, with `.` and `-` replaced by `_`. This matches the name of a newly created named template and the object in that template.

martonvago, line 111: I'm happy to do it another way, but the idea was:
- If no name given > `resource_properties_.py`. This matches the name of a newly created nameless template  and the object in that template.
- If name given > `resource_properties_my_resource.py`, with `.` and `-` replaced by `_`. This matches the name of a newly created named template and the object in that template.

src/seedcase_sprout/templates/resource_properties.py.jinja2
martonvago, line 1: Fields I wasn't sure about:
- `schema.missing_values` and `field.missing_values`: I don't think we'll use these, so we could even remove them from the classes?
- `resource.encoding`: doesn't make sense for parquet files.
- `TableSchemaForeignKeyProperties` doesn't seem to have any required fields by default, but (if it's included) it probably should? Otherwise how would the reference work? We could add this to the checks as well.
- Should we include `constraint.json_schema`?

src/seedcase_sprout/__init__.py
```diff
@@ -17,6 +17,7 @@
     check_resource_properties,
 )
 from .create_properties_template import create_properties_template
+from .create_resource_properties_template import create_resource_properties_template
```
lwjohnst86, line 20: I was thinking about this, it should be a verb other than `create_` since we aren't creating a template, we're using it. So maybe `use_`?

src/seedcase_sprout/create_resource_properties_template.py
```diff
@@ -0,0 +1,56 @@
...
+    fields: list[FieldProperties] | None = None,
+    path: Path | None = None,
+) -> Path:
+    """Creates the resource properties template.
```
lwjohnst86, line 20:
```suggestion
    """Create a script using the resource properties template.
```

src/seedcase_sprout/create_resource_properties_template.py
```diff
@@ -0,0 +1,56 @@
...
+
+    The resource name and the fields' name and type information can be included.
+
+    Args:
+        resource_name: The name of the new resource. Defaults to None.
+        fields: The fields (columns) of the new resource. Defaults to None.
+        path: The path to the package folder. Defaults to the current working directory.
+
+    Returns:
+        The path to the template file.
```
lwjohnst86, line 30:
```suggestion
        The path to the newly created script file.
```

src/seedcase_sprout/paths.py
lwjohnst86, line 111:
```suggestion
    def resource_properties_script(self, resource_name: str = "") -> Path:
```

Since it isn't the path to the template, its the path to the newly created file.

src/seedcase_sprout/paths.py
```diff
@@ -105,3 +107,15 @@ def resource_batch_files(self, resource_name: str) -> list[Path]:
     def properties_template(self) -> Path:
```
lwjohnst86, line 107: I think this should be changed too to reflect the changes made here. But in another PR :relaxed:

src/seedcase_sprout/paths.py
```diff
@@ -8,6 +8,8 @@
 
 from pathlib import Path
 
+from seedcase_sprout.internals.create import _in_snake_case
```
lwjohnst86, line 11:
```suggestion
from seedcase_sprout.internals import _in_snake_case
```

Refer to the internals module, not the specific one.

src/seedcase_sprout/internals/create.py
```diff
@@ -12,3 +12,16 @@ def _create_resource_data_path(resource_name: str) -> str:
...
+    """Creates a snake-case version of a package or resource name.
+
+    Args:
+        name: The package or resource name.
+
+    Returns:
+        The package or resource name in snake case.
+
+    """
+    return name.replace(".", "_").replace("-", "_")
```
lwjohnst86, line 27: Also to lower case

src/seedcase_sprout/internals/create.py
```diff
@@ -12,3 +12,16 @@ def _create_resource_data_path(resource_name: str) -> str:
             E.g., "resources/test-resource/data.parquet"
     """
     return str(Path("resources", resource_name, "data.parquet"))
+
+
+def _in_snake_case(name: str) -> str:
```
lwjohnst86, line 17: I swear we had a snake case function. I guess we deleted it :stuck_out_tongue_closed_eyes:

I think this would be better placed in another script called `as.py` or `to.py` or `convert.py`, since this isn't creating, its converting to something else.

src/seedcase_sprout/create_resource_properties_template.py
```diff
@@ -0,0 +1,56 @@
...
+            " should only include alphanumeric characters and `.-_`."
+        )
+    resource_name = resource_name or ""
+
+    env = Environment(loader=FileSystemLoader(TEMPLATES_PATH), autoescape=True)
+    env.filters["_in_snake_case"] = _in_snake_case
+    template = env.get_template("resource_properties.py.jinja2")
+    text = template.render(resource_name=resource_name, fields=fields)
+
+    template_path = PackagePath(path).resource_properties_template(resource_name)
```
lwjohnst86, line 54:
```suggestion
    output_path = PackagePath(path).resource_properties_template(resource_name)
```

or something similar.

src/seedcase_sprout/paths.py
lwjohnst86: Could it be `resource_properties.py` without the `_` at the end?

I also don't looovvveeee that path is here, since it feels like it's starting to bleed out of scope of Sprout. Could we not include a helper function here and leave it up to the person to give the correct path that they want? So like, we can use the template copier like `path = PackagePath() / scripts / resource_properties.py` or something? It isn't often this function is used anyway.

src/seedcase_sprout/templates/resource_properties.py.jinja2
lwjohnst86: I think those are fine decisions :+1: And I think we should leave constraints for now. It will help to see things in action before settling on how it will be used.

lwjohnst86 (changes requested):
I made several comments that will probably require some further changes so I didn't review until the end. But great start!! Really nice to see how it's coming along :grin:

src/seedcase_sprout/create_resource_properties_template.py
martonvago: Ah okay, so call the `resource_properties.py` file a script instead of a template? Then I'll change it everywhere :+1:

src/seedcase_sprout/__init__.py
martonvago: Oh I see what you mean! I was calling both the jinja files and the generated Python files templates, but if we call the Python files scripts, then could this be `create_resource_properties_script`? Just because from a user's perspective maybe it's more informative to say what the function outputs (not how it does it).

src/seedcase_sprout/internals/create.py
martonvago: Yeah, I also think it was deleted at some point. I think this is a simpler version of it because it only works for strings matching the resource / package name pattern.

src/seedcase_sprout/paths.py
martonvago: Sure, re losing the `_`!

martonvago: Sorry, I'm not following your second point. Are you saying we should put this function elsewhere or that it should do something different? :sweat_smile:

src/seedcase_sprout/templates/resource_properties.py.jinja2
martonvago: Okay, then I'll make issues for removing `missing_values` from the classes and adding required checks for `TableSchemaForeignKeyProperties`.

I'll remove constraints from the template/script for now.

src/seedcase_sprout/paths.py
martonvago: Oh I see, okay, sounds good!

src/seedcase_sprout/internals/create.py
martonvago: I just double checked and resource / package names cannot have upper case letters, so they would fail the checks

signekb (changes requested):
Very nice! 🌟 Just some very minor suggestions. Looking forward to trying out the new flow with the templates 🌱

src/seedcase_sprout/internals/create.py
```diff
@@ -1,5 +1,7 @@
 from pathlib import Path
 
+from seedcase_sprout.internals.to import _to_snake_case
+
```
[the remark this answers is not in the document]
martonvago: I think outside `internals` yes, but here this is inside `internals`, so that's why I thought it needed the `.to`. Or does it not?

src/seedcase_sprout/create_resource_properties_script.py
```diff
@@ -0,0 +1,59 @@
...
+        The path to the newly created script file.
+
+    Raises:
+        ValueError: If the resource name is not correct (e.g., contains spaces).
+
+    Examples:
+        ```{python}
+        import seedcase_sprout as sp
+
+        sp.create_resource_properties_script("my-resource")
```
martonvago, line 40:
```suggestion
        with sp.ExamplePackage():
            sp.create_resource_properties_script("my-resource")
```

src/seedcase_sprout/create_resource_properties_script.py
```diff
@@ -0,0 +1,59 @@
...
+        ```
+    """
+    if resource_name and not _is_resource_name_correct(resource_name):
+        raise ValueError(
+            f"The resource name '{resource_name}' is not correct. Resource names"
+            " should only include lowercase alphanumeric characters and `.-_`."
```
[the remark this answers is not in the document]
signekb: 🙈

tests/test_create_resource_properties_script.py
```diff
@@ -0,0 +1,79 @@
...
+
+@mark.parametrize("args", [[], ["my-resource"]])
+def test_existing_script_not_overwritten(tmp_cwd, args):
+    """If a script already exists for a resource, it should not be overwritten."""
+    script_path = create_resource_properties_script(*args)
+    script_path.write_text("test")
+
+    create_resource_properties_script(*args)
+
+    assert script_path.read_text() == "test"
```
signekb, line 79: Should we give a warning or something when the file already exists? So it's clearer what's actually happens?

signekb (commented):
Looks good to me 👍 I have one suggestion

tests/test_create_resource_properties_script.py
signekb: Yeah, it's true we've avoided it so far, but I actually don't remember the reasoning for this? Maybe @lwjohnst86 does? 
I think it makes sense and will give a nicer user experience to include warnings like this to reassure that the file hasn't been overwritten.

My suggestion would be something simple, like this using the `warnings` module: 

```
warn(
            f"Properties script already exists at {script_path}. "
            "The existing file will not be overwritten.",
            UserWarning,
        )
```

Maybe with a custom warning instead of `UserWarning`? 

I'm not sure that we need the warnings to be logged at this point? :)

signekb: Oh, yeah, I think you're right. I'm not sure whether the CLI is still on the table? At least, we've removed it from the documentation, right?

I *guess* you *could* frame a message about not overwriting a file as a warning? Like *warning* the user that it isn't being overwritten, if that's what they intended to do? 🤷‍♀️

lwjohnst86 (changes requested):
Very nice! Just some small comments :relaxed:

src/seedcase_sprout/create_resource_properties_script.py
```diff
@@ -0,0 +1,60 @@
...
+        with sp.ExamplePackage():
+            sp.create_resource_properties_script("my-resource")
+        ```
+    """
+    if resource_name and not _is_resource_name_correct(resource_name):
+        raise ValueError(
+            f"The resource name '{resource_name}' is not correct. Resource names"
+            " should only include lowercase alphanumeric characters and `.-_`."
+        )
+    resource_name = resource_name or ""
```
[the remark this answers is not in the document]
martonvago: The `if` checks whether a non-none (or non-empty) `resource_name` is correct. If the user called the function with a `resource_name`, we want to check that it's correct before using it.

If the user didn't call the function with a `resource_name`, we set it to `""` in line 49. This is because some functions after this expect it to be a string, so it cannot be none.

src/seedcase_sprout/internals/create.py
```diff
@@ -12,3 +14,15 @@ def _create_resource_data_path(resource_name: str) -> str:
...
+def _create_resource_properties_script_filename(resource_name: str = "") -> str:
+    """Creates the filename for the resource properties script from the resource name.
+
+    Args:
+        resource_name: The name of the resource. Defaults to "".
+
+    Returns:
+        The filename.
+    """
+    return f"resource_properties{resource_name and '_'}{_to_snake_case(resource_name)}"
```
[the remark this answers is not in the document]
martonvago: Like: if `resource_name` is truthy (not `""`), then add a connecting `_`. Otherwise, don't add `_`.

src/seedcase_sprout/internals/to.py
```diff
@@ -1,3 +1,16 @@
...
+    """Creates a snake-case version of a package or resource name.
+
+    Args:
+        name: The package or resource name.
+
+    Returns:
+        The package or resource name in snake case.
+
+    """
+    return name.replace(".", "_").replace("-", "_")
```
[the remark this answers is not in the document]
martonvago: Sure! I just didn't add it because package and resource names should already be lower case

src/seedcase_sprout/templates/resource_properties.py.jinja2
```diff
@@ -0,0 +1,59 @@
...
+    # ],
+    # licenses=[
+    #     sp.LicenseProperties(
+    #         ## Required:
+    #         name="",
+    #         ## Optional:
+    #         path="",
+    #         title="",
+    #     ),
+    # ],
```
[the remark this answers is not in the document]
martonvago: We can remove it for sure! I just mechanically added all the fields in the class/schema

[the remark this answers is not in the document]
lwjohnst86: Yea, let's remove it :+1:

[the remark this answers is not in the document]
martonvago: Ok, done!

lwjohnst86 (approved):

PR #1374, opened 2025-06-04
refactor: :fire: remove `MissingValueProperties`

martonvago:
# Description

This PR removes `MissingValueProperties` because in a data frame loaded from tabular data missing values will be represented by `None`. So no custom missing values. 

If you think we should keep this class (in anticipation of other kinds of data?), feel free to close this PR.

This PR needs a quick review.

## Checklist

- [x] Added or updated tests
- [x] Updated documentation
- [x] Ran `just run-all`


lwjohnst86 (approved):
:broom: :soap: Nice clean up :+1:

PR #1383, opened 2025-06-06
refactor: :recycle: don't overwrite the `properties.py` script

lwjohnst86:
# Description

If the file exists, we shouldn't overwrite it. That way, it can be included in the `main.py` function and can be repeated ran without causing issues.

This PR needs a quick review.

## Checklist

- [x] Added or updated tests
- [x] Updated documentation
- [x] Ran `just run-all`


martonvago (changes requested):
Nice!!

signekb (dismissed):
Makes sense. Approved, except Marton's comment 👍

martonvago (approved):

signekb (approved):

PR #1385, opened 2025-06-06
refactor: :recycle: re-export pprint and textwrap

lwjohnst86:
# Description

This is mostly for the guide, but also to make it easier to use these functions that are commonly used when using Sprout. So we don't have to always write `import textwrap` etc.

This PR needs a quick review.

## Checklist

- [x] Ran `just run-all`


lwjohnst86:
This is needed for #1386

martonvago (approved):

signekb (approved):

PR #1392, opened 2025-06-12
fix: :bug: change `fields_match` to be an array in `extract_resource_properties()`

signekb:
# Description

It's supposed to be a string, but there's an error in the JSON schema. According to a PR in the [data package repo](https://github.com/frictionlessdata/datapackage/pull/1051), this should be fixed in datapackage V2.1 🤞 

I encountered this issue bc the website build fails bc there was an error in `resources.qmd` when writing the properties.

This PR needs a quick review.

## Checklist

- [X] Ran `just run-all`

### DIFF
--- a/src/seedcase_sprout/extract_resource_properties.py
+++ b/src/seedcase_sprout/extract_resource_properties.py
@@ -45,7 +45,9 @@ def extract_resource_properties(data: pl.DataFrame) -> ResourceProperties:
 
     resource_properties = ResourceProperties()
     resource_properties.type = "table"
-    resource_properties.schema = TableSchemaProperties(fields_match="equal")
+    # The fields_match property is supposed to be a string, but there's an error in the
+    # JSON schema. This should be fixed in datapackage V2.1.
+    resource_properties.schema = TableSchemaProperties(fields_match=["equal"])
     resource_properties.schema.fields = _extract_field_properties(data)
 
     return resource_properties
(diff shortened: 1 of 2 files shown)


martonvago (commented):
Ah we're seeing this now because the camel/snake case fix was merged in!

But do we want to change the type of `fields_match` in `TableSchemaProperties` to array as well? Because now it's type hinted as a string (well, `FieldsMatchType`). There are also some other references to `fields_match` in the tests.

If we don't want to make a bigger change to the "wrong" type in `TableSchemaProperties`, we could 
- not set `fields_match` in Sprout until the issue is fixed
- fix it ourselves in the schema, but not by editing the JSON directly, but by resetting that one property in `check_properties` / `check_resource_properties` when the schema is loaded

martonvago:
@signekb  Yeah, I think we don't use it in Sprout for anything, so it's okay to not set it at all for now. (We should set it eventually only because Sprout doesn't use the default value for it, which is not obvious if datapackage.json is ever used/shared outside of Sprout.)

martonvago:
Hmm, I don't feel strongly about either deleting it or changing it to an array.
I guess your argument about being consistent with how we dealt with the `list` type missing is convincing though. So I'm happy to change it to an array!

lwjohnst86 (approved):

signekb:
Ah, ok, should we attend that meeting?

PR #1402, opened 2025-06-19
docs: :memo: link to raw sprout logo

signekb:
# Description

Instead of using a relative path. This should work when we publish to PyPI (relative path doesn't).

This PR needs a quick review.

## Checklist

- [X] Updated documentation
- [X] Ran `just run-all`


martonvago (approved):
images

lwjohnst86 (approved):

PR #1403, opened 2025-06-19
refactor: :wrench: set up mypy and fix issues from it

### LINKED ISSUE #1219: Setup mypy to run in project
I don't think this is set up yet.

martonvago:
# Description

This PR sets up mypy and fixes the errors flagged by running it in non-strict mode.

Closes #1219

## Checklist

- [x] Added or updated tests
- [x] Updated documentation
- [x] Ran `just run-all`


src/seedcase_sprout/as_readme_text.py
```diff
@@ -40,7 +40,7 @@ def join_names(licenses: list[LicenseProperties] | None) -> str:
     Returns:
         A comma-separated list of names.
     """
-    return ", ".join(license.name for license in licenses) if licenses else "N/A"
+    return ", ".join(str(license.name) for license in licenses) if licenses else "N/A"
```
martonvago, line 43: This is a very common situation: we know that this property is not none (because we've checked the properties already), but mypy has no way of inferring this.

tests/test_create_properties_script.py
```diff
@@ -42,6 +42,8 @@ def test_works_with_custom_path(tmp_path):
 def load_properties(path: Path) -> PackageProperties:
     """Loads `properties` object from file."""
     spec = spec_from_file_location("test_module", path)
+    assert spec
+    assert spec.loader
```
martonvago, line 46: I thought this was okay to do in tests to tell mypy these are not none. If the assert fails, the test fails -- as it should.

justfile
```diff
@@ -2,7 +2,7 @@
     just --list --unsorted
 
 # Run all build-related recipes in the justfile
-run-all: install-deps format-python check-python check-unused test-python check-security check-spelling check-commits build-website
+run-all: install-deps format-python check-python type-check-python check-unused test-python check-security check-spelling check-commits build-website
```
lwjohnst86, line 5:
```suggestion
run-all: install-deps format-python check-python check-unused test-python check-security check-spelling check-commits build-website
```

Should be included within the  `just check-python`

justfile
```diff
@@ -29,6 +29,10 @@ test-python:
 check-python:
   uv run ruff check .
 
+# Check Python types
+type-check-python:
```
lwjohnst86, line 33:
```suggestion
```

mypy.ini
lwjohnst86, line 1: We'll have to remember to include this in the template repo too.

justfile
lwjohnst86, line 1: We'll need to update the justfile in the template repo too

src/seedcase_sprout/check_data.py
```diff
@@ -89,7 +91,9 @@ def _check_column_names(
     columns_in_data = data.schema.names()
     columns_in_resource = [
         field.name
-        for field in get_nested_attr(resource_properties, "schema.fields", default=[])
+        for field in cast(
```
lwjohnst86, line 94: What is this doing?

src/seedcase_sprout/check_data.py
```diff
@@ -178,16 +183,16 @@ def _get_column_type_error(
         A `ValueError`.
     """
     allowed_types = _map(_get_allowed_polars_types(field.type), str)
-    allowed_types = (
+    allowed_types_str = (
```
lwjohnst86, line 186: Any reason this is needed to change? In general, the object type shouldn't be in the name.

src/seedcase_sprout/get_nested_attr.py
```diff
@@ -36,14 +36,14 @@ class Outer:
         get_nested_attr(Outer(), "middle.inner")
         ```
     """
-    attributes = attributes.split(".")
-    if any(not attribute.isidentifier() for attribute in attributes):
+    attributes_list = attributes.split(".")
```
lwjohnst86, line 39: Same reasoning as above, not necessary to include the object type in the name, unless mypy is suggesting it?

lwjohnst86 (changes requested):
Nice! some suggestions and questions

src/seedcase_sprout/check_data.py
martonvago: The mypy error here was that `allowed_types` starts as a list but then gets reassigned to be a string. So I just introduced a new variable for the string. Could be called anything else.

lwjohnst86 (approved):

signekb (approved):
👏 👏

PR #1428, opened 2025-06-25
refactor: :recycle: raise error for non-Parquet batch files

### LINKED ISSUE #1281: `read_resource_batches()` should check that batch file is a Parquet file before reading
`read_resource_batches()` should check that batch file is a Parquet file before reading.
Otherwise `pl.read_parquet` will throw an error.

martonvago:
# Description

This PR checks that batch files have the `.parquet` extension before reading them. This is to prevent `pl.read_parquet` from throwing its own error.

Closes #1281

This PR needs a quick review.

## Checklist

- [x] Added or updated tests
- [x] Updated documentation
- [x] Ran `just run-all`


src/seedcase_sprout/read_resource_batches.py
```diff
@@ -80,7 +81,15 @@ def _read_parquet_batch_file(
...
         The Parquet file as a DataFrame with a timestamp column added.
+
+    Raises:
+        ValueError: If the file is not a Parquet file.
     """
+    if path.suffix != ".parquet":
+        raise ValueError(
+            "Failed to read batch file. Expected a file with a "
+            f"`.parquet` extension but found {path}."
+        )
```
martonvago, line 92: ((If the file has the `.parquet` extension but is not a Parquet file, reading it with polars will still fail as before.))

lwjohnst86 (dismissed):

signekb (commented):
Ah, that's nice 👍 Just one comment on the docstring.

src/seedcase_sprout/read_resource_batches.py
```diff
@@ -43,6 +43,7 @@ def read_resource_batches(
 
     Raises:
         FileNotFoundError: If a file in the list of paths doesn't exist.
+        ValueError: If the file is not a Parquet file.
         ValueError: If the batch file name is not in the expected pattern.
         ValueError: If the timestamp column name matches an existing column in the
```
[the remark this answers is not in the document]
martonvago: Ooh that's a good find! Okay, I'll do that :+1:

src/seedcase_sprout/read_resource_batches.py
```diff
@@ -34,15 +34,14 @@ def read_resource_batches(
...
+        paths: A list of paths for all the Parquet files in the resource's `batch/`
+            folder. Use `path_resource_batch_files()` to help provide the correct paths
+            to the batch files. Defaults to the batch files of the given resource.
 
     Returns:
         A list of DataFrame objects from all the batch files.
 
     Raises:
-        FileNotFoundError: If a file in the list of paths doesn't exist.
         ValueError: If the batch file name is not in the expected pattern.
```
lwjohnst86, line 45: Is this relevant anymore? We only read files that match the pattern.

src/seedcase_sprout/read_resource_batches.py
```diff
@@ -34,15 +34,14 @@ def read_resource_batches(
...
         ValueError: If the timestamp column name matches an existing column in the
             DataFrame.
```
lwjohnst86, line 47: This would be a failure on our end wouldn't it? Like, we shouldn't even allow column names to be written if it will conflict. And/or that our column name isn't unique enough.

lwjohnst86: I know this is outside this PR though, so maybe I'll open an Issue about it.

lwjohnst86 (approved):

PR #1437, opened 2025-06-25
fix: :adhesive_bandage: fix `mypy` errors in strict mode

### LINKED ISSUE #1427: `mypy` strict mode?

martonvago:
# Description

This PR sets mypy to strict and fixes the new errors.

Closes #1427 

## Checklist

- [x] Added or updated tests
- [x] Updated documentation
- [x] Ran `just run-all`


mypy.ini
```diff
@@ -1,5 +1,9 @@
 [mypy]
 python_version = 3.12
+strict = True
 
 [mypy-quartodoc.*]
 ignore_missing_imports = True
+
+[mypy-tests.*]
+disallow_untyped_defs = False
```
martonvago, line 9: Otherwise it will flag all test functions as lacking a `-> None` return type.

_renderer.py
```diff
@@ -8,10 +8,11 @@
 from quartodoc import MdRenderer, layout
 from quartodoc._griffe_compat import docstrings as ds
 from quartodoc.pandoc.blocks import DefinitionList
+from quartodoc.renderers.md_renderer import ParamRow
 from tabulate import tabulate
 
 
-class Renderer(MdRenderer):
+class Renderer(MdRenderer):  # type: ignore[misc]
```
martonvago, line 15: `MdRenderer` is not typed

src/seedcase_sprout/examples.py
```diff
@@ -389,7 +389,7 @@ def __enter__(self) -> PackagePath:
 
         return package_path
 
-    def __exit__(self, *_) -> None:
+    def __exit__(self, *_) -> None:  # type: ignore[no-untyped-def]
```
martonvago, line 392: There are some unused args here (`*_`) that `mypy` wanted listed and typed. As we don't use them, I thought it was simpler to ignore this error.

src/seedcase_sprout/internals/functionals.py
martonvago, line 1: We have to use generics here to have the map functions preserve the type information in the callables.

tests/test_paths.py
```diff
@@ -51,7 +51,8 @@ def test_resource_batch_files_returns_file_paths_when_batches(tmp_path):
         batch_folder = package_path.resource_batch(resource)
         batch_folder.mkdir(parents=True)
         (batch_folder / "sub-folder").mkdir()
-        [(batch_folder / file).touch() for file in ["file", "file.txt", "file.parquet"]]
+        for file in ["file", "file.txt", "file.parquet"]:
+            (batch_folder / file).touch()
```
martonvago, line 55: Do not use list comprehension when result is not used

mypy.ini
lwjohnst86: I'd say we ignore all files in `test`. We don't need mypy running on those.

signekb (commented):
Did a rather surface-level review and it looks good to me 👍 Just one comment (+ the discussion from last week about what to do with the checks of test files), otherwise approved ✅

src/seedcase_sprout/examples.py
martonvago: Sounds good to me! :heavy_check_mark:

signekb (approved):
✅ ✅

lwjohnst86 (approved):
:tada:

PR #1486, opened 2025-07-23
docs: :memo: update DOI to refer to all versions

lwjohnst86:
# Description

The previous DOI only referred to the first version that was released. This DOI refers to all versions and always resolves to the latest version.

This PR needs a quick review.

## Checklist

- [x] Updated documentation
- [x] Ran `just run-all`


signekb (approved):
Ahhh, nice! I don't know much about DOIs, but very nice if this works 🌟

PR #1493, opened 2025-07-24
fix: :bug: do not create duplicate batch files

### LINKED ISSUE #1478: [bug]: Batch file is duplicated when `write_resource_batch()` is rerun with `main.py`
### Steps to reproduce

When `write_resource_batch()` is a part of the `main.py` pipeline, the same data is written to a new file with a new UUID (i.e., data is duplicated but in a new batch file). 

### Expected result

The same data should only be written once, i.e., not be duplicated into a new file. 

### Actual result (the bug)

The same data is written to a new file with a new name. 

```shell

```

### Severity/Priority

None

martonvago:
# Description

This PR makes it so that batch files with exactly the same data are not duplicated.

Closes #1478 

This PR needs a quick review.

## Checklist

- [x] Added or updated tests
- [x] Updated documentation
- [x] Ran `just run-all`

### DIFF
--- a/src/seedcase_sprout/write_resource_batch.py
+++ b/src/seedcase_sprout/write_resource_batch.py
@@ -23,7 +23,8 @@ def write_resource_batch(
     Use this function to write the original data that is in a tidy Polars
     DataFrame to the resource folder provided by the `path` property in the
     `resource_properties`. The function saves the DataFrame as file with a
-    timestamped, unique name, as a backup. See the
+    timestamped, unique name, as a backup. If the data is exactly the same as
+    the data in one of the existing batch files, it won't be duplicated. See the
     [design](https://sprout.seedcase-project.org/docs/design/) docs for an
     explanation of this batch file. Data is always checked against the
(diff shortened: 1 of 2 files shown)


src/seedcase_sprout/write_resource_batch.py
```diff
@@ -58,7 +59,15 @@ def write_resource_batch(
...
     check_data(data, resource_properties)
 
-    batch_path = PackagePath(package_path).resource_batch(str(resource_properties.name))
+    package = PackagePath(package_path)
+    resource_name = str(resource_properties.name)
+
+    for batch_path in package.resource_batch_files(resource_name):
+        batch = pl.read_parquet(batch_path)
+        if data.equals(batch):
+            return batch_path
```
martonvago, line 68: Very simple strategy. Probably rather slow, but that hasn't been much of a concern so far

lwjohnst86 (approved):
Nice! :grin:

signekb (approved):

PR #1536, opened 2025-09-10
ci: :fire: don't need `update-from-template.yml` workflow

lwjohnst86:
# Description

Most of the time, when you update from the template, there are merge conflicts to deal with, so it can't really be done automatically. That's also ok, we can update whenever we directly work on the repo or whenever we feel like it.

This PR needs a quick review.


signekb (approved):
Maybe slightly rewrite the PR title and then ✅

PR #1568, opened 2025-10-21
test: :white_check_mark: fix failing test by adding scale and precision to `pl.Decimal`

### LINKED ISSUE #1567: Fix failing test after Polars update
# Description 

See failed workflow [here](https://github.com/seedcase-project/seedcase-sprout/actions/runs/18489032672/job/52678142545).

There seems to have been an update to the (unstable) `pl.Decimal` which makes one test fail.

signekb:
# Description

I believe this test started failing bc of a Polars update. [`pl.Decimal`](https://docs.pola.rs/api/python/stable/reference/api/polars.datatypes.Decimal.html) is considered unstable by the Polars team and "might change without it being considered a breaking change".
I fixed it by adding scale and precision to it in the one test that failed. We do use `pl.Decimal` in another test but since it doesn't fail, I've left it as is.

Closes #1567 

Needs a quick review.

## Checklist

- [X] Ran `just run-all` (will make a separate PR with changes unrelated to this PR)


lwjohnst86 (approved):
:tada: amazing :grin:

PR #1570, opened 2025-10-22
docs: :memo: remove reference to GitHub installation in guide

signekb:
# Description

The command doesn't install from source/GitHub but from PyPI.

Needs a quick review.

## Checklist

- [X] Formatted Markdown
- [X] Ran `just run-all`


lwjohnst86 (approved):
:+1: Good catch there.

PR #1585, opened 2025-11-04
docs: :memo: clarify intro and data constraints

joelostblom:
# Description

I think it's helpful to be more specific on what Sprout does already on the intro page (without getting into too many details). I was inspired by some of the language in https://sprout.seedcase-project.org/docs/design/ and elaborated on it.

Let me know if I have done the commits correctly here; my understanding is that since this is not a repo about only documentation, I should use the `docs: :memo:` indicator for all docs-related commits.

Closes https://github.com/seedcase-project/seedcase-sprout/issues/1574 and starts to address https://github.com/seedcase-project/seedcase-sprout/issues/1575

Needs an in-depth review.

## Checklist

- [x] Formatted Markdown
- [ ] Ran `just run-all`

PR #1588, opened 2025-11-05
docs: :memo: minor edits to overview page

signekb:
# Description

Based on #1585, I noticed some (very) minor things, I wanted to edit.

Needs a quick review.

## Checklist

- [X] Ran `just run-all`

### DIFF
--- a/docs/index.qmd
+++ b/docs/index.qmd
@@ -12,8 +12,8 @@ Sprout helps to ensure and enforce that research data are well-designed,
 discoverable, well-documented, and ultimately more (re)usable for later
 analyses.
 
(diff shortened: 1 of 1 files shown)


lwjohnst86 (approved):

joelostblom (approved):
Agree with the edits 👍  I actually considered the your suggested order of the bullets to match the ordering of the words in the preceding sentence, but went with metadata first because it seemed to me to be more central to what sprout does. But then I thought it didn't flow as well to say "metadata and data" as saying "data and metadata", so I ended up keeping "data" first in the text and "metadata" first in the bullets. But I like the flow of your suggestion better so lets change.

Issue #157, opened 2024-02-12
Discussion: Postgres vs Parquet, which to implement right away?


lwjohnst86:

K-Beicher:
In my view definitely postgres, as that can also handle user security.  I'm also worried that if we save data in parquet then we'll have a job converting it back into for instance csv, if a user would like to export their data in a format that can be read my another system.

Does anyone know how STATA, SPSS, R and Python deals with parquet?

lwjohnst86:
@K-Beicher yea, I think that might be a good starting point. Saving between Parquet and CSV is very easy, so it isn't an issue.

R and Python handle Parquet just fine, since there are packages available to do that. SPSS I think can handle it, SAS I think so too. Stata unfortunately doesn't.

signekb:
Kris and Luke: We want to handle files in a database structure (bc we can't handle users in parquet files, for instance).
Philip: Will check up if polars can handle this.

philter87:
So there are many considerations in this discussion and I am not sure where to start: 
- **CSV, parquet or database tables**. We are using polars to parse a CSV. Polars is able to write to CSV, parquet and surprisingly non-existing database tables with: `df.write_database(table_metadata.name, database_url, engine="adbc")`. My biggest argument for not using a database for the actual data was the complexity of dynamically creating tables. But this is not an issue with polars (and it works for both sqlite and postgres, so we can run it locally without spinning up a postgres database)
- **Exporting to CSV**. Polars can easily create CSVs no matter if it is based on a parquet file or a database
- **Security**. I am not too concerned with the security aspect of having files on the server (this was mentioned in #62). The web app will decide what people can view and edit. Normal users will not be able to directly see or edit these files. The admin might be able to see the files directly on the server, but the admin should not make manual corrections to these
- **Performance**. I ran some small tests comparing sqlite vs parquet with a big CSV (142MB and 6.5 mio rows).
  - **Speed**: SQLite took 5 seconds and parquet took 0.8 seconds to persist the CSV
  - **Size**: SQLite uses 160MB and parquet uses 18 MB
- **Appending data**. With SQLite it is more obvious to add new data to a table. This is not possible natively with parquet files. You would need to have multiple files or merge files
- **Joins?** Using a SQL database will have the possibility to do joins. I think "joins" are too complicated for normal users, but if this is needed then we need to use a SQL database
[...]

PR #1597, opened 2025-11-12
docs: :memo: clarify guide on resource-level metadata

### LINKED ISSUE #1596: [discussion]: Why is the raw data file used for inferring field metadata instead of the cleaned one?

[This section of the docs](https://sprout.seedcase-project.org/docs/guide/resources#extracting-properties-from-the-data) includes the following passage:

> Let’s add these steps to our main.py file: First, we need to load the original data from the raw/ folder into a Polars DataFrame, then we can extract the field properties from it, and use these properties in the creation of our resource properties script:

I am a bit confused as to why the raw data file is used for field metadata inference. Wouldn't the cleaned data file be more useful since column names and types can change during cleaning?

On a related topic, do we expect the cleaning steps to be written out in `main.py`? My hunch would be that we don't since cleaning is an art in itself that requires a certain amount of data exploration and thus likely happens in its own script with a preliminary EDA notebook.

joelostblom:
# Description

Further clarification to the intro guide based on my understanding of how things work. I left the last two section untouched due to me not quite grokking them as elaborated on in https://github.com/seedcase-project/seedcase-sprout/issues/157.

Some additional description in commit messages and inline comments.

closes #1596 (although that not the main issue motivating this PR)

Needs an in-depth review.

## Checklist

- [x] Formatted Markdown
- [ ] Ran `just run-all`


docs/guide/resources.qmd
```diff
@@ -1,49 +1,40 @@
...
-    edit the resource properties script and run the `main.py` file to
-    re-create the properties and write them to the
-    `datapackage.json` file.
-
-Many of these steps are taken care of by functions in Sprout, but some
-require writing your own code.
+For this guide, you will use (fake) data that is already tidy. You can
+[find the data
+here](https://raw.githubusercontent.com/seedcase-project/data/refs/heads/main/patients/patients.csv).
+We've placed this data in a `raw/` folder in the data package and called
```
lwjohnst86:
```suggestion
We've placed this data in a `raw/` folder in the Data Package and called
```

docs/guide/resources.qmd
```diff
@@ -101,59 +78,123 @@ print(file_tree(package_path.root()))
...
+that data actually contains what you expect it to contain. These checks can
+protect you from human errors introduced when adding new data or editing an
+existing data resource.
 
-::: callout-warning
-`extract_field_properties()` extracts the field properties from the
-Polars DataFrame's schema and maps the Polars data type to a Data
-Package [field
-type](https://datapackage.org/standard/table-schema/#field-types).
+### Creating a script to help manage data-level metadata
```
lwjohnst86:
```suggestion
### Creating a script to manage the metadata
```

This guide is already about the resource, so this header can be simplified.

docs/guide/resources.qmd
```diff
@@ -101,59 +78,123 @@ print(file_tree(package_path.root()))
...
+```
+
+1. We already had this line in our script from the previous section of the
+   guide.
+2. This is the only new line of code. It creates the resource properties
+   script. Note that the script is only created once, it will not be
+   overwritten if it already exists.
+3. We have removed the writing of `datapackage.json` and the creation of the
+   readme to make it easier to see what is new. You will learn how to write the
+   data-level metadata and the readme later in this guide.
```
lwjohnst86:
```suggestion
   resource-level metadata and the README later in this guide.
```

docs/guide/resources.qmd
```diff
@@ -101,59 +78,123 @@ print(file_tree(package_path.root()))
...
-original data from the `raw/` folder into a Polars DataFrame, then we
-can extract the field properties from it, and use these properties in
-the creation of our resource properties script:
+As with the template for the package-level metadata, the comments in this file
+indicates which properties are required and which are optional. You can see
+that you would need to fill out a `title` and a `description` just as we did
+previously for the package-level metadata. However, you also need to fill out
+the `name` and the `type` inside `FieldProperties` (a "field" is the same as a
+"column" or "variable" in your dataset). Doing this manually can be tedious for
+datasets with many columns, so sprout provides a way to extract this metadata
```
lwjohnst86:
```suggestion
datasets with many columns, so Sprout provides a way to extract this metadata
```

docs/guide/resources.qmd
```diff
@@ -101,59 +78,123 @@ print(file_tree(package_path.root()))
...
+directly from the data file.
+
+### Extracting column metadata directly from the data
+
+To ease the process of adding `fields` to your resource metadata, Sprout
+provides a function called `extract_field_properties()`, which allows you to
+extract metadata from each column in your dataset. To use this function, we need to read in our data as a Polars DataFrame and clean it until we are happy with the name and data type for each column.
+In the code snippet below, we have added all the required steps to our `main.py` file.
```
lwjohnst86:
```suggestion
In the code chunk below, we've added all the required steps to our `main.py` file.
```

docs/guide/resources.qmd
```diff
@@ -218,25 +269,43 @@ Your folders and files should now look like:
...
+    sep='\n'
+)
 ```
 :::
 
-In the resource properties script, the `name` property is set to
-`patients`. However, the two other required properties, `title` and
-`description`, are empty. You will need to fill these in yourself in the
-script, like so:
+## Writing the data-level metadata to `datapackage.json`
```
lwjohnst86:
```suggestion
## Writing the metadata to `datapackage.json`
```

docs/guide/resources.qmd
```diff
@@ -289,92 +358,139 @@ resource_properties_patients = sp.ResourceProperties(
...
 ```
 
-To include these properties in your data package, you need to include
-the resource properties in the `properties.py` file of your data
-package. You can do this by adding the following lines to the
-`properties.py` file:
+You can click the banner below to view the full file at this point:
+
+::: {.callout-tip icon=false collapse=true}
+## `scripts/package_properties.py`
```
lwjohnst86:
```suggestion
## Inside the `package_properties.py` script
```

lwjohnst86 (changes requested):
Nice! Some comments :clinking_glasses:

joelostblom (commented):
Thank for the review @lwjohnst86 , updated!

docs/guide/resources.qmd
```diff
@@ -1,17 +1,17 @@
...
+title: "Resource metadata"
 order: 3
 jupyter: python3
 ---
 
 ## What is a data resource
 
-So far, we have seen how to create and manage project-level metadata for our
+So far, we have seen how to create and manage package metadata for our
 data package. In this section, we will learn about how we can add data files
```
signekb:
```suggestion
data package. In this section, we will look at how we can add data files
```
Maybe?

docs/guide/resources.qmd
```diff
@@ -101,16 +101,16 @@ that data actually contains what you expect it to contain. These checks can
...
-Like with the project-level metadata, you will create a script for the data
-resource that allows you to easily edit the metadata properties later on. You
-can do this by using the `create_resource_properties_script()` function. This
-function needs the `resource_name` parameter to specify the name of the
-resource, which is used to identify the resource in the data package. It is
-required and *should not* be changed, since it's used in the file name of the
-data resource's properties script. Since a data package can contain multiple
-resources, the resource's `name` must also be unique.
+As we did for the package metadata, you will create a script to manage the
+resource medata. You can do this by using the
```
signekb:
```suggestion
As you did for the package metadata, you will create a script to manage the
resource medata. You can do this by using the
```
Maybe? At least the switch from "we" to "you" feels a bit odd to me here :)

docs/guide/resources.qmd
```diff
@@ -101,16 +101,16 @@ that data actually contains what you expect it to contain. These checks can
...
+`create_resource_properties_script()` function. This function needs the
+`resource_name` parameter to specify the name of the resource, which is used to
+identify the resource in the data package. It is required and *should not* be
+changed, since it's used in the file name of the data resource's properties
+script. Since a data package can contain multiple resources, the resource's
```
signekb:
```suggestion
script. Because a data package can contain multiple resources, the resource's
```
Just to avoid since x 2 (used in the previous sentence as well)

signekb (dismissed):
Very nice 👌  A few suggestions:

lwjohnst86 (changes requested):
Nice! Some more comments :grin:

joelostblom (commented):
Thanks @lwjohnst86 , I accepted all your suggestions

lwjohnst86 (changes requested):
Just reformat and then I will approve and merge :+1:

lwjohnst86 (approved):

PR #1605, opened 2025-11-21
feat: ✨ integrate `check-datapackage`

### LINKED ISSUE #1582: Integrate `check-datapackage` into Sprout

martonvago:
# Description

This PR integrates `cdp` 🎉 .
Says many files, but most were just deleted 🔥 .

Closes #1582 

Needs an in-depth review.

## Checklist

- [x] Formatted Markdown
- [x] Ran `just run-all`


src/seedcase_sprout/check_properties.py
```diff
@@ -131,50 +139,138 @@ def _generic_check_properties(
...
-    properties_dict = properties.compact_dict
-
-    errors = cdp.check_properties(properties_dict)
+    package_required_checks = _map(
+        PACKAGE_SPROUT_REQUIRED_FIELDS,
+        lambda field: cdp.RequiredCheck(
+            jsonpath=f"$.{field}",
+            message=f"'{field}' is a required property",
+        ),
+    )
```
martonvago: Some required fields are required already in `cdp`, but I thought not separating those out looked cleaner. Issues will be deduplicated anyway.

One thing to note about deduplication is that for an issue to count as a duplicate all properties must be the same, including `message`. This actually places a restriction on the message format, especially in `RequiredCheck`s, because the format is determined by `jsonschema`. To prevent bugs, maybe we should not include the `message` when deciding if two issues are the same (at least for `type="required"`)?

Not sure how `explain` will influence this.

src/seedcase_sprout/check_properties.py
```diff
@@ -131,50 +139,138 @@ def _generic_check_properties(
...
+            "| $.contributors[*].title"
+            "| $.sources[*].title"
+            "| $.licenses[*].name"
+            "| $.licenses[*].path"
+            f"| {' | '.join(not_blank_resource_fields)}"
+        ),
+        message="This property must not be empty.",
+        check=lambda value: bool(value),
+        type="not-blank",
+    )
```
martonvago: Could also create a separate `CustomCheck` for each path instead of the union

src/seedcase_sprout/check_properties.py
```diff
@@ -131,50 +139,138 @@ def _generic_check_properties(
...
+            "Sprout doesn't use the `data` field, instead it expects data "
+            "in separate files that are given in the `path` field."
+        ),
+        check=lambda value: value is None,
+        type="no-inline-data",
+    )
+    exclude_resource_data = cdp.Exclusion(jsonpath="$.resources[*].data")
+    exclude_resource_path_or_data_required = cdp.Exclusion(
+        jsonpath="$.resources[*]", type="required"
+    )
```
martonvago: Only matches the case when both `path` and `data` are missing. Ignoring because we don't want any feedback on `data`. The missing `path` will be flagged separately.

src/seedcase_sprout/check_properties.py
```diff
@@ -131,50 +139,138 @@ def _generic_check_properties(
...
+    exclude_resource_path_type = cdp.Exclusion(
+        jsonpath="$.resources[*].path", type="type"
+    )
+    exclude_resource_path_pattern = cdp.Exclusion(
+        jsonpath="$.resources[*].path", type="pattern"
+    )
```
martonvago: Same reasoning as above: Sprout has its own rules for what paths should look like.

src/seedcase_sprout/check_properties.py
```diff
@@ -131,50 +139,138 @@ def _generic_check_properties(
...
-
-    if errors:
-        raise ExceptionGroup(
-            f"The following checks failed on the properties:\n{properties}", errors
-        )
 
     return properties
 
 
+def _check_resource_path_format(resource_properties: Any) -> bool:
```
martonvago: Moved from `get_sprout_resource_errors.py`

pyproject.toml
```diff
@@ -73,3 +74,6 @@ dev = [
 [build-system]
 requires = ["hatchling"]
 build-backend = "hatchling.build"
+
+[tool.uv.sources]
+check-datapackage = { git = "https://github.com/seedcase-project/check-datapackage" }
```
martonvago: Hope this is correct for the time being

tests/sprout_checks/test_check_properties.py
```diff
@@ -60,9 +57,23 @@ def test_passes_correct_properties(properties):
     assert check_resource_properties(properties.resources[0]) == properties.resources[0]
     assert check_resource_properties(properties.resources[1]) == properties.resources[1]
 
-    # Even when resources isn't there.
-    delattr(properties, "resources")
-    assert check_properties(properties) == properties
```
martonvago: Moved to test below

src/seedcase_sprout/check_properties.py
```diff
@@ -77,6 +77,18 @@ def check_properties(properties: Any) -> PackageProperties:
...
 
+class DataResourceError(Exception):
+    """Error listing issues in a data resource."""
+
+    def __init__(
+        self,
+        data_package_error: cdp.DataPackageError,
+    ) -> None:
+        """Create a `DataResourceError` from a `cdp.DataPackageError`."""
+        message = str(data_package_error).replace("package.resources[0]", "resource")
```
martonvago: Don't think this can be done more intelligently at this point, but it would be nicer if I wasn't replacing blindly.

Example output:
```
E   seedcase_sprout.check_properties.DataResourceError: 4 issues were found in your `datapackage.json`:
E   
E   At resource.description:
E   |
E   | description: None
E   |              ^^^^
E   'description' is a required property
```

Should we change/remove the reference to `datapackage.json`?

lwjohnst86: Would it help to instead take `Issues` as an arg and replace via that?

lwjohnst86 (commented):
Some small comments, otherwise, looking goooodddd

lwjohnst86 (changes requested):
Looking gooodddd :grin: small comments about the error/replacing issue.

lwjohnst86 (approved):
Amazing :tada:

PR #1636, opened 2026-01-10
docs: :memo: reorder interface toc to match index page

joelostblom:
# Description

The [interface index page lists inputs before outputs](https://sprout.seedcase-project.org/docs/design/interface/), which I think also makes more sense intuitively, so I suggests that the pages in the ToC are in the same order

Needs a quick review.


lwjohnst86 (approved):
:tada:

PR #1671, opened 2026-03-19
fix: :bug: expose unexposed types in properties

martonvago:
# Description

This PR exposes the 2 unexposed types in properties. I wanted to use one in `feasibility-data` and saw it wasn't there.

Needs a quick review.

## Checklist

- [x] Formatted Markdown
- [x] Ran `just run-all`


joelostblom (approved):
lgtm! Although I'm unsure about the history of things like `example_data_patients`, so maybe someone who knows about that should take a look too

lwjohnst86 (approved):
:tada: nice catch :+1:

PR #1696, opened 2026-04-14
refactor: ♻️ import functionals from soil

joelostblom:
# Description

Move to a shared place in soil.

Needs an in-depth review.

## Checklist

- [x] Ran `just run-all`


lwjohnst86 (approved):
:tada:

PR #1698, opened 2026-04-17
refactor: ♻️ rename `Properties` classes to be more specific

joelostblom:
# Description

Closes https://github.com/seedcase-project/seedcase-soil/pull/21

Needs a quick/an in-depth review.

## Checklist

- [x] Ran `just run-all`


martonvago (dismissed):
🙌🙌🙌

docs/design/interface/inputs.qmd
```diff
@@ -22,7 +22,7 @@ Properties
     [dataclasses](https://docs.python.org/3/library/dataclasses.html)
     we've created to represent the properties (metadata) of a data
     package or resource. They always use the `*Properties` suffix, such
-    as `PackageProperties` and `ResourceProperties`.
+    as `SproutProperties` and `ResourceProperties`.
```
lwjohnst86:
```suggestion
    as `SproutProperties` and `ResourceProperties`. Because there
    isn't a direct mapping between the Data Package schema and
    Python classes, we decided to call the "base" properties object
    that we converted from the schema into classes `SproutProperties`
    to signify that the mapping is our interpretation of the schema
    and is based on our needs for Sprout.
```

@martonvago would you say that's a fair way to explain it?

lwjohnst86 (changes requested):
Nice! Only a very minor change to explain the `Sprout*` part.

docs/design/interface/inputs.qmd
martonvago: Mostly, I think!
Just:
- there could be a direct mapping, it's just that in Sprout we don't need (or actively don't want) the full flexibility of the DP schema
- calling `SproutProperties` a "base object" is a bit misleading because it's not a base for the other properties classes. Plus, we do have an actual base class (`BaseProperties`).

joelostblom: I made an attempt to combine your suggestions in 2c848a7

lwjohnst86 (dismissed):
:tada: love it :+1:

lwjohnst86:
@joelostblom can you resolve the conflicts?

joelostblom:
Resolved!

lwjohnst86 (approved):

PR #1703, opened 2026-04-27
refactor: ♻️ use Soil when reading and writing

### LINKED ISSUE #1702: Sync with `read`/`wite_properties` in soil?
I'm suggesting adding a `write_properties` function in soil in PR https://github.com/seedcase-project/seedcase-soil/pull/27. I wonder if we should try to sync that somehow with what is in sprout currently. I'm not familiar with all the additional bells and whistles in sprout with the dedenting etc, so I'm not sure what can be factored out and not, but at least soil's `write_properties` seems like it could be a replacement for the internal `write_json` in sprout?

The same reasoning would go for read_properties too. We can also decide whether to keep the names exactly the same or whether it is confusing that there is a function with the same name in two related packages.

martonvago:
# Description

This PR changes to `seedcase_soil.write_properties` and `seedcase_soil.read_properties` for JSON reading/writing.

Closes #1702 

Needs an in-depth review.

## Checklist

- [x] Formatted Markdown
- [x] Ran `just run-all`


src/seedcase_sprout/read_properties.py
```diff
@@ -30,12 +31,12 @@ def read_properties(path: Path | None = None) -> SproutProperties:
         ```
 
     Raises:
-        FileNotFound: If the `datapackage.json` file doesn't exist.
-        JSONDecodeError: If the `datapackage.json` file couldn't be read.
+        ss.errors.FileDoesNotExistError: If the file cannot be found.
+        ss.errors.JSONFormatError: If the JSON file is malformatted.
```
martonvago: From Soil. `errors` is not actually exposed through `init`, so `ss.errors` is just descriptive.

src/seedcase_sprout/read_properties.py
```diff
@@ -1,7 +1,8 @@
 from pathlib import Path
 
+import seedcase_soil as ss
```
joelostblom: What do you think about importing as `so`? I'm thinking that would be in line with that we do `fl` for `seedcase_flower` and `sp` for `seedcase_sprout` (and less importantly what I started doing for soil elsewhere). I'm also thinking that `ss` has some unfortunate historical connotations that we might want to avoid any chance of invoking.

joelostblom (dismissed):
Thanks for making this change! I think it looks good overall; just one comment about the import convention.

src/seedcase_sprout/read_properties.py
martonvago: Ah, I also thought of that but then wasn't sure other people would too

lwjohnst86 (approved):

PR #1759, opened 2026-07-16
refactor: 🔥 remove `write_resource_batch()`, as we read from staging

lwjohnst86:
# Description

Based on what we're doing in feasibility-data, we don't need `write_resource_batch()` anymore since we'll be saving to `staging/` instead (and doing the writing outside of Sprout. A big issue with `write_resource_batch()` was that it checked against the `datapackage.json`. But we want to save to staging and use that data to generate the metadata. The most important checks are between staging and the final data resource, not before that.

I also removed the section in the guide about saving to the final data resource. I deleted it because it was actually in the `resource-metadata.qmd` file, so didn't belong. Plus, we'll be updating that staging to resource flow anyway.

Needs a thorough review.

## Checklist

- [x] Ran `just run-all`


docs/design/interface/flows.qmd
```diff
@@ -92,46 +92,10 @@ flowchart
...
-    resource_properties --> write_resource_batch
-    data -.-> check_data
-    resource_properties -.-> check_data
-```
-
 ## Checking data against the properties
 
 The data must always match what is described in the properties. This means that
-the data must have the same column names, column types, and column constraints.
+the data must have the same column names, column types,> and column constraints.
```
martonvago:
```suggestion
the data must have the same column names, column types, and column constraints.
```

martonvago (changes requested):
Just a typo

docs/design/interface/flows.qmd
```diff
@@ -203,50 +167,13 @@ flowchart
...
-    list_data_updated --> write_resource_batch
-    properties_updated --> write_resource_batch
-```
-:::
+<!-- TODO: Consider this more, but hide it for now. -->
 
 ## Deleting an observational unit
 
 If you need to delete an observational unit from the data, you can use this
 flow. The `delete_observational_unit()` function will delete the observational
```
signekb:
```suggestion
flow. The `delete_observational_unit()` function deletes the observational
```

docs/design/interface/flows.qmd
```diff
@@ -203,50 +167,13 @@ flowchart
...
-unit from the data and output a list of DataFrames with the deletions. The
-`write_resource_batch()` function will save the updated data back to the batch
-files.
+unit from the data and output a list of DataFrames with the deletions.
```
signekb:
```suggestion
unit from the data and outputs a list of DataFrames containing the updated data.
```

signekb (commented):
🔥 🔥 
Two minor suggestions:

martonvago (dismissed):
🎉

signekb (dismissed):

signekb (approved):

PR #1768, opened 2026-07-22
fix: 🐛 remove extra quartodoc build step

martonvago:
# Description

There was this ?leftover? quartodoc build step that was not running in the project env and caused the website build to fail. Also needs to be fixed in Template Python Package, but I don't seem to have write access ([issue](https://github.com/seedcase-project/template-python-package/issues/321)).

Closes #

Needs a quick review.

## Checklist

- [x] Ran `just run-all` -- pre-commit is also failing remotely, but that's a separate pyrefly issue

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -124,9 +124,6 @@ jobs:
       - name: Build function reference docs
         run: uv run quartodoc build
 
-      - name: Build function reference docs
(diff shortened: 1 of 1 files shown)


signekb (approved):

joelostblom (approved):

lwjohnst86:
Fyi, this should have been "ci" type not "fix"

PR #1777, opened 2026-08-05
docs: 🏗️ add design of `init-metadata` CLI

### LINKED ISSUE #1686: Consider design of a CLI for creating the template scripts
e.g. `create_properties_script()`

lwjohnst86:
# Description

Not sure about the name `init-metadata` yet, but better than `compose-metadata` (my original suggestion) and the  original `create-template` was not accurate (we are copying from the template, not creating a template). 

Close #1686

Needs a thorough review.

## Checklist

- [x] Ran `just run-all`


docs/design/interface/cli.qmd
```diff
@@ -28,7 +28,7 @@ As a convenience, we provide the command `sprout` as a shorthand for
 instances.
 :::
 
-`seedcase-sprout` has several commands: `extract-metadata`, `create-template`,
+`seedcase-sprout` has several commands: `extract-metadata`, `compose-metadata`,
```
martonvago: How about `create-metadata`?

docs/design/interface/python.qmd
```diff
@@ -43,3 +43,32 @@ flowchart
...
+`resource`.
+
+```{mermaid}
+%%| label: fig-flow-compose-metadata
+%%| fig-cap: >-
+%%|   Diagram showing the flow of objects and functions to compose the metadata
+%%|   in a Python script.
+%%| fig-alt: >-
+%%|   A flow diagram showing the steps to extract resource metadata from a
+%%|   Parquet file.
```
martonvago:
```suggestion
```

docs/design/interface/cli.qmd
```diff
@@ -55,11 +55,45 @@ seedcase-sprout extract-metadata [PARQUET-PATH] --output/-o [OUTPUT-PATH]
...
+out some metadata when none yet exists. It creates a Python script with empty
+`*Properties()` dataclasses. It has the signature:
+
+```bash {filename="Terminal"}
+seedcase-sprout compose-metadata [OUTPUT-PATH] --type [TYPE]
+```
+
+The `--type` determines whether to make a file with metadata for the package
+overall (top-level metadata) or for a resource file. The default is the package
+overall.
```
martonvago: So if type is `package`, would the script just list a commented-out resource properties (as it does now)?

martonvago (changes requested):
Very nice!

docs/design/interface/cli.qmd
lwjohnst86: Hm, I thought of that too, but it isn't completely accurate because it isn't making the metadata, it's making a template that the user would then use to "create" the metadata. :thinking:

docs/design/interface/cli.qmd
lwjohnst86: Yea, exactly :relaxed:

docs/design/interface/python.qmd
lwjohnst86:
```suggestion
%%|   A flow diagram showing the steps to creating the Python file with the
%%|   empty metadata fields.
```

docs/design/interface/cli.qmd
signekb: Hmmmm, `init-metadata`, maybe?

docs/design/interface/cli.qmd
```diff
@@ -55,11 +55,45 @@ seedcase-sprout extract-metadata [PARQUET-PATH] --output/-o [OUTPUT-PATH]
...
     parquet --> extract
     output_opt --> extract
     extract --> output
 ```
+
+### {{< var wip >}} `compose-metadata`
+
+The `compose-metadata` command creates a Python script with metadata fields that
+haven't been filled in yet. This command is meant to be used to start filling
+out some metadata when none yet exists. It creates a Python script with empty
```
signekb:
```suggestion
haven't been filled in yet. It creates a Python script with empty
```
To me, that sentence doesn't add much, so we could remove it

docs/design/interface/cli.qmd
signekb: However, there's something about this that could sound like initialising or composing the `datapackage.json` file and not a Python script 🤔

docs/design/interface/cli.qmd
```diff
@@ -55,11 +55,45 @@ seedcase-sprout extract-metadata [PARQUET-PATH] --output/-o [OUTPUT-PATH]
...
+`*Properties()` dataclasses. It has the signature:
+
+```bash {filename="Terminal"}
+seedcase-sprout compose-metadata [OUTPUT-PATH] --type [TYPE]
```
signekb: Does the output path include the file name? Or is that set?
What's the file name when the type is resource?

docs/design/interface/cli.qmd
signekb: If type is resource, should it then be (manually?) be added to a package properties script?

signekb (changes requested):
Nice! Looking forward to these cli commands 👌 Some questions:

docs/design/interface/cli.qmd
martonvago: I kinda like `init-metadata`. `extract-metadata` also produces a script and not `datapackage.json` 🤷

martonvago (dismissed):

docs/design/interface/cli.qmd
lwjohnst86: Agreed! Let's go with `init-metadata` for now, since as @martonvago said it matches the behaviour of `extract-metadata`. We can always change it later.

martonvago (approved):

signekb (approved):
I like it!

PR #1778, opened 2026-08-05
fix: 🏷️ resolve Python check errors (e.g. with types)

lwjohnst86:
# Description

Went through and fixed the errors from the typechecker and Ruff.

Needs a quick review.

## Checklist

- [x] Ran `just run-all`

### DIFF
--- a/src/seedcase_sprout/check_properties.py
+++ b/src/seedcase_sprout/check_properties.py
@@ -217,7 +217,7 @@ def _generic_check_properties(
             "Sprout doesn't use the `data` field, instead it expects data "
             "in separate files that are given in the `path` field."
         ),
-        check=lambda value: value is None,  # type: ignore # typechecker can't determine output of lambda
+        check=lambda value: value is None,  # type: ignore # typechecker can't infer lambda parameter type
         type="no-inline-data",
     )
     exclude_resource_data = cdp.Exclusion(jsonpath="$.resources[*].data")
--- a/src/seedcase_sprout/check_properties.py
+++ b/src/seedcase_sprout/check_properties.py
@@ -1,5 +1,5 @@
 from dataclasses import replace
-from typing import Any
+from typing import Any, Optional
 
 import check_datapackage as cdp
 from seedcase_soil import fmap
@@ -133,7 +133,7 @@ def check_resource_properties(properties: Any) -> ResourceProperties:
 
 def _generic_check_properties(
     properties: SproutProperties,
-    exclusions: list[cdp.Exclusion] = [],
+    exclusions: Optional[list[cdp.Exclusion]] = None,
     error: bool = True,
 ) -> list[cdp.Issue]:
     """Checks Sprout's requirements against the Data Package spec.
@@ -160,6 +160,9 @@ def _generic_check_properties(
     Raises:
         DataPackageError: an error flagging issues in the properties.
     """
+    if exclusions is None:
+        exclusions = []
+
     package_required_checks = fmap(
         PACKAGE_SPROUT_REQUIRED_FIELDS,
         lambda field: cdp.RequiredCheck(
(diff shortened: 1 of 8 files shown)


src/seedcase_sprout/check_properties.py
```diff
@@ -189,13 +192,13 @@ def _generic_check_properties(
             f"| {' | '.join(not_blank_resource_fields)}"
         ),
         message="This property must not be empty.",
-        check=lambda value: bool(value),
+        check=lambda value: bool(value),  # type: ignore # typechecker can't determine output of lambda
```
martonvago: Were these also flagged by Ruff? I can't seem to get them.

lwjohnst86: These were flagged by pyrefly after I updated it.

signekb: And we don't want to fix/change them instead of ignoring them? :)

signekb (commented):
Approved, but with one comment:

src/seedcase_sprout/check_properties.py
martonvago:
```suggestion
        check=bool,
```

martonvago: Ah, switching to the pre-release version flagged them for me too

src/seedcase_sprout/check_properties.py
```diff
@@ -189,13 +192,13 @@ def _generic_check_properties(
...
         type="not-blank",
     )
     resource_path_string = cdp.CustomCheck(
         jsonpath="$.resources[*].path",
         message="Resource path must be of type string.",
-        check=lambda value: isinstance(value, str),
+        check=lambda value: isinstance(value, str),  # type: ignore # typechecker can't determine output of lambda
```
martonvago:
```suggestion
        check=lambda value: isinstance(value, str),  # type: ignore # typechecker can't infer lambda parameter type
```
Because the problem is the input rather than the output.
Same for the others.

src/seedcase_sprout/check_properties.py
```diff
@@ -204,7 +207,7 @@ def _generic_check_properties(
             "Resource path must have the format "
             "`resources/<resource-name>/data.parquet`."
         ),
-        check=lambda value: _check_resource_path_format(value),
+        check=lambda value: _check_resource_path_format(value),  # type: ignore # typechecker can't determine output of lambda
```
martonvago:
```suggestion
        check=_check_resource_path_format,
```

martonvago (changes requested):

signekb:
Would this not be a `style: 🎨 `?
- `style` instead of `fix` bc this does not change the meaning of the code and it's not about fixing a bug
- 🎨 instead of 🏷️ because this is about improving the structure/format of the code (e.g., type hints) and not about actually updating the types?

src/seedcase_sprout/check_properties.py
```diff
@@ -214,7 +217,7 @@ def _generic_check_properties(
             "Sprout doesn't use the `data` field, instead it expects data "
             "in separate files that are given in the `path` field."
         ),
-        check=lambda value: value is None,
+        check=lambda value: value is None,  # type: ignore # typechecker can't determine output of lambda
```
lwjohnst86:
```suggestion
        check=lambda value: value is None,  # type: ignore # typechecker can't infer lambda parameter type
```

src/seedcase_sprout/properties.py
```diff
@@ -44,7 +44,7 @@ def compact_dict(self) -> dict[str, Any]:
         """
         return asdict(
             obj=self,
-            dict_factory=lambda tuples: {
+            dict_factory=lambda tuples: {  # type: ignore # typechecker can't determine output of lambda
```
lwjohnst86:
```suggestion
            dict_factory=lambda tuples: {  # type: ignore # typechecker can't infer lambda parameter type
```

src/seedcase_sprout/check_properties.py
lwjohnst86: Fixing them makes the code substantially more complicated/larger. It requires making new functions for each use of the lambda's here.

lwjohnst86: I started to fix them but it made the code more complex (I felt anyway).

lwjohnst86:
Good questions @signekb. Mostly I agree, but the fact that some code was actually changed (by @martonvago's edits), it now is less of a style and more of an actual change.

martonvago (approved):

PR #1781, opened 2026-08-10
ci: 👷 check commits from latest tag, earlier ones will fail

lwjohnst86:
# Description

Cocogitto fails because there are non-conforming commits. This only checks from the last tag.

Needs a quick review.

## Checklist

- [x] Ran `just run-all`


signekb (approved):
Ah, makes sense 👍

